PR #4, opened 2015-04-26
Added multimap/multiset containers

GatorQue:
It is done, multimap and multiset are now implemented.

### DIFF
--- a/fnv_1.h
+++ b/fnv_1.h
@@ -5,7 +5,6 @@ The MIT License(MIT)
 
 Embedded Template Library.
 https://github.com/ETLCPP/etl
-https://github.com/ETLCPP/etl
 
 Copyright(c) 2014 jwellbelove
 
--- a/imap.h
+++ b/imap.h
@@ -434,16 +434,6 @@ namespace etl
 
 
     //*************************************************************************
-    /// Assignment operator.
-    //*************************************************************************
-    imap& operator = (const imap& rhs)
-    {
-      assign(rhs.cbegin(), rhs.cend());
-
-      return *this;
-    }
-
-    //*************************************************************************
     /// Gets the beginning of the map.
     //*************************************************************************
     iterator begin()
@@ -779,7 +769,7 @@ namespace etl
     ///\param position The position that would precede the value to insert.
     ///\param value    The value to insert.
     //*********************************************************************
-    iterator insert(iterator position, const value_type& value)
+    iterator insert(iterator, const value_type& value)
     {
       // Default to no inserted node
       Node* inserted_node = nullptr;
@@ -790,7 +780,7 @@ namespace etl
         Data_Node& node = allocate_data_node(value);
 
         // Obtain the inserted node (might be nullptr if node was a duplicate)
-        inserted_node = insert_node(find_node(root_node, position.p_node), node);
+        inserted_node = insert_node(root_node, node);
       }
       else
       {
@@ -811,7 +801,7 @@ namespace etl
     ///\param position The position that would precede the value to insert.
     ///\param value    The value to insert.
     //*********************************************************************
-    iterator insert(const_iterator position, const value_type& value)
+    iterator insert(const_iterator, const value_type& value)
     {
       // Default to no inserted node
       Node* inserted_node = nullptr;
@@ -822,7 +812,7 @@ namespace etl
         Data_Node& node = allocate_data_node(value);
 
         // Obtain the inserted node (might be nullptr if node was a duplicate)
-        inserted_node = insert_node(find_node(root_node, position.p_node), node);
+        inserted_node = insert_node(root_node, node);
       }
       else
       {
@@ -1151,7 +1141,7 @@ namespace etl
         }
       }
 
-      // Return root node if nothing or duplicate was found
+      // Return root node if nothing was found
       return root_node;
     }
 
@@ -1325,36 +1315,6 @@ namespace etl
     }
 
     //*************************************************************************
-    /// Find the node whose key is not considered to go before the key provided
-    //*************************************************************************
-    const Node& find_lower_node(const Node* position, const key_value_parameter_t& key) const
-    {
-      // Something at this position? keep going
-      const Node* lower_node = position;
-      while (lower_node)
-      {
-        // Downcast lower node to Data_Node reference for key comparisons
-        const Data_Node& data_node = imap::data_cast(*lower_node);
-        // Compare the key value to the current lower node key value
-        if (node_comp(key, data_node))
-        {
-          lower_node = lower_node->children[kLeft];
-        }
-        else if (node_comp(data_node, key))
-        {
-          lower_node = lower_node->children[kRight];
-        }
-        else
-        {
-          break;
-        }
-      }
-
-      // Return the lower_node position found
-      return lower_node;
-    }
-
-    //*************************************************************************
     /// Find the node whose key is considered to go after the key provided
     //*************************************************************************
     Node* find_upper_node(Node* position, const key_value_parameter_t& key) const
@@ -1367,44 +1327,6 @@ namespace etl
       {
         // Downcast position to Data_Node reference for key comparisons
         Data_Node& data_node = imap::data_cast(*node);
-        // Compare the key value to the current upper node key value
-        if (node_comp(key, data_node))
-        {
-          upper_node = node;
-          node = node->children[kLeft];
-        }
-        else if (node_comp(data_node, key))
-        {
-          node = node->children[kRight];
-        }
-        else if (node->children[kRight])
-        {
-          upper_node = find_limit_node(node->children[kRight], kLeft);
-          break;
-        }
-        else
-        {
-          break;
-        }
-      }
-
-      // Return the upper node position found (might be nullptr)
-      return upper_node;
-    }
-
-    //*************************************************************************
-    /// Find the node whose key is considered to go after the key provided
-    //*************************************************************************
-    const Node* find_upper_node(const Node* position, const key_value_parameter_t& key) const
-    {
-      // Keep track of parent of last upper node
-      const Node* upper_node = nullptr;
-      // Start with position provided
-      const Node* node = position;
-      while (node)
-      {
-        // Downcast position to Data_Node reference for key comparisons
-        const Data_Node& data_node = imap::data_cast(*node);
         // Compare the key value to the current upper node key value
         if (node_comp(key, data_node))
         {
--- a/imultimap.h
+++ b/imultimap.h
@@ -27,9 +27,9 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 SOFTWARE.
 ******************************************************************************/
 
-#ifndef __ETL_ISET__
-#define __ETL_ISET__
-#define __ETL_IN_ISET_H__
+#ifndef __ETL_IMULTIMAP__
+#define __ETL_IMULTIMAP__
+#define __ETL_IN_IMULTIMAP_H__
 
 #include <iterator>
 #include <algorithm>
@@ -37,7 +37,7 @@ SOFTWARE.
 #include <stddef.h>
 
 #include "nullptr.h"
-#include "set_base.h"
+#include "map_base.h"
 #include "type_traits.h"
 #include "parameter_type.h"
 #include "pool.h"
@@ -49,28 +49,30 @@ SOFTWARE.
 namespace etl
 {
   //***************************************************************************
-  /// A templated base for all etl::set types.
-  ///\ingroup set
+  /// A templated base for all etl::multimap types.
+  ///\ingroup map
   //***************************************************************************
-  template <typename T, typename TCompare>
-  class iset : public set_base
+  template <typename TKey, typename TMapped, typename TKeyCompare>
+  class imultimap : public map_base
   {
   public:
 
-    typedef const T     key_type;
-    typedef const T     value_type;
-    typedef TCompare    key_compare;
-    typedef TCompare    value_compare;
-    typedef value_type& const_reference;
-    typedef value_type* const_pointer;
-    typedef size_t      size_type;
+    typedef std::pair<const TKey, TMapped> value_type;
+    typedef const TKey                     key_type;
+    typedef TMapped                        mapped_type;
+    typedef TKeyCompare                    key_compare;
+    typedef value_type&                    reference;
+    typedef const value_type&              const_reference;
+    typedef value_type*                    pointer;
+    typedef const value_type*              const_pointer;
+    typedef size_t                         size_type;
 
     //*************************************************************************
     /// How to compare two key elements.
     //*************************************************************************
     struct key_comp
     {
-      bool operator ()(key_type& key1, key_type& key2) const
+      bool operator ()(const key_type& key1, const key_type& key2) const
       {
         return key_compare()(key1, key2);
       }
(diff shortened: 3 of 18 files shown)


jwellbelove:
Great stuff! I'll pull it when I get back from holiday next weekend.

Then maybe I'll get my arse in gear and finish the unordered maps and sets!


I'm not sure what we should be adding next. I think we've covered most of the standard library that needed to be optimised for embedded, and C++ 11 additions that are achievable with C++ 03. Any suggestions?


Thinking of work I've done in the past for embedded projects, I did contemplate adding a templated message bus or a message passing cooperative task scheduler, for those that don’t want a full blown embedded OS and would rather not write there own. Also maybe a timer scheduler that can be hooked to a timer interrupt, with registered callbacks, supporting one shot and periodic timers.


John.






From: Ryan Lindeman
Sent: ‎Sunday‎, ‎26‎ ‎April‎ ‎2015 ‎3‎:‎13‎ ‎am
To: ETLCPP/etl





It is done, multimap and multiset are now implemented.



You can view, comment on, or merge this pull request online at:

  https://github.com/ETLCPP/etl/pull/4

Commit Summary
Merge pull request #1 from ETLCPP/master
add multimap and multiset containers

File Changes
M fnv_1.h (1) 
M imap.h (88) 
A imultimap.h (1830) 
A imultiset.h (1813) 
M iset.h (87) 
M map_base.h (4) 
A multimap.h (110) 
A multiset.h (110) 
M set.h (1) 
M set_base.h (5) 
M test/test_map.cpp (3) 
A test/test_multimap.cpp (949) 
A test/test_multiset.cpp (912) 
M test/test_set.cpp (13) 
M test/vs2013/etl.vcxproj (6) 
M test/vs2013/etl.vcxproj.filters (18) 

Patch Links:
https://github.com/ETLCPP/etl/pull/4.patch
https://github.com/ETLCPP/etl/pull/4.diff

GatorQue:
All of those sound like great ideas, can you elaborate more about the
message bus and message passing task scheduler?  Are you thinking of a type
of light weight threading scheme since there is no OS/CPU support for
multiple processes?  Also, I think that having a robust timer/event system
would be a great addition.  I also think having a general purpose GPIO and
Serial bus interface is useful so you can write your embedded application
once against it and just introduce new GPIO or Serial port implementations
without changing the application software.  In other words, a Hardware
Abstraction Layer mechanism would also be helpful in many ways for porting
capabilities from one processor to another.  My favorite example of this is
the POCO C++ open source library.  What about some abstract robust way of
deploying software over the network/serial/console port?  Any of these seem
like a good place to work on next?

Ryan Lindeman

jwellbelove:
Hi Ryan,


The message bus idea is sort of like an enhance observer pattern. Similar in some ways to a hardware serial bus, messages are transmitted on the bus and interested components pick 

them up. The sender doesn’t have to know about the parties consuming the messages. The receivers may either register for specific messages or receive all of them and do the filtering themselves. 


The message passing scheduler is basically a simple prioritised or round robin scheduler that passes control to a task whenever a message is present in its queue. The task is allowed to run to completion before control is returned to the scheduler. The is no pre-emptive scheduling. This can seriously simplify the creation of separate tasks as there will never be contention between shared memory or variables. With some simple lock-free FIFOs messages may also be sent from interrupts.


I’m glad you like the timer/event system idea; it's something I’ve found myself writing many times when I’ve worked on systems with no OS. Otherwise the code ends up getting clogged with ad-hoc solutions.


There may also be scope for adding a very low overhead signal/slot mechanism; simpler than Qt or boost and probably capacity limited as with the containers.


I like the idea of abstracted IO facilities, though getting a truly flexible, though efficient implementation would require a lot of careful thought.


One thing I have in the back of my mind is that I don't really want the ETL to turn into an ‘all singing, all dancing’ library that tries to cram in every ‘neat’ idea. I’d really like it to stay tailored to small, resource limited systems.



(Also I don't want it to take over all of my spare time!!)


John.





From: Ryan Lindeman
Sent: ‎Tuesday‎, ‎28‎ ‎April‎ ‎2015 ‎2‎:‎01‎ ‎am
To: ETLCPP/etl
Cc: John Wellbelove




All of those sound like great ideas, can you elaborate more about the
message bus and message passing task scheduler? Are you thinking of a type
of light weight threading scheme since there is no OS/CPU support for
multiple processes? Also, I think that having a robust timer/event system
would be a great addition. I also think having a general purpose GPIO and
Serial bus interface is useful so you can write your embedded application
once against it and just introduce new GPIO or Serial port implementations
without changing the application software. In other words, a Hardware
Abstraction Layer mechanism would also be helpful in many ways for porting
capabilities from one processor to another. My favorite example of this is
the POCO C++ open source library. What about some abstract robust way of
deploying software over the network/serial/console port? Any of these seem
like a good place to work on next?

Ryan Lindeman

GatorQue:
All of those ideas sound like something I am interested in. I agree with
you on the last point, is there a line we want to draw or in other words -
is there something you think we should split out already?  Here are a few
questions I have about your descriptions:

Message Bus: What would be templated in this system?  I did a similar
project at my last job where the only thing that was really templated was
the ability to extract/set bits within a data word. The system was called a
FieldMapper since you predefined a list of fields inside a message and
created a FieldMapper object that took the list of fields and granted
access to each fields contents. At the heart of the system was a way of
setting/extracting specific field value types from the raw array of
characters you received from the communication device.  But I'm not sure if
this is what you had in mind or not.  Would this system be based on a
zero-copy system where we pass around pointers to the messages to be passed
and let the compiler determine the destinations at compile time based on
the template arguments?

The message passing scheduler seems straight forward enough and is likely
related to the timer/event and signal/slot mechanism (all of them seem to
have function pointers that would be registered to handle different types
of events).

I think we can safely throw out the abstract IO facilities for now until we
get more of a demand for something like that.

Do you envision the message bus system using one of our containers we have
created for ETL or something else entirely?  Thanks for your continued
discussion on this.

Ryan Lindeman

jwellbelove:
The templated part of the message bus (and message queues for the  scheduler) could be similar to that used in the etl::visitor pattern class. The message types handled would be specified in the class declaration causing a pure virtual handler member function to be declared for each one. The implementer of the derived class would be forced to override every handler, ensuring that no message is accidentally left out. I’m not yet if this would scale well if the coder was forced to define a class with dozens of message types. I have worked on some ideas where the message types were defined by automatically assigned id.s (at compile time) , where the templated message handler automatically built its own switch/case tree to.


At my last job I devised a message handler system based on the template pure virtual methods idea. I had wanted to just pass pointers to messages, but the technical director insisted that we always passed copies. This made the message queues a little inefficient in that they were required to handle N copies of the longest message type. I would have preferred a system where messages were allocated from pools and were automatically released after processing.


For the message bus I envisage that it would work as a single threaded task. When a message is passed to the bus all of the interested parties are called with the message. Control is only passed back when the last registered party has been called. I suppose it could be seen as a software  CAN bus, where all messages are broadcast and can be picked up by any party that is interested in the message. I'm not sure yet how much different it would ultimately be to etl::observer. The container in etl::observer is there to store pointers to all of the registered observers for the subject.


There are two ways I can see the  scheduler/timer/message/event system going. 

One is to integrate all of the systems together with all of the components communicating via messages. The downside of this would be to force the user of the library to use a messaging system whether it was appropriate for their application or not.


The other is to completely separate the components.

The scheduler would merely ask the task whether it required processing time and pass control if it does, the order based on the task priority policy. The tasks may or may not use message processing queues. The timer services would call a registered callback function (based on etl::function). It would be up to the implementer as to whether this is translated into a message to a task or not.


I like this approach better as it can be used by someone without the worry that they will have to compile in unnecessary and unwanted library code.



John.




From: Ryan Lindeman
Sent: ‎Wednesday‎, ‎29‎ ‎April‎ ‎2015 ‎2‎:‎59‎ ‎am
To: ETLCPP/etl
Cc: John Wellbelove




All of those ideas sound like something I am interested in. I agree with
you on the last point, is there a line we want to draw or in other words -
is there something you think we should split out already? Here are a few
questions I have about your descriptions:

Message Bus: What would be templated in this system? I did a similar
project at my last job where the only thing that was really templated was
the ability to extract/set bits within a data word. The system was called a
FieldMapper since you predefined a list of fields inside a message and
created a FieldMapper object that took the list of fields and granted
access to each fields contents. At the heart of the system was a way of
setting/extracting specific field value types from the raw array of
characters you received from the communication device. But I'm not sure if
this is what you had in mind or not. Would this system be based on a
zero-copy system where we pass around pointers to the messages to be passed
and let the compiler determine the destinations at compile time based on
the template arguments?

The message passing scheduler seems straight forward enough and is likely
related to the timer/event and signal/slot mechanism (all of them seem to
have function pointers that would be registered to handle different types
of events).

I think we can safely throw out the abstract IO facilities for now until we
get more of a demand for something like that.

Do you envision the message bus system using one of our containers we have
created for ETL or something else entirely? Thanks for your continued
discussion on this.

Ryan Lindeman

GatorQue:
John,

Here is what I have come up with so far, but first some background.  My
favorite description of messaging systems can be found here (
http://www.eaipatterns.com/index.html). In thinking about the Message Bus,
I have come up with the following that I feel scales well:

Class: channel_base
Base Class: none
Purpose: To serve as the base class for all message channels so that we can
have a container full of channel_base classes (see message_scheduler class
below).
Methods:
bool done() - Pure virtual method that must return true if there are
messages to be processed
next() - Pure virtual method that must be defined for moving to the next
message to be processed
process() - Pure virtual method that might trigger a message to be processed

Class: message_channel<TMessage, MAX_MESSAGES>
Base Class: channel_base
Purpose: Provide a pool of TMessage type classes that are managed internally
Methods:
bool get(TMessage*&) - returns false if TMessage* was not provided
(internal pool of empty TMessage pointers was empty)
put(TMessage*) - puts TMessage into internal FIFO (queue?) for processing
by another Task/Thread
bool done() - returns true if there is at least one message in the FIFO
(queue?) to process (e.g. !queue->empty())
TMessage* operator*() - allows access to the head of the queue
next() - Returns the head of the FIFO (queue?) to the internal pool of
empty TMessage pointers (e.g. queue->pop())
process() - Default implementation just calls the next() method to skip
processing the message

Class: observer_channel<TMessage, MAX_MESSAGES, MAX_OBSERVERS>
Base Class: message_channel<TMessage, MAX_MESSAGES>
Purpose: Provides a way to notify registered observers of this channel of
messages to be processed during the process() method call. Internally
stores a list of observers for the TMessage type. Probably provides a
typedef for the etl::ifunction method to use when registering an observer.
Methods:
process() - Redefines the message_channel::process() method to instead
notify all registered observers of the most recent message available

Class: message_scheduler
Base Class: none
Purpose: To provide a sorted container (etl::multimap?) of channel_base*
classes (TValue) according to priority (TKey) specified at registration
time. Then during the process method call, the message_scheduler will loop
through each registered channel_base class and call its process method for
each channel_base whose done method returns true in order of priority.
Typedefs provided:
typedef etl::ifunction<message_scheduler, &process_func>* process_func_ptr
Classes provided:
Methods:
add_channel(channel_base*, int priority) - Adds the provided channel with
the given priority
remove_channel(channel_base*, int priority) - Removes the specified channel
with the given priority
process() - Target of process_func. Checks each channel_base for messages
to be processed

------
In typing this up I can still see a few flaws with the architecture but
perhaps it will help us discuss further the details.

GatorQue:
John,

Here is my second pass at the message bus classes.  I'm doing my best to
use the observer pattern as defined, but it doesn't seem to work well with
the etl::function algorithm.  I like this second pass better than my first
pass as the notification process is now handled internally in the
observer_channel class and the message_scheduler class doesn't need to know
anything about the message types (they just get registered as needed).  I
think the pattern scales well because you can just use the message_channel
class by itself in simple situations, or message_scheduler in complicated
priority situations.  What do you think?  Is there another architecture
that you feel would work better than this one?  Thanks...

Class: channel_base
Base Class: none
Purpose: To serve as the base class for all message channels so that we can
have a container full of channel_base classes (see message_scheduler class
below).
Methods:
bool done() - Pure virtual method that must return true if there are
messages to be processed
next() - Pure virtual method that must be defined for moving to the next
message to be processed
process() - Pure virtual method that might trigger a message to be processed

Class: message_channel<TMessage, MAX_MESSAGES>
Base Class: channel_base
Purpose: Provide a pool of TMessage type classes that are managed internally
Methods:
bool get(TMessage*&) - returns false if TMessage* was not provided
(internal pool of empty TMessage pointers was empty)
put(TMessage*) - puts TMessage into internal FIFO (queue?) for processing
by another Task/Thread
bool done() - returns true if there is at least one message in the FIFO
(queue?) to process (e.g. !queue->empty())
TMessage* operator*() - allows access to the head of the queue
next() - Returns the head of the FIFO (queue?) to the internal pool of
empty TMessage pointers (e.g. queue->pop())
process() - Default implementation just calls the next() method to skip
processing the message

Class: channel_observer<TMessage, MAX_MESSAGES>
Base Class: etl::observer<std::pair<message_channel<TMessage,
MAX_MESSAGES>, TMessage> >
Purpose: Provide a class that can handle providing the notification event
to an externally provided class
Typedefs provided:
typedef TMessage message_type;
typedef message_channel<message_type, MAX_MESSAGES> channel_type;
typedef std::pair<channel_type, message_type> channel_message;
typedef etl::ifunction<channel_type, message_type>* function_ptr;
Methods:
set_function(function_ptr* function)
notification(channel_message);

Class: observer_base
Base Class: none
Purpose: To serve as the base class for observer_channel classes so they
can be stored in a container
Methods:
process() - Pure virtual method that will be defined by the derived class

Class: observer_channel<channel_observer<TMessage, MAX_MESSAGES>,
MAX_OBSERVERS>
Base Class: observer_base AND etl::observable<channel_observer<TMessage,
MAX_MESSAGES>, MAX_OBSERVERS>
Purpose: Provides a way to notify registered observers of the
message_channel that will be registered for this observer_channel
Typedefs provided:
typedef TMessage message_type;
typedef message_channel<message_type, MAX_MESSAGES> channel_type;
typedef std::pair<channel_type, message_type> channel_message;
Attributes:
channel_type* channel;
Methods:
ctor()
ctor(channel_type* channel)
set_channel(channel_type* channel)
notify_observers(channel_message& message) - Implements calling the
channel_observer previously registered for this observer_channel
process() - Uses the channel_type (previously set) to process the first
message available
 if(channel && !channel->done())
   notify_observers(std::make_pair(channel, *channel))

Class: message_scheduler
Base Class: none
Purpose: To provide a sorted container (etl::multimap?) of observer_base*
classes (TValue) according to priority (TKey) specified at registration
time. Then during the process method call, the message_scheduler will loop
through each registered channel_base class and call its process method for
each channel_base whose done method returns true in order of priority.
Typedefs provided:
typedef etl::ifunction<message_scheduler, &process_func>* process_func_ptr
Classes provided:
Methods:
add_channel(observer_base*, int priority) - Adds the provided channel with
the given priority
remove_channel(observer_base*, int priority) - Removes the specified
channel with the given priority
process() - Target of process_func. Calls the process() method on each
observer_base to process channel messages

GatorQue:
I forgot the remove the process methods of channel_base and message_channel
- they are no longer necessary for my second implementation. In fact, there
doesn't need to be any functions in channel_base (or a base class
channel_base at all) anymore with the new design since we no longer need to
store a pool of message_channels, instead we store a
map/multimap/set/multiset of observer_base classes.


Ryan Lindeman

jwellbelove:
Hi Ryan,
Sorry I haven't commented yet on your ideas, but it's been a hectic week since getting back from holiday and I haven't been able to properly sit down and think about any ETL stuff. This weekend I shall be up in South Wales taking my step-daughter and her friend on a caving trip. Hopefully next week I shall be able to get a bit more *me* time and catch up with my coding.

John

GatorQue:
No worries. Have fun on your trip!

GatorQue:
I have completed the priority_queue implementation.

jwellbelove:
Excellent work!
 
I made my weekend more interesting by accidentally deleting the wrong partition
on my computer.
I'm in the process of rebuilding the OS and restoring all of my data from the
backup. D'OH!!

PR #5, opened 2015-10-06
Enum type - fix doc to_string references

0xc0170:
not certain if this goes to master or development branch.

### DIFF
--- a/enum_type.h
+++ b/enum_type.h
@@ -69,15 +69,15 @@ SOFTWARE.
 /// 
 /// direction = value;                   // **** Compilation error ****
 /// 
-/// std::cout << "Direction = " << direction.to_string(); // Prints "Direction = North"
+/// std::cout << "Direction = " << direction.c_str(); // Prints "Direction = North"
 ///\endcode
 /// If a conversion to a string is not required then the 'ENUM_TYPE' declaration may be omitted.
-/// In that case the to_string() function will return a "?". This will also be the case for any 
+/// In that case the c_str() function will return a "?". This will also be the case for any
 /// enumeration value that does not have an ENUM_TYPE entry.
 ///\ingroup utilities
 
 //*****************************************************************************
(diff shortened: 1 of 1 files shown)


jwellbelove:
Thanks.
Are you using this library in any projects or just testing it out?
What compiler are you using?

0xc0170:
Not yet using, just started reading through the files to get familiar with this library. I mainly use gcc/armcc/iar, still pre-c++1x world :tired_face: .

jwellbelove:
The pre-C++1x world is what is was aimed at.

It's based on some ideas I developed whilst writing for embedded products in the past. I got fed up of having to keep writing piles of boilerplate code around arrays to compensate for not being able to use STL containers. Also there were a load of new features in C++0x11 STL that were possible to implement in C++0x03. Replies to enquiries to Keil, IAR etc about supporting C++0x11 were "Don't hold your breath!".

I've mostly developed with VS2013 & 2015, GCC, and implemented a few classes with Keil's ARM compiler. It would be great to get feedback about getting it to compile correctly under other embedded compilers.

I've got a bit of time off between contracts, so hopefully I will be able to catch up with some of the changes I have had planned for some time. I'm currently working on an intrusive linked list which will be used in the implementation of unordered_map, set etc, which will then be used as the basis of a set of cache containers.

I was also thinking of creating a timer service class that would hook into the processors timer hardware through etl::function callbacks. Instead of a regular timer interrupt 'tick' scanning a table for due timer events it would calculate the time to the next due event and program the timer count accordingly. This would significantly reduce the number and frequency of timer interrupts.

jwellbelove:
I'm trying to create a set of tutorials on the project website. So far I've created one for etl::function.
If there are any parts of the library that you think requires a better explanation of how the item should be used and possibly why you may want to use it in the first place, then please let me know.
Thanks
John.

Issue #22, opened 2017-04-13
iflat_map::const_iterator invalidates const correctness


clegg89:
In commit 4b84cc955ef3c38e62769e689b2451de6f95a6f7 Iterator classes were added to etl::iflat_map. The const_iterator class defined inside iflat_map contains operator overloads which invalidate the const_iterator's const correctness, specifically: https://github.com/ETLCPP/etl/blob/a3d51f1c795480b0a8da50eadf1c2eaae1c8f366/src/iflat_map.h#L251-L264

Removing the `reference` and `pointer` overloads fixes the issue.

PR #23, opened 2017-04-13
Removed incorrect methods from etl::iflat_map::const_iterator class

clegg89:
Fix for issue #22


jwellbelove:
Thanks, also fixed other flat map/sets.

PR #44, opened 2017-07-19
fixed some compiler warnings

LnnrtS:
Hello,
I am starting to use the etl library on an cortex-m4 target with gcc.
There are some compiler warnings that I fixed. There are probably more of them in the whole code but I just fixed what I came across so far. With this pull request I want to find out if you are interested in having those warnings fixed and if you like this style.
Those are the relevant warnings.
 - Wsuggest-override
 - Wshadow
 - Wnon-virtual-dtor

I know that those warnings could be safely ignored for these specific cases but as this is a template library it should build cleanly no matter what warnings are enabled in the application code using it.

### DIFF
--- a/src/error_handler.h
+++ b/src/error_handler.h
@@ -56,8 +56,8 @@ namespace etl
     //*************************************************************************
     struct free_function : public etl::function<void, const etl::exception&>
     {
-      free_function(void (*p_function)(const etl::exception&))
-        : etl::function<void, const etl::exception&>(p_function)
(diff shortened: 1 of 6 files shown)


jwellbelove:
Your changes have broken the build.
You have added C++11 features. This library is intended to be compatible with C++03
You have modified the master branch.. Please push changes to a new or the development branch.

LnnrtS:
Yeah, seen that :(
Are you in general interested in fixing those warnings? If yes, I would set up msbuild and do it properly

Issue #45, opened 2017-08-21
nullptr_t not correctly named when host nullptr is available


jeremyherbert:
Hi,

I've just started to play with this library (OSX using clang), and it seems that I have found a minor issue. In `type_traits.h` on line 237, it has the following code: ```is_same<nullptr_t,```. However, this assumes that the custom `nullptr_t` class from `nullptr.h` is being used; if `nullptr` is defined on the platform and being included from `cstddef` (like it is on mine), it needs to be referenced from the `std` namespace, ie `std::nullptr`. Changing this reference on line 237 allows my code to compile.

PR #46, opened 2017-08-21
define nullptr_t for platforms with existing nullptr implementation

jeremyherbert:
Fix as per https://github.com/ETLCPP/etl/issues/45


jwellbelove:
I've decided to fix it in a slightly different way to your push, as the issue is actually related to a fix I made to try to get the ETL to compile on Arduino projects.

PR #50, opened 2017-11-01
Fixed missing changes to support strict builds.

AshleyRoll:
Hi John, seems that there was a missing few changes in the exception.h file after you built support for the strict options in your last commit.


jwellbelove:
Thanks.
There was quite a lot of changes needed to get those warnings eliminated.
I've recently changed the coding style slightly which should eliminate most of the warnings by default in future.

PR #59, opened 2018-01-09
replace usages of _DEBUG with ETL_DEBUG

apmorton:
fixes some -Wundef errors when _DEBUG was not defined


jwellbelove:
Thanks for the fix @apmorton

PR #64, opened 2018-03-09
CMake integration

scott-eddy:
This PR adds a top-level CMakeLists.txt file that can build ETL under MSVC and GNU/GCC.  I've tested it locally and both compilers build the library with no errors. 

In order to have the library compile in a stand-alone fashion an `etl_profile.h` was created that `#include`'s different headers based on pre-processor derivatives, allowing users to make a custom profile by providing the CMAKE option `-DETL_PROFILE=PROFILE_CUSTOM` and providing a new header `custom_profile.h`.  Note this will require a change in the [setup documentation](https://www.etlcpp.com/setup.html) which I'm happy to write, if there is a way to contribute to the `etlcpp.com`


src/profiles/etl_profile.h
```diff
@@ -0,0 +1,60 @@
...
+SOFTWARE.
+******************************************************************************/
+#ifndef __ETL_PROFILE_H__
+#define __ETL_PROFILE_H__
+#ifdef PROFILE_MSVC 
+  #include "msvc_x86.h"
+#elif PROFILE_GCC_GENERIC
+  #include "gcc_generic.h"
+#elif PROFILE_GCC_LINUX_X86
+  #include "gcc_linux_x86"
```
apmorton: missing `.h`

src/profiles/etl_profile.h
```diff
@@ -0,0 +1,60 @@
...
+#elif PROFILE_GCC_WINDOWS_X86
+  #include "gcc_windows_x86"
```
apmorton: missing `.h`

CMakeLists.txt
```diff
@@ -0,0 +1,65 @@
...
+  src/crc8_ccitt.cpp
+  src/error_handler.cpp
+  src/pearson.cpp
+  src/random.cpp
+  src/private
+  src/private/pvoidvector.cpp
+  )
+target_include_directories(etl
+  PUBLIC
+  src/
```
apmorton, line 31: unfortunately the current layout of etl doesn't really lend itself to being added to the include path like this, although that may be something to discuss in another issue.

For example, adding `src` to the include path makes these wonderfully generic includes valid, which will almost certainly cause conflicts in enough cases for it to be an issue:
```C
#include <doxygen.h>
#include <exception.h>
#include <function.h>
#include <hash.h>
#include <jenkins.h>
#include <log.h>
#include <platform.h>
#include <utility.h>
#include <version.h>
```

I don't know how other people are currently using ETL, but I have the git repo as a submodule, and due to the layout of the etl repo ended up adding only the parent directory of the submodule to the include path in order to prevent these kinds of conflicts.

Note this isn't particularly an issue with your PR, since there isn't really a sane thing to do here currently, but it is something to think about before cmake support gets merged.

Ideally the repo would get reorganized to look something like this:

`include/etl/*.h`
`src/*.cpp`

the cmake `target_include_directories` then becomes simply `include`, with the expectation that users would consume includes like so: `#include <etl/vector.h>`

jwellbelove:
I originally had a different directory layout but adding support for Arduino dictated the 'src' directory scheme. It may be time to re-evaluate whether sticking to Arduino library standards is still a worthwhile thing to do.

CMakeLists.txt
scott-eddy: I think that purposed layout seems reasonable.  I'm happy to include it in this PR pending @jwellbelove advice.  I don't have an Arduio to deploy on so I would not be able to personally test to see if breaking up the header location would be a problem.

src/profiles/etl_profile.h
scott-eddy: Thanks, fixed

src/profiles/etl_profile.h
scott-eddy: fixed

PR #72, opened 2018-04-22
Fix assumption that min/max macros will always be defined on a Microsoft compiler

apmorton:
`min` and `max` are not implicitly defined on MSVC, and the check performed in `integral_limits.h` can inadvertently define them in cases where they weren't already defined.

- If `Windows.h` is never included, they will not be defined.
- If `NOMINMAX` is defined before including `Windows.h` they will not be defined.

Using `push_macro` and `pop_macro` ensures they are restored to whatever state they were previously, while still allowing us to be sure they are undefined while evaluating `integral_limits.h`


apmorton:
I probably should have taken a closer look at the rest of the code base - it seems these are scattered all over the code base.  I will update this PR to fix them all when I can.

PR #84, opened 2018-07-27
Use etl/ prefixed includes in sources and tests

MarSik:
ETL is supposed to be included using #include <etl/header.h>
according to the changelog entry from Sunday, March 18, 2018.

This change updates the tests and etl source files to use
the same style.

This allows compiling etl as part of a separate project (git submodule)
using the same flags as for user code that includes etl.

Signed-off-by: Martin Sivak <mars@montik.net>

PS: I encountered many errors while compiling the test suite. Is it maintained and supposed to work with GCC 7 on Linux (Fedora 26 to be exact)?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -34,11 +34,8 @@ add_library(etl
   )
 target_include_directories(etl
   PUBLIC
-  include/etl
-  include/etl/atomic
+  include
   include/etl/profiles
-  PRIVATE
-  include/etl/private
   )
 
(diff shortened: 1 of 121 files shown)


jwellbelove:
I compile and run the ETL test suite with Visual Studio 2017. GCC 4.9.2 on Windows, and GCC 5.4.0 on Ubuntu before every push.

jwellbelove:
I'm currently trying to update my Windows MinGW and Ubuntu to GCC 8.1, but it's frustratingly hard to do.
So unbelievably convoluted!

Issue #86, opened 2018-08-14
Exchange use of deprecated/removed std::bind1st for std::bind when c++11 is used


ArekSredzki:
`std::bind1st` was deprecated in c++11 and removed in c++17. As such, etl does not build with modern libc++ versions with `-std=c++17`. Instead, `std::bind` should be used.

PR #87, opened 2018-08-14
 Exchange use of deprecated/removed std::bind1st for std::bind when c++11 is used #86

ArekSredzki:


jwellbelove:
Can you modify this to use the ETL's **ETL_CPP11_SUPPORTED** macro instead?

ArekSredzki:
@jwellbelove Done :)

PR #88, opened 2018-08-14
Fix clang warnings on etl::unordered_map, etl::unordered_multimap, and etl::unordered_set

annius:
1. Fixes extra semicolon warnings
2. Suppresses unused parameter warnings
3. Fixes unused variable warnings

### DIFF
--- a/include/etl/unordered_map.h
+++ b/include/etl/unordered_map.h
@@ -841,7 +841,7 @@ namespace etl
     ///\param position The position to insert at.
     ///\param value    The value to insert.
     //*********************************************************************
-    iterator insert(const_iterator position, const value_type& key_value_pair)
+    iterator insert(const_iterator, const value_type& key_value_pair)
     {
       return insert(key_value_pair).first;
     }
--- a/include/etl/unordered_map.h
+++ b/include/etl/unordered_map.h
@@ -781,8 +781,6 @@ namespace etl
       bucket_t* pbucket = pbuckets + index;
       bucket_t& bucket = *pbucket;
 
-      size_t s = pbuckets->size();
-
       // The first one in the bucket?
       if (bucket.empty())
       {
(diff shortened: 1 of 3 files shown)


jwellbelove:
**__attribute__** cannot be used in the library as it is not cross compiler compatible.
The **#ifdef ETL_DEBUG_COUNT** should not be necessary. What warning are you getting?
The warning for unused parameters could probably be avoided by deleting the parameter name, just leaving the type declaration.

jwellbelove:
Appologies, I can see the issue with ETL_DEBUG_COUNT now.

jwellbelove:
I think that the ETL_DECLARE_DEBUG_COUNT should include its own semicolon and have the semicolon removed from all of the instances where it is used. This would make it consistant with the other macro uses in the library.

annius:
Thanks for your feedback @jwellbelove . For some additional context, the unused variable and unused parameter warnings can be replicated by creating a file `main.cpp` with the following contents:
```
#include "include/etl/unordered_map.h"
#include "include/etl/unordered_set.h"
#include "include/etl/unordered_multimap.h"

int main() {
  etl::unordered_map<std::size_t, std::string, 10> my_map;
  etl::unordered_set<std::size_t, std::string, 10> my_set;
  etl::unordered_multimap<std::size_t, std::string, 10> my_mmap;
}
```
and then by attempting to compile with
```
clang++ main.cpp -std=c++11 -Wall -Werror -Wextra -Wpedantic -Wno-extra-semi
```

I've removed the param name as you suggested.

annius:
I've reverted the DEBUG_COUNT commit because I'd like a little more clarification on your suggestion before proceeding. 

If you add a semicolon to the definition of `ETL_DECLARE_DEBUG_COUNT` here https://github.com/ETLCPP/etl/blob/master/include/etl/debug_count.h#L44 shouldn't it be added to all of the definitions that follow that line?

annius:
https://github.com/ETLCPP/etl/pull/89

jwellbelove:
Yes, I think that the macro definitions should have semicolons added, and removed from all of the places they are used. This should fix the warning when debug count is not compiled in.

PR #102, opened 2018-10-11
Adding non-const emplace overloads for vector

jpan127:


jpan127:
1. Removed the const-ness for the parameters of `vector::emplace_back`
2. Added a variadic version for C++11 and above
3. I had to comment out the `test_factory` test and add `crc16_modbus.cpp` to the source list for the tests to pass.  I think `test_factory` is deprecated, and `crc16_modbus.cpp` is some dependency.  Some advice here would be appreciated, I do not intend to merge these changes to the CMakeLists.

PR #112, opened 2019-01-06
Update flat_map.h to handle C++98 compilation

bolry:


jwellbelove:
Looks like I need to make a check-sheet for adding rvalue references to the ETL so I don't miss anything next time.

jwellbelove:
I'm going to retert the last changes to flat_map as it has unfinished changes. It got caught up with the other changed files by accident.

PR #117, opened 2019-02-19
function.h add const to operator ()

creibetanz:
Change-Id: Id70a05ab1bfdb95499a3c6622379c8bb639f5f40

### DIFF
--- a/include/etl/function.h
+++ b/include/etl/function.h
@@ -59,7 +59,7 @@ namespace etl
     //*************************************************************************
     /// The function operator that will be overridden.
     //*************************************************************************
-    virtual void operator ()(TParameter) = 0;
+    virtual void operator ()(TParameter) const = 0;
   };
 
   //***************************************************************************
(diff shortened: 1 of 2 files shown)


creibetanz:
Until now it is not possible to do things like:
``` cpp
void free_foo(int i) {
}
struct Example {
  const etl::ifunction<int> *on_action;
  void foo(int i) {
    (*on_action)(i); //No matching function for call to object of type 'const etl::ifunction<int>'
  }
};
void call_with_one(const etl::ifunction<int> &f) {
  f(1); //No matching function for call to object of type 'const etl::ifunction<int>'
}
int main() {
  Example obj;
  const etl::function<void, int> ef(&free_foo);
  ef(); // No matching function for call to object of type 'const etl::function<void, int>'
  obj.on_action = &ef;
  obj.foo(2);
}
```

Const correctness however is sometimes quite usefull.

jwellbelove:
One other issue that comes to mind from this is to figure out how to add support for calling 'const' member functions.

PR #122, opened 2019-03-23
Fixed compiler warnings

cbladt:
Simply fixed the compiler warnings that appeared when using etl in my project. Hope that you will take it into consideration.

### DIFF
--- a/include/etl/function.h
+++ b/include/etl/function.h
@@ -53,6 +53,7 @@ namespace etl
   class ifunction
   {
   public:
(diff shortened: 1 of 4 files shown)


jwellbelove:
You can't directly use 'delete' and 'auto' in the ETL as it must remain, as much as possible', C++03 compatible.
Any C++11 and above keywords must be defined as ETL macros. (See ETL_CONSTEXPR)

cbladt:
Fair enough. I was to quick there.

Issue #121, opened 2019-03-21
How do you actually install this library ?


GeorgeFlorian:
Hello !

I am trying to use this library on PlatformIO IDE, but I'm having no luck. 
I'm using it with `lib_deps = 930` inside `platformio.ini`.

I've read here: https://www.etlcpp.com/setup.html and I made `etl_profile.h` and I put it inside `include` folder in the project's folder:
 
```
#ifndef __ETL_PROFILE_H__
#define __ETL_PROFILE_H__

#define ETL_THROW_EXCEPTIONS
#define ETL_VERBOSE_ERRORS
#define ETL_CHECK_PUSH_POP

#endif
```

With:
```
#include <etl_profile.h>
#include <deque.h>
```
compiling gives me the following error:
```
compilation terminated.
*** [.pioenvs/esp32doit-devkit-v1/lib14e/Embedded Template Library_ID930/crc16.cpp.o] Error 1
compilation terminated.
*** [.pioenvs/esp32doit-devkit-v1/lib14e/Embedded Template Library_ID930/binary.cpp.o] Error 1
In file included from .piolibdeps/Embedded Template Library_ID930/src/c/ecl_timer.c:32:0:
.piolibdeps/Embedded Template Library_ID930/src/c/ecl_timer.h:34:22: fatal error: ecl_user.h: No such file or directory
```

Without `#include <etl_profile.h>` it says:

```
compilation terminated.
*** [.pioenvs/esp32doit-devkit-v1/lib14e/Embedded Template Library_ID930/binary.cpp.o] Error 1
In file included from .piolibdeps/Embedded Template Library_ID930/src/crc16.cpp:33:0:
.piolibdeps/Embedded Template Library_ID930/src/platform.h:49:25: fatal error: etl_profile.h: No such file or directory
```

jwellbelove:
You don't need to include the **etl_profile.h** explicitly. It's already included by **platform.h**.
If **platform.h** can't see it then the header path must be included in the compiler's search paths.
**platform.h** should not be in the ETL's **src** sub-directory.
**ecl_timer.c** is only required if you are writing a pure C application; it's not really part of the C++ ETL proper (It's just an extra I threw in for C programmers)

The cpp files were removed from the library a while ago when it went to a 'header only' implementation.
You can delete all of the cpp files in **src**.

maxgerhardt:
[...]

PR #124, opened 2019-04-15
PlatformIO support (WIP)

ristomatti:
This is a WIP PR related to discussion at #121 and should not be merged as such. If you use labels, please add a "WIP" label on this one.

After some fiddling I managed to create a suitable `library.json` file based on instructions here: https://docs.platformio.org/en/latest/librarymanager/config.html

I mapped the values in `library.properties` to best matching fields. There's bunch of other fields you would likely want to add like `keywords`, `export` (to filter out folders like `temp`), `examples` (to possibly add only some of the examples). Also I know ETL isn't Arduino specific and it'd likely work for other platforms also, I added only the ones I have now confirmed working.

I am able to compile the QueuedMessageRouter example with this `platformio.ini` configuration:
```ini
[env:esp32dev]
board     = esp32dev
platform  = espressif32
framework = arduino
lib_deps  = https://github.com/ristomatti/etl.git#platformio-support
```
And this `include/etl_profile.h`:
```cpp
#ifndef __ETL_PROFILE_H__
#define __ETL_PROFILE_H__

#define ETL_VERBOSE_ERRORS
#define ETL_CHECK_PUSH_POP

#endif
```

### DIFF
--- a/library.json
+++ b/library.json
@@ -1,0 +1,27 @@
+{
+  "name": "Embedded Template Library",
+  "version": "10.21.2",
+  "authors": {
+    "name": "John Wellbelove",
+    "email":  "<john.wellbelove@etlcpp.com>"
+  },
+  "homepage": "https://www.etlcpp.com/",
+  "license": "MIT",
+  "description": "A C++ template library tailored for embedded systems. Requires some support from STL. See http://andybrown.me.uk/2011/01/15/the-standard-template-library-stl-for-avr-with-c-streams/ or https://github.com/mike-matera/ArduinoSTL.git for Arduino.",
+  "repository": "https://github.com/ETLCPP/etl.git",
+  "platforms": [
+    "espressif8266",
+    "espressif32"
+  ],
+  "frameworks": "arduino",
+  "build": {
+    "flags": [
+      "-I include"
+    ],
+    "srcFilter": [
+        "-<src/*>",
+        "-<temp/*>",
+        "-<test/*>"
+    ]
+}
+}


ristomatti:
The version number is obviously wrong as it's the last version to officially support Arduino. I could not find the current version with a quick glance though?

ristomatti:
Would it make sense to you if there was a `etl_profile.h` included as a build parameter here which would have a set of #ifdefs checking for Arduino availability and within it #ifdefs for various Arduino cores that would #include suitable settings for the given platform? This way it would be possible to remove the awkward step of figuring where to document this requirement.

For the multiplarform support some hints could likely be found by checking how bblanchon has done it for ArduinoJson https://github.com/bblanchon/ArduinoJson

Yet another thing I noticed is that the comment on ETL's homepage about Arduino is outdated in the sense that these days it does not imply AVR. For instance the cores for Espressif chips include STL support. ESP32 core even nas an RTOS there as its built on top of ESP-IDF.

ristomatti:
@jwellbelove Since you decided to merge this right away I assume you'll be adding more stuff to `library.json` later on? You'll find info on adding new library on their index here: https://docs.platformio.org/en/latest/librarymanager/creating.html

But awesome, thanks!

jwellbelove:
I was going to update it soon, firstly with the latest version number. (I assume that works for your setup). I need to get PlatformIO setup at home for my ST Nucleo 401 board to test it out. Once I'm satisfied that works I'll register it as a library.

ristomatti:
But please also add some of the metadata (keywords etc) so it'll be easier to find. I would consider also specifically listing the examples that apply to the frameworks listed as supported in `library.json`.

Should I create a feature request (or a PR if I manage to) on simplifying the installation, I'm referring to the mandatory `etl_profile.h` definition step which IMO is quite confusing.

Another way to tackle the issue would include a hint of this prerequisite as a comment on top of the examples that'll be prominently listed on the library index.

jwellbelove:
I did think about reintroducing some sort of 'auto' profile that interrogates your platform. I originally setup the library this way, but it started to get very messy, and hard to get correct for every possible platform. It was also very hard to test without owning *all* of the possible development platforms. In the end I decided that the person using the ETL knew best about about what device, OS and compiler they were using. Maybe some more words on the subject would be in order.

jwellbelove:
I've created a branch from master to use for additions to PlatformIO setup.
feature/PlatformIO

ristomatti:
> It was also very hard to test without owning all of the possible development platforms. In the end I decided that the person using the ETL knew best about about what device, OS and compiler they were using.

I can definately understand that but at least when it comes to Arduino, I would doubt most of the people downloading libraries know pretty much anything about the boards and compilers. PlatformIO also does a good job hiding that as it's main idea is to be able to use a single tool regardless of the development platform or framework. Basically attempting to abstract away most of the work associated with installing the toolkit, setting up compilers and debuggers into preconfigured settings behind a single config file. The predefined settings and build flags can then be overridden etc.

Naturally they can only do that for the platforms and framework they currently support but for a hobbiest like me, they have pretty much everything covered. From the supported frameworks, I'm currently only interested in Arduino as it simplifies things enough so that it's possible for me to get encouraging results with my limited time and lack of indepth knowledge in microcontroller inner workings.

From their the Arduino supported boards I have available for testing the classic Atmel's ATMEGA328P and ATMEGA2560, Espressif's ESP8266 and ESP32 and from STM only STM32F103C8T6. I can see if I can get my simple test code to compile and run on each. It's just a modified version of the QueuedMessageRouter example, with `etl::unordered_map` mashed in and filled and read randomly from a list of `const char*` keys and `unsigned long` values from the `millis()` function. So it clearly is not a very comprehensive test but at least I'll see if it will compile at all.

ESP32 I've already tested and ESP8266 at least compiled but both Arduino cores have the STL available. I could do a quick test know actually, so I'll follow up with the results.

ristomatti:
@jwellbelove A conclusion of this ~6 hour exercise of brute forcing the library to compile on the boards I had available in the form of a PR #128. 90% of the time was spent on STM32F103 which is the least popular or mature platform to be used with an Arduino, supported only by a small community.

jwellbelove:
Thanks for all of your hard work! Much appreciated.

jwellbelove:
I had a go at using PlatformIO for my ST32F401 Nucleo board with the Queued message router project.
It seemed to work out OK.

ristomatti:
Great! Thanks for the update. Are there some specific templates that you could mention worth trying out to further test the compatibility?

jwellbelove:
I used the Arduino platform. I will try to use the STM CMSIS platform next.

ristomatti:
I assume you meant "framework" (in `platformio.ini` terminology)? That will be interesting. I've got no experience of the other platforms so they'd be difficult for me to test.

What are your first impressions on PlatformIO as an professional embedded developer? In case you didn't notice - you can also use other editors (although the level of integration varies). I've tested it with CLion for example and it works pretty nicely. Have to first create the required CMake files running `pio init --ide clion` in the project folder.

jwellbelove:
Setting it up for Arduino was pretty painless, I'm hoping that CMSIS will be as easy.
I've not tried playing with CMake very much yet. If PlatformIO can automate some of that I would see that as a plus.
As a professional programmer, frameworks like CMSIS and similar are the most important. If it can get you started quickly and even allow appications to be easily ported between frameworks, then it looks very promising.
On the other hand, programmers can also be very conservative and will avoid unnecessary change or risk (especially when project deadlines are looming), so even a great idea can sometimes fail to get traction.

ristomatti:
Please let me know how it went when you have time to do this. I've currently got no intention on progressing to frameworks targeted to professionals but I'm just purely curious if and how well PlatformIO manages to do this. I've been consering of someday taking test plunge to Espressif's ESP-IDF (RTOS for their ESP32 micro) which is also supported by PIO.

PR #131, opened 2019-06-14
Updated doxygen file

khoitd1997:
Hi, I was browsing the source code and try to generate Doxygen docs when I noticed that there are a couple of values that are hardcoded, resulting in the things not working. I changed a couple of Doxygen config so it works and looks nicer.


jwellbelove:
Thanks

Issue #151, opened 2019-09-22
Improve platformio support


puzrin:
PIO support is outdated and buggy. Here is description of steps, required to make things work.

## PIO config

Config is restricted without need to few platforms and miss timer-related feature. Version in registry is not updated.

1. [Here](https://github.com/puzrin/etl/commit/caa5da8ec4c4e1c87e56e7eac11fd13c228a3561) is list of minimal mandadory changes. At least, that will make library useable via git links, as dependency.
   - May be content of other fields should be improved. I did only minimal possible changes to make things buildable.

2. PIO registry crawler is frozen by outdated version number in `library.properties` (arduino config). Versions in both `library.properties` AND `library.json` MUST be updated and be in sync with `include/etl/version.h`
   - After config updated, at least one new tagged release required, to kick PIO crawler.
   - I strongly recommend add CI test to make sure versions in all files are in sync.

3. IMO, announcing package in pio registly as "Embedded Template Library" is "too arduinish" :). I'd suggest change field "name" to "etl". Did not updated this in my commit, because don't know your opinion. IMO [this line](https://github.com/puzrin/etl_check/blob/master/platformio.ini#L21) in app config looks more nice than long name with space and letters of different sizes.

## PIO demo

https://github.com/puzrin/etl_check - something like this should be added into examples. Users should see, everything is buildable in PIO.

Note, `src/` folder contains garbage. I just needed something to make sure build pass and my fixes of etl config are correct. Required files from `include/` are specially done to reuse existing configs with minimal intrusion.

`platformio.ini` temporary refers to cloned repo until mainstream updated. If you copy it to mainstream's samples - don't forget to fix.

I'm not C programmer. So i can't make good code samples, but can care about fixing PIO issues. Use refered repo as template.

## Dev notes

If you need to test library config changes - clone repo to `lib/` folder of demo AND remove `lib_deps` from demo config (`platform.ini`). All subfolders of `lib/` are automatically used as dependencies.

---

@jwellbelove what do you think? I'm interested to get etl useable with pio asap. Of cause, i could use fork with fixes, but fixed mainstream is preferable.

I did not created PR because some of described changes need to be done at your side, and some depend on your opinion. Let me know if any help required.

puzrin:
https://github.com/puzrin/etl_check/commit/581a705638ac4c4dda22b1650ac4963ff49a3b66

[...]

PR #153, opened 2019-09-26
Add platformio demo

puzrin:
Ref: #151.


puzrin:
I see "Merging is blocked message". If PR should be done in different way - let me know.

PR #174, opened 2019-11-30
Add NO_STL std::reverse implementation

grigorig:
Follows the example implementation on [1].

[1] https://en.cppreference.com/w/cpp/algorithm/reverse


jwellbelove:
I'm currently revisiting the ETL_NO_STL option. I'm editing code so that the unit tests are able to be run with this setting, to ensure that there are no missing alternate STL functions and classes.
The missing reverse is one I found this morning. I hope to get the updated code pushed this weekend.

PR #175, opened 2019-12-03
Make string

rolan-reznik:
make_string utility function.
It provides possibility to safely create etl::string from string literal.
Example:
```
    auto testString = etl::make_string("test567");

    std::cout << "testString: " << testString.data() << std::endl;
    std::cout << "testString: " << testString.max_size() << std::endl;
    std::cout << "testString: " << testString.size() << std::endl;
```

Output:
```
testString: test567
testString: 7
testString: 7
```


rolan-reznik:
Sorry, this pull request have some other commits. Will create another one.

PR #176, opened 2019-12-03
Added utility function etl::make_string 

rolan-reznik:
make_string utility function.
It provides possibility to safely create etl::string from string literal.
Example:
```
auto testString = etl::make_string("test567");

 std::cout << "testString: " << testString.data() << std::endl;
 std::cout << "testString: " << testString.max_size() << std::endl;
 std::cout << "testString: " << testString.size() << std::endl;
```
Output:
```
testString: test567
testString: 7
testString: 7
```

### DIFF
--- a/include/etl/cstring.h
+++ b/include/etl/cstring.h
@@ -251,6 +251,15 @@ namespace etl
     }
   };
 #endif
+
+  //***************************************************************************
+  /// make string from string literal or char array
+  //***************************************************************************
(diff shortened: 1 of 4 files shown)


rolan-reznik:
> Would it be possible to have one make_string, overloaded on character type?

I thought about that. But I have not found simple solution. To do that return type (string) somehow should be deduced from argument (char) type. Probably with some template tricks it can be achieved.

Issue #171, opened 2019-11-25
unsupported crc16(XModem) standart


danielklioc:
Hello,

I am using etl in crc calculation.
I use crc16_ccitt with intial value 0xFFFF. But I need support for crc16_ccitt with intial value 0x0000.

I have created copy of policy with intial value 0x0000 and can create pull request.
Is it proper solution?

jwellbelove:
Does this CRC variant with an initial value have a specific name i.e. CRC16 DNP, CRC16 DECT etc
If so, then create a policy and frame check sequence derived class with this name.

hoo2:
I believe it is CRC16-CCITT poly:0x1021 (or 0x8408 for reverse order). It is used for X.25, V.41, HDLC FCS, XMODEM, Bluetooth, PACTOR, SD, DigRF and many others...

danielklioc:
@jwellbelove yes, it has specific name CRC-16/XMODEM. Is name crc16_xmodem and crc_policy_16_xmodem ok?
@hoo2 crc16_ccitt initial value 0xFFFF, for xmodem 0x0000, but poly is same. This is why I need different policy.

jwellbelove:
There are a few solutions to this.

1. Add the ability to give **any** CRC a specified initial value.
2. Create a CRC and policy for each and every named variant of CRC16 with zero as an initial value.
3. Modify the CRC16 policy so that can be given its initial value and create a top level derived class for each named variant from it.
4. Create a CRC16 that uses zero as its initial value and typedef all of the variants from it.

I sort of have a preference for #42  as it maximises code reuse.

jwellbelove:
Yes, create could create a crc_policy_16_xmodem, and a crc16_xmodem derived class.
If other CRC variants use exactly the same algorithm then they could maybe be typedef'd from it (though it may cause issues if you want to overload on the CRC type)

jwellbelove:
Were you planning to do a pull request in the near future for this?

danielklioc:
I have written another implementation with template and typedef, to have option to choose ccitt or ccitt_xmodem standart. Cannot finalize it because of compile error.

Could you help to find issue?

my changes in crc16_ccitt.h
`
...
namespace etl
{
  //***************************************************************************
  /// CRC16 CCITT policy.
  /// Calculates CRC16 CCITT using polynomial 0x1021
  //***************************************************************************
  template<uint16_t I>
  struct crc_policy_16_ccitt
  {
    typedef uint16_t value_type;

[...]

PR #180, opened 2019-12-07
Added option of crc16 ccitt_xmodem standart

danielklioc:
Added option of crc16 ccitt_xmodem standart. (#171)


jwellbelove:
Thanks for doing a pull request for this. It inspired me to take another look at the CRC generation routines and refactor them for better code reuse and size.

hparracho:
@jwellbelove I have a small generic CRC class I use in the own lib.
If you're interested you can adapt it to etl. Feel free you use it as you want.
https://github.com/hparracho/Xarmlib/blob/master/include/api/api_crc.hpp

PR #186, opened 2020-01-13
Use builtin exchange if compiling for <C++14.

denravonska:
Use the provided version of `exchange` if building for less than C++14. This fixes a compile error when using C++11.


denravonska:
> Change the `__cplusplus < 201300` to the ETL style `!ETL_CPP14_SUPPORTED`

Fixed.

PR #201, opened 2020-03-18
Made template compatible with cpp11

Mipsters:
If you want me to fix the tests too, please tell me which file it is in  
I have yet to get into the way tests are run in this project


jwellbelove:
The tests for **exchange** are in **test/test_utility.cpp**

Add a copy of **TEST(test_exchange)** (lines 194 to 204), give it a new name (say, **test_exchange_different_types**) and change one of the variables to a different integral type, such as char. This should force the compiler to choose the two template parameter version of **exchange**.

PR #239, opened 2020-07-16
error_handler.h: Change C++11 type aliases to typedefs

balintkissdev:
Fixing C++98/C++03 compatibility issues with error_handler.h by using `typedef` instead of C++11 type aliases.


jwellbelove:
My bad! I copied the code from etl::delegate, which is a C++11 only template.

PR #256, opened 2020-08-04
Builds now on Linux with clang-9

finger42:
* renamed CHECK_EQUAL with CHECK
* removed artifact stashing due to an limitiation for upload filesize to github
* reorderd yml file to call cmak and make in the same step, so that CC and CXX is
  the same for etl and the tests


test/test_atomic_gcc_sync.cpp
```diff
@@ -43,7 +43,7 @@ namespace
       std::atomic<int> compare;
       etl::atomic<int> test;
 
-      CHECK_EQUAL(compare.is_lock_free(), test.is_lock_free());
```
jwellbelove, line 46: Does clang-9 really choke on comparing two bools?!

jwellbelove:
Is it absolutely necessary to change **all** of the `CHECK_EQUAL` to `CHECK`?
Some of the tests are just comparing integral and pointer types. I prefer `CHECK_EQUAL` to `CHECK` where ever possible as it returns a more useful result on failure. If CHECK_EQUAL does not compile then I would prefer a local comparison operator to be defined in the test cpp file to allow its use.

finger42:
> Is it absolutely necessary to change **all** of the `CHECK_EQUAL` to `CHECK`?
> Some of the tests are just comparing integral and pointer types. I prefer `CHECK_EQUAL` to `CHECK` where ever possible as it returns a more useful result on failure. If CHECK_EQUAL does not compile then I would prefer a local comparison operator to be defined in the test cpp file to allow its use.

You are right.

I will change that and start another PR.

PR #262, opened 2020-08-28
Reference wrapper fix for pure abstract classes

rolan-reznik:
This change allows to use reference_wrapper with pure abstract classes.


jwellbelove:
I've pulled this into a hotfix branch, but in doing so I've noticed that the ETL's reference_wrapper API does not match that of the STL.
I will rectify this before merging.
It will mean that your change will disappear, as std::reference_wrapper does not allow values to be assigned using the assignment operator. To do that you must use get() or operator T& () const

rolan-reznik:
>
> I've pulled this into a hotfix branch, but in doing so I've noticed that the ETL's reference_wrapper API does not match that of > the STL.
> I will rectify this before merging.
> It will mean that your change will disappear, as std::reference_wrapper does not allow values to be assigned using the
> assignment operator. To do that you must use get() or operator T& () const

Yes, I have seen that, but was not sure how to proceed with it.
Thank you for the fast response.

PR #275, opened 2020-09-26
Fix faulty assert in `list<int,0>::available`

bolry:
Prevent zero sized `list<int, 0> x;` from always asserting.

### DIFF
--- a/include/etl/list.h
+++ b/include/etl/list.h
@@ -291,8 +291,15 @@ namespace etl
     //*************************************************************************
     size_t available() const
     {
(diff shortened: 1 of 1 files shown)


jwellbelove:
I've merge your 3 PR changes into a hotfix branch.
I want to make a few minor changes, as I think that the functions _should_ assert if a pool has not been set for the etl::list
I also think that the calculation for available may be incorrect if a shared pool is being used.

bolry:
That's OK. But I still wrote an [issue](https://github.com/ETLCPP/etl/issues/276) regarding how I feel about `full` and `available` for the `etl::list<int, 0>` corner case.

PR #277, opened 2020-09-28
introduce bip buffer spsc atomic

bku-sue:
This is a blueprint for a single producer single consumer bi-partite buffer implementation, which is very beneficial to have in embedded context as it provides a DMA-friendly quasi-circular buffer. This branch is certainly not ready for merging, I just wanted to get the ball rolling to get this feature in. So let's discuss how to implement this properly. E.g. it's currently using array_view with mutable option, which should most likely be replaced.

Based on the works of Andrea Lattuada and James Munns:
https://blog.systems.ethz.ch/blog/2019/the-design-and-implementation-of-a-lock-free-ring-buffer-with-contiguous-reservations.html
Whose design was inspired by Simon Cooke:
https://www.codeproject.com/Articles/3479/The-Bip-Buffer-The-Circular-Buffer-with-a-Twist

### DIFF
--- a/include/etl/bip_buffer_spsc_atomic.h
+++ b/include/etl/bip_buffer_spsc_atomic.h
@@ -1,0 +1,585 @@
+///\file
+
+/**
+ * @note Based on the works of Andrea Lattuada and James Munns:
+ * https://blog.systems.ethz.ch/blog/2019/the-design-and-implementation-of-a-lock-free-ring-buffer-with-contiguous-reservations.html
+ * Whose design was inspired by Simon Cooke:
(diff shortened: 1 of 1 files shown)


jwellbelove:
That looks interesting. I sounds similar to something I did years ago buffering data with system that 'swapped' buffers when one go t full. One was being filled whilst the other was being processed.

I've merged the PR into a new feature branch `feature/introduce-bip-buffer-spsc-atomic` where it can be worked on further.
Hopefully you have access to this branch. Let me know if you have any difficulties pulling or doing a PR to there.

PR #293, opened 2020-10-26
Keil 5 fixes

rolan-reznik:


rolan-reznik:
__is_convertible_to documentation: https://www.keil.com/support/man/docs/armcc/armcc_chr1359124947145.htm

Issue #312, opened 2020-12-03
Unknown new-header on Arduino megaAVR Boards


HpLightcorner:
**Issue**
When compiling the ETL for the Arduino Nano-Every, it fails due to a missing include in `<memory.h>`. For some reason, `avr-gcc` does not handle `<new>`, only `<new.h>`.

**Suggested Solution**
change `#include <new>` to `#include <new.h>` and `#include <memory>` to `#include <memory.h>`. I am not sure if this will affect other platforms?

I will make a PR - Tests are running fine on avr-gcc with Arduino Nano-Ever - lets see if other platforms work?

HpLightcorner:
Okay, learned something new [here](https://stackoverflow.com/questions/301586/difference-between-using-includefilename-and-includefilename-h-in-c). 

I understand that `#include <new>` is a good idea to make all features of the header available, so the best solution might be to add another define like `#define ETL_INCLUDE_C_STYLE_NEW_HEADER ` to do the following:

```
#ifdef ETL_INCLUDE_C_STYLE_NEW_HEADER
#include <new.h>
#else
#include <new>
#endif
```

HpLightcorner:
Some further information: The Arduino-Core for megaAVR has a custom `new.h` and `new.cpp` file, it looks like the reason why ETL compiles with `#include <new.h>`

I had a look at ETL and `#include <new>` seems to be very common in several files. Isn't that a somehow an unwanted dependency?

jwellbelove:
Thanks for the suggestions. I'll take a look at that.
Arduino compatibility has always been an issue for the ETL, as it leaves out parts of the standard library and headers that other platforms provide.

HpLightcorner:
@jwellbelove have you seen my PR on that one?

jwellbelove:
OK, just seen it. I'll check it out.

jwellbelove:
I've been trying this code for Arduino 1.8.13

```
#include <new>

int i;

void setup() {
  // put your setup code here, to run once:
  ::new ((void*)&i) int(1);
}

void loop() {
  // put your main code here, to run repeatedly:
}
```
I haven't been able to get it to complain about `#include <new>` yet.

jwellbelove:
I'm looking at automatic detection (if not explicitly defined) for the <new> header, in `platform.h`

```
// Figure out if we have a standard library <new> header available.
#if !defined(ETL_USING_STD_NEW)
  #if defined(__has_include)
[...]

PR #313, opened 2020-12-03
Include new on megaAVR Boards (Arduino Nano Every)

HpLightcorner:
Addressing #312


HpLightcorner:
Okay, that is interesting, seems like Arduino makes a Workaround necessary here? good that you set up all the tests!

HpLightcorner:
@jwellbelove I tested this branch on some Arduino-Targets. From my perspective this could be a nice solution as management of the '#include <new>' comes into one central file in the private folder. However, it is necessary to check if there is already a builtin version of 'placement new'. Not sure how to do this compiler-independent.

HpLightcorner:
@jwellbelove I think  `ETL_NO_NEW_HEADER_AVAILABLE` might be the most easy to use solution for the given problem - if someone has to use a platform with these specific needs the adjustment in `etl_profile.h` seems to be the most easy way. And ETL does only use placement new operator as far as I have seen, so everything should be fine.

HpLightcorner:
image

jwellbelove:
I've pulled it into a hotfix branch to check over the changes.

jwellbelove:
I've pulled it into a hotfix branch to check over the changes.

PR #338, opened 2021-02-17
fix bug in ibitset find_next method

Chiraffollo:


test/test_bitset.cpp
```diff
@@ -1012,6 +1012,21 @@ namespace
       CHECK_EQUAL(4U, data.find_next(true,  1));
     }
 
+    //*************************************************************************
+    TEST(test_find_next_multi_byte_bitset)
+    {
+      etl::bitset<16> data;
+      data.set(12);
+      data.set(22);
```
jwellbelove, line 1020: 22 is larger that the size of the bitset. Did you mean to do that?

jwellbelove: I'm nearly ready to push my fix, but I'll add your test too, when it's fixed.

Chiraffollo: Sorry, that's my fault.
Simple rule: don't change anything between testing and commit.

Chiraffollo: fixed

Chiraffollo (commented):
test is fixed

PR #340, opened 2021-02-28
Export headers as system includes

aboseley:
Gcc doesn't apply warning to system includes paths. This change
allows a project to use high warning levels without getting swamped by ETL
issues.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,7 +9,7 @@ option(NO_STL "No STL" OFF)
 
 add_library(etl INTERFACE)
 
-target_include_directories(etl INTERFACE include)
+target_include_directories(etl SYSTEM INTERFACE include)
 
 target_compile_definitions(etl INTERFACE)
 


jwellbelove:
I've had a report from one user that the addition of SYSTEM to the CMakeLists,txt file causes link errors for one of their builds.
They've had to remove it from their copy.
Any thoughts?

Their email:-

After our last conversation I started using the new state chart build in my code and it’s been working pretty well. However I just noticed my bootloader build, which uses much of the same code, started getting build errors. I did some digging and narrowed it down to commit hash 47102c7a in the ETL library. I’m using CMAKE to build my code, and this commit modified the CMakeLists file. Do you have any thoughts on what the problem might be and how to possibly fix it?

 For reference here is the build error, I verified that the only code you can see in the error is from the ETL itself so there’s no concerns of proprietary information:

[ 47%] [32mBuilding CXX object astro_src/saturn_bootload_app/CMakeFiles/saturn_bootload_app.dir/src/fwupdate_saturn_sm.cpp.obj[0m

In file included from [01m[K/astro/external/etl/include/etl/type_traits.h:69:0[m[K,

                 from [01m[K/astro/external/etl/include/etl/algorithm.h:43[m[K,

                 from [01m[K/astro/external/etl/include/etl/array.h:38[m[K,

                 from [01m[K/astro/external/etl/include/etl/state_chart.h:36[m[K,

                 from [01m[K/astro/astro_src/saturn_bootload_app/hdr/fwupdate_saturn_sm.hpp:15[m[K,

                 from [01m[K/astro/astro_src/saturn_bootload_app/src/fwupdate_saturn_sm.cpp:12[m[K:

[01m[K/usr/bin/gcc-arm-none-eabi-7-2017-q4-major/arm-none-eabi/include/c++/7.2.1/type_traits:68:3:[m[K [01;31m[Kerror: [m[Ktemplate with C linkage

   [01;31m[Ktemplate[m[K<typename _Tp, _Tp __v>

   [01;31m[K^~~~~~~~[m[K

GatorQue:
A quick google search on the error seems to indicate that this might have something to do with a template appearing in the middle of an "extern c" block like this:
`
#ifdef __cplusplus
extern "C" {
#endif

#include <vector> // could generate this error

template <typename T> // would also generate this error
struct MyExample
{
    T data;
};

#ifdef __cplusplus
}  // end extern "C"
#endif
`
I haven't looked at the ETL sources recently, is it possible that fwupdate_saturn_sm.cpp or .hpp is trying to `#include` some ETL file within an "extern c" block? Or is there a misplaced "extern c" in the ETL sources somewhere? I'm not sure why this would only be exposed by marking ETL as a "system" include file though.

jwellbelove:
Thanks, I'll share your thoughts with the guy that had the issue, though he says they have found 'another workaround'.

aboseley:
It will change the order in which the includes are searched.  I wonder if its exposing a conflicting name in the ops code base.

aboseley:
This look to be the line
https://github.com/gcc-mirror/gcc/blob/586a0829dc38626ca27b79e5dd689fae32fc8ca8/libstdc%2B%2B-v3/include/std/type_traits#L69

PR #341, opened 2021-03-08
Remove executable permission on C++ source files

bolry:
Some headers and test files have executable permission. This should only be found on script files with shebangs.


jwellbelove:
Interesting! Being as I don't normally work in Linux/Unix (except to run CI locally) I have no idea how those files got executable permission.
I had to look up what a 'Shebang' was. It's not a phrase that is in common usage in the embedded work I do. Pretty much all projects I've worked in the last 40 years have not used Unix/Linux, neither as a development or as a target platform. 
(Bare metal / RTOS)

jwellbelove:
I run the CI under WSL, so maybe that's where it happened.

PR #342, opened 2021-03-08
Partial fix to compiling with C++98/03

bolry:
Some errors fixed when trying to compile the ETL library in C++98 mode.


bolry (commented):
Some fixes handle undefined `size_t` here fixed with include of `stddef.h`.
One didn't have `swap` here fixed with "algorithm.h"
Some things ignored with #if required C++11 support.

PR #351, opened 2021-03-11
Add free-standing sanity tests for C++03/11/14/17

bolry:
The tests all currently failed due to outstanding issues in the current release (v20.6.0).

Some of the tests need to be adjusted to remove deprecated files for any specific version of C++.

Should later be integrated into CI.


jwellbelove:
Shouldn't the CMakeLists.txt file for `test/cpp03sanity/CMakeLists.txt` refer to `CXX_STANDARD 03` rather than `CXX_STANDARD 98`?

bolry:
No, the only valid C++ version in CMake currently are 98/11/14/17/20 (using latest CMake). Haven't seen any compiler that separates C++98 from C++03 it being *mostly* a bugfix of 98.

PR #353, opened 2021-03-14
Feature/make sanity test also run with clang compiler

bolry:
Also make sure <stdint.h> macros are fully used without considering from what version of C99/11 the header file comes from.
Also, building library as object files to bypass building an uninteresting library file that will not be used.

`--cmake-clean-cache` does not change the compiler used from a previous build cache. This needs one directory cache for each compiler. This can be seen by running `make VERBOSE=1` to see the compile commands.


jwellbelove:
Thanks for all your help. I started the ETL as a fun project to share ideas, but now it's almost turned into a full time job!

PR #356, opened 2021-03-23
Add moveable parameter support to delegate

Chiraffollo:
Currently, it is not possible to forward parameters that can't be copied but only moved as parameter to a delegate call.
Therefore, I suggest to use std::forward to enable this feature. #357


jwellbelove (changes requested):
Change `std::forward` to `etl::forward` for compatibility with the No STL option.

PR #374, opened 2021-04-22
Add hierarchical FSM capabilities to the FSM.

jovere:
This adds the hierarchical state machine capabilities as an inherited class to the already existing FSM.

The states needed modification to add pointers to the parent and child states in the FSM, as well as a check to call the parent if it exists.

In order to work properly, states with no change need to return `ifsm_state::NO_CHANGE` rather than their given state id.
Otherwise, when an event isn't handled, it will return the parent state rather than the active state.

Also, in this implementation, a state cannot return a different state during the `on_enter_state()` function, as it would cause complexity with the enters and exits through the hierarchy.  An assert has been added to check for that.

### DIFF
--- a/include/etl/fsm.h
+++ b/include/etl/fsm.h
@@ -68,6 +68,7 @@ SOFTWARE.
 namespace etl
 {
   class fsm;
+  class hfsm;
 
   /// Allow alternative type for state id.
 #if !defined(ETL_FSM_STATE_ID_TYPE)
@@ -78,6 +79,13 @@ namespace etl
 
   // For internal FSM use.
   typedef typename etl::larger_type<etl::message_id_t>::type fsm_internal_id_t;
+
+  template <typename TContext, typename TDerived, const etl::fsm_state_id_t STATE_ID_,
+          typename T1 = void, typename T2 = void, typename T3 = void, typename T4 = void,
+          typename T5 = void, typename T6 = void, typename T7 = void, typename T8 = void,
+          typename T9 = void, typename T10 = void, typename T11 = void, typename T12 = void,
+          typename T13 = void, typename T14 = void, typename T15 = void, typename T16 = void>
+  class fsm_state;
 
   //***************************************************************************
   /// Base exception class for FSM.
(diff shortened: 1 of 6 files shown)


jwellbelove:
I've just been playing with the HFSM code.
The one small issue that is slightly bugging me is the fact that ordinary FSMs expose the API required by the HFSM and adds three pointers to the FSM states.

jwellbelove:
Is it necessary to have both `p_default_active_child` _and_ `p_active_child`?
It seems that **both** of them will either contain ETL_NULLPTR _or_ the address of a state.

Or am I missing something?

jovere:
> I've just been playing with the HFSM code.
> The one small issue that is slightly bugging me is the fact that ordinary FSMs expose the API required by the HFSM and adds three pointers to the FSM states.
> 
> I think I can probably live with that, unless I can think of a simple solution.

Yeah, that was the one thing that kinda bothered me as well.  I would have inherited from the ifsm_state, but with the CRTP, I couldn't see a way to have all the generated specializations use the correct format without a giant copy paste.  It might be possible to do it with an `std::is_base_of` on the state machine passed to the state templates.  However, that's not a feature available for C++03.

jovere:
> Is it necessary to have both `p_default_active_child` _and_ `p_active_child`?
> It seems that **both** of them will either contain ETL_NULLPTR _or_ the address of a state.
> 
> Or am I missing something?

The default child is used for when starting in a state that should automatically go into if that state is the specified transition.  For example, in the test, I have it transition to the "Running" state, and when the "do enters" for the running state occurs, it automatically transitions into the "WindingUp" state, since that is the default child.  The `p_default_active_child` never changes, while the `p_active_child` changes based upon the present active state.

jwellbelove:
OK, I understand.

jwellbelove:
What do you think about removing add_initial_child and making add_child set the default to the first child state that is added.
This order is enforced at the moment anyway due to the asserts.

      running.add_child_state(windingUp);
      running.add_child_state(atSpeed);
      running.add_child_state(windingDown);
      
      becomes

jovere:
I guess that seems reasonable, as long as we document it to show that the first call will set the default state.  I'm assuming add child would be something along the following pseudo code?

```
add_child ( child )
{
  if (default != nullptr)
  {
    default = child
  }

  /* Do rest of add_child here */
}
```

jwellbelove:
That's what I was thinking of.

In the standard FSM, the first pointer in the state list is the default initial state too.

jovere:
Seems fine to me.

PR #378, opened 2021-05-19
Add container_type type definition to etl::queue to better match std::queue

dingari:
In my project I'm trying to pass `etl::queue` to a library function that expects a "queue-like" container. The library is expecting `etl::queue` to have a type defintion for `container_type`, just like `std::queue` - and the compilation fails with the error 

```
no type named 'container_type' in 'etl::queue
```

Adding this typedef to the `etl::queue` class solves the issue, and the class is more in-line with `std::queue`.

### DIFF
--- a/include/etl/priority_queue.h
+++ b/include/etl/priority_queue.h
@@ -418,6 +418,7 @@ namespace etl
   public:
 
     typedef typename TContainer::size_type size_type;
+    typedef TContainer                     container_type;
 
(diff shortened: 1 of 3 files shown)


jwellbelove:
It's probably worth adding this to stack & priority_queue too, for max compatibility.

dingari:
Sure! I added the `container_type` typedef to both `etl::stack` and `etl::priority_queue`.

Let me know if I missed anything.

PR #330, opened 2021-01-29
Feature/introduce bip buffer spsc atomic

bku-sue:
This version is working quite reliably on our project now, any timeline for adding this to mainline?

PR #402, opened 2021-07-06
introduce bip buffer spsc atomic

bku-sue:
This is a clean and feature complete commit of the previously submitted version (#330), which in turn can be ignored.

### DIFF
--- a/include/etl/bip_buffer_spsc_atomic.h
+++ b/include/etl/bip_buffer_spsc_atomic.h
@@ -1,0 +1,400 @@
+///\file
+
+/**
+ * @note Based on the works of Andrea Lattuada and James Munns:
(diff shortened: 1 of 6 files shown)


jwellbelove:
I noticed that clearing or destructing the buffer does not call the destructors for the buffered items. Was this bip buffer only intended for trivially destructible types?

Also, the top level template class does not define copy/move constructors or assignment operators. Should these be marked as disabled?

bku-sue:
Yes, it's intended for raw data, but I guess destructing can be added as well if necessary.

I also don't see the use-case for copying the container, so it can be disabled.

bku-sue:
Hi! Can I be of further assistance? Shall I fix the mentioned points and submit a follow-up PR?

jwellbelove:
Yes, that would be helpful as I'm a bit bogged down with trying to get my new C++11/C++17 version of `etl::variant` working at the moment.

PR #408, opened 2021-08-02
to_string_helper: add string conversion helper for denominated integers

bku-sue:
Storing physical values in integer types means that in many cases the representation is not the most meaningful when printing these values. E.g. to keep most of the precision of a voltage measurement, one would store it in either mV or uV format integer, but printing should still be best for user readable "2.9543V". That's what these new functions do, staying with this example:
```
etl::string<10> measurement_V;
int value_uV = 2954300;
etl::private_to_string::add_integral_denominated(2954300, 1000000, measurement_V);
```
The optional format's precision allows fixing a minimal precision to be printed, appending 0s if necessary.

### DIFF
--- a/include/etl/private/to_string_helper.h
+++ b/include/etl/private/to_string_helper.h
@@ -320,6 +320,73 @@ namespace etl
     }
 
     //***************************************************************************
+    /// Helper function for divided integers.
+    //***************************************************************************
+    template <typename T, typename TIString>
+    void add_integral_denominated(T value,
+                                  typename etl::make_unsigned<T>::type denominator,
+                                  TIString& str,
+                                  const etl::basic_format_spec<TIString>& format,
+                                  bool append = false)
+    {
+      typedef typename TIString::iterator   iterator;
+      typedef typename TIString::value_type type;
+
+      if (!append)
+      {
+        str.clear();
+      }
+
+      iterator start = str.end();
+
+      {
+        // split to integral and fractional parts using the denominator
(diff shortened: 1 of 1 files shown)


jwellbelove:
It needs a top level `etl::to_string` function to call the helper function.
You would need a way to uniquely overload the function name.

bku-sue:
What's your suggestion to solve this?

jwellbelove:
I've been working on the code for a couple of days getting all of the corner cases and rounding issues sorted.
One thing that occurred to me is that the denominator should probably be a power-of-ten index rather than a straight number, as denominators like 123456 don't really make any sense.

bku-sue:
It's true that passing simple numbers as denominator allows invalid input to be specified. However I didn't want to add a `pow()` calculation as the additional computing doesn't seem like a good tradeoff. Another thing I have considered since is using `std::ratio` instead, that way some `enable_if` magic could prevent invalid input denominator (and multiplying with the ratio's nominator is small extra effort).

jwellbelove:
To get the formatting correct for all cases I had to calculate the power-of-10 from the denominator anyway (to get the decimal digits required from the input). Using a power-of-10 would actually make the code simpler by removing loops.

jwellbelove:
The denominator parameter could also be read as a "how many digits do I want to shift the decimal point?"

bku-sue:
So basically the parameter will be the denominator's exponent then. It makes sense to pass this if this value is directly needed for the operation.

PR #416, opened 2021-08-22
literal with suffix

CanastraRF:
I've add lot of literal suffix to complie MISRA rules.


jwellbelove:
Does MISRA demand that `size_t` literal initialisation requires `UL` rather than just `U`?
All I can find in MISRA is the line "Constants of unsigned type should have a "U" suffix".

I'm concerned about platforms where `size_t` may be smaller than an `unsigned long`, causing compiler errors or warnings.

CanastraRF:
MISRA does not demand that size_t literal initialisation requires UL.
My experience from 8bit / 32bit / 64bit Compiler is, that size_t is at least unsigned long -> UL

IAR 8bit Compiler: size_t 32bit, unsigned long 32 Bit, unsigned int 16 Bit
ARMCC 5/6 32 bit Compiler: size_t 32bit, unsigned long 32 Bit, unsigned int 32 Bit
GCCARM 32 bit Compiler: size_t 32bit, unsigned long 32 Bit, unsigned int 32 Bit
MSVC  32 bit Compiler: size_t 32bit, unsigned long 32 Bit, unsigned int 32 Bit
MSVC  64 bit Compiler: size_t 64bit, unsigned long 32 Bit, unsigned int 32 Bit

We had problems with integer overflow with IAR 8bit Compiler without the L for size_t
So we recommended that size_t should always become an UL suffix.
Perhaps there are other compiler this different definition of size_t.

jwellbelove:
Does MISRA require the suffix for something like this?
`uint64_t value = uint64_t(0x0000000000000001);`

SR4ven:
Was the 212 KB binary .vs/slnx.sqlite intentionally committed here?

jwellbelove:
No

PR #430, opened 2021-09-26
Add peek to queue_spsc_atomic

fractalembedded:
This adds a `peek` method to the `queue_spsc_atomic` container. Peek is the same as pop except the queue is not modified by this operation. There is no need for a `void` peek as that would essentially be a no operation. If this pattern looks good, let me know if we should more broadly apply it to the other queue classes. I wanted to wait for feedback before putting that effort in.

I also wasn't sure if I should use the `front` naming convention or not. I also am not sure if this should be returning a reference or copying into a reference argument as I have it. I have one question about move semantics which I've posted inline to the code review.

In general, I'm not 100% sure about the direction and code in this PR, and would consider it more of a starting point for a conversation.

### DIFF
--- a/include/etl/queue_spsc_atomic.h
+++ b/include/etl/queue_spsc_atomic.h
@@ -366,6 +366,24 @@ namespace etl
 #endif
 
     //*************************************************************************
+    /// Peek the next value in the queue without removing it.
+    //*************************************************************************
+    bool peek(reference value)
+    {
+      size_type read_index = read.load(etl::memory_order_relaxed);
+
+      if (read_index == write.load(etl::memory_order_acquire))
+      {
+        // Queue is empty
+        return false;
+      }
+
+      value = p_buffer[read_index];
+
+      return true;
+    }
+
+    //*************************************************************************
     /// Pop a value from the queue.
     //*************************************************************************
     bool pop(reference value)
--- a/test/test_queue_spsc_atomic.cpp
+++ b/test/test_queue_spsc_atomic.cpp
@@ -271,6 +271,52 @@ namespace
       CHECK(!queue.pop());
     }
 
+    TEST(test_size_push_peek_pop)
+    {
+      etl::queue_spsc_atomic<int, 4> queue;
+
+      CHECK_EQUAL(0U, queue.size());
+
+      queue.push(1);
+      queue.push(2);
+      queue.push(3);
+      queue.push(4);
+      CHECK_EQUAL(4U, queue.size());
+
+      int i;
+
+      CHECK(queue.peek(i));
+      CHECK_EQUAL(1, i);
+      CHECK_EQUAL(4U, queue.size());
+
+      CHECK(queue.peek(i));
+      CHECK_EQUAL(1, i);
+      CHECK_EQUAL(4U, queue.size());
+
+      CHECK(queue.pop());
+      CHECK_EQUAL(3U, queue.size());
+
+      CHECK(queue.pop());
+      CHECK_EQUAL(2U, queue.size());
+
+      CHECK(queue.pop());
+      CHECK_EQUAL(1U, queue.size());
+
+      CHECK(queue.peek(i));
+      CHECK_EQUAL(4, i);
+      CHECK_EQUAL(1U, queue.size());
+
+      CHECK(queue.peek(i));
+      CHECK_EQUAL(4, i);
+      CHECK_EQUAL(1U, queue.size());
+
+      CHECK(queue.pop());
+      CHECK_EQUAL(0U, queue.size());
+
+      CHECK(!queue.peek(i));
+      CHECK(!queue.peek(i));
+    }
+
     //*************************************************************************
     TEST(test_multiple_emplace)
     {


include/etl/queue_spsc_atomic.h
```diff
@@ -365,6 +365,24 @@ namespace etl
...
+    {
+      size_type read_index = read.load(etl::memory_order_relaxed);
+
+      if (read_index == write.load(etl::memory_order_acquire))
+      {
+        // Queue is empty
+        return false;
+      }
+
+      value = p_buffer[read_index];
```
farixembedded, line 381: I'm not a C++ expert by a long shot. I believe we do _not_ want to use move semantics here as we are not destroying the read object after. Is that correct?

jwellbelove: I think the function should be named 'front' to keep with the convention used by the STL, which the ETL follows.

jwellbelove: It's correct that it should not use move semantics, as it would invalidate the value for the subsequent accesses.

jwellbelove:
Opps! I was setting the base of the pull request and accidentally deleted the branch.
Could you re-do the pull request?

jwellbelove:
OK, I think I've fixed it now.

fractalembedded:
oops, I missed the GitHub notifications for this for a couple days. Sorry about that.

I'll rename from `peek` to `front`. One follow on question to that, `std::queue<T,Container>::front` returns a reference to value in the queue rather than returning by value. Would we want to do a return by value like I have it or consider following the stdlib's method of return by reference (and I suppose we'd want a const overload version as well). What are your thoughts on that?

include/etl/queue_spsc_atomic.h
farixembedded: I'll rename from peek to front. One follow on question to that, [`std::queue<T,Container>::front`](https://en.cppreference.com/w/cpp/container/queue/front) returns a reference to value in the queue rather than returning by value. Would we want to do a return by value like I have it or consider following the stdlib's method of return by reference (and I suppose we'd want a const overload version as well). What are your thoughts on that?

fractalembedded:
> OK, I think I've fixed it now.

Awesome thanks!

include/etl/queue_spsc_atomic.h
jwellbelove: I think it should follow the STL/ETL convention of returning a reference. It can be handy when the objects in the queue are large and a copy just to peek a value would be wasteful.

farixembedded: Sounds good. I'll modify to match that and post a comment when I have it fixed up.

jwellbelove:
Looking at the code I realised that if the `front()` function uses the same API as `pop()`, then it will always result in a copy.

I think the function should return a pointer to the item / nullptr if the queue is empty,
or return a reference, with undefined behaviour if the queue is empty.

```
int* p = queue.front();

if (p)
{
  std::cout << "Peeked value = " << *p;
}
```

or

```
if (!queue.empty())
{
  int& i = queue.front();
  std::cout << "Peeked value = " << i;
}
```

jwellbelove:
Also, the function should be `const` as it does not modify the queue.

jwellbelove:
Another option is to return an `etl::optional<etl::reference_wrapper>`.

fractalembedded:
Awesome, I like all those suggestions. I'd lean to reference return with undefined behavior for an empty queue. I'm a little short on time at the moment (first kid just about due), but I'll follow up with the edits when I get a chance. Thanks for the feedback and patience :).

fractalembedded:
oops, I didn't realize this got merged. It looks like the changes discussed above didn't make it in. Is that intentional? I'm happy to open another PR with those changes. Should I make that PR against the ` ETLCPP:feature/#430-add-peek-to-queue_spsc_atomic` branch?

fractalembedded:
I fixed this up and pushed it. Let me know how it looks and the if it is good the best way to get this into PR again.

A couple follow ups to your questions:
>  EDIT: Actually, the normal queue has both const and non-const front() implementations.

I fixed it to be `const`. I think we do not want a non-const version as this is a SPSC setup, the consumer shouldn't really be modifying the queue at all other that popping/viewing. That being said, I can't think of how it would causes errors if the consumer can modify the content of `front` since it should be the only thing popping (which would make the ref invalid). Might be useful in some cases where you want to avoid making a copy to modify it. Let me know what you think and I can add the non-const version.

> I think the function should return a pointer to the item / nullptr if the queue is empty,
> or return a reference, with undefined behaviour if the queue is empty.

I lean to returning a reference and letting it be undefined behavior when empty. I actually personally like the pointer with the nullptr for empty in general, but `std::queue` uses references and I'm pro sticking to the `std` libs interfaces as much as possible. It wouldn't take much to sway me the other way though.

jwellbelove:
I've modified all of the SPSC & MPMC queues to have `front()` implementations, returning const & non-const references.
I kept the non-const references to keep consistent with the standard STL/ETL queue.

Issue #375, opened 2021-04-27
how to use an ETL container on an "external" memory bank


jerabaul29:
I am using some nice non volatile FRAM chips (my specific modules are https://www.adafruit.com/product/4719 from Adafruit, but I guess the exact model makes no difference). These are nice, big banks of "external" non volatile memory.

I would like to have some ETL containers living fully (i.e., both the data buffer, and the "container magics" around them) on such an external memory bank. These have a very simple API, something like:

```
fram.write8(address, byte_value);
fram.read8(address);
```

This allows to write and read individual bytes of data at individual addresses easily.

Any way / how to make an ETL container live on such an external memory bank with an API to access it? I looked at the documentation but did not find hints of how I could do something like this.

jwellbelove:
`etl::vector` and `etl::deque` can use external memory blocks to store data, but they expect it to operate as normal random access memory. There is no facility to access the memory block via an API. For the containers to be non-volatile, _all_ of the member variables of the container would have to be situated in the FRAM to ensure that container started in the correct state.
You would need to create your own containers to interface with this FRAM.

jerabaul29:
Yes, agree about the all member variables on FRAM point - this is what I meant, you write it much better.

Ok, so if I understand well you mean that this is outside the scope of ETL more or less, right?

jwellbelove:
It would be quite a lot of work to create non-volatile versions of all of the containers in the ETL. The current ones could not easily be adapted without seriously affecting their performance for use with normal RAM.

jerabaul29:
I understand and that is very fair 😊 Closing as out of the scope of ETL, thanks for the discussion!

jerabaul29:
I have been thinking about this some more as this is a feature that would be really convenient to have when working with some data that really needs to survive reboots / interruptions etc. I have some boards that specifically have some non volatile memory banks just for this purpose, feels very sad to not use these.

I was thinking that maybe a weaker form of this would be possible and still nearly as useful. My idea is the following: could we have a couple of functions, defined on any container, that:

- given a container reference, return how many bytes of memory are needed to store the full container (i.e. metadata, content, etc; everything necessary to dump and restore the container). Something like (I think the ```container``` type is not the right one, and / or I suppose this may need to be defined on each container class, but you get the idea :) ):

```
size_t complete_memory_print(etl::container const & container_in)
```

[...]

PR #433, opened 2021-09-30
Feature/pool ext  variant pool ext

daniel-brosche:
as discussed in #375


daniel-brosche:
Everything seems to work but two tests fail which has nothing to do with my changes:
../test/test_delegate.cpp:1021:1: error: Failure in test_call_if_and_not_valid_returning_void: !function_called
../test/test_delegate.cpp:1022:1: error: Failure in test_call_if_and_not_valid_returning_void: !parameter_correct
FAILURE: 1 out of 3951 tests failed (2 failures).
Test time: 4.38 seconds.

daniel-brosche:
@jwellbelove here are the states of function_called and parameter_correct

image

jwellbelove:
`function_called` and `parameter_correct` should have both been set to `false` in `SetupFixture`.
As `was_called` is `false`, that means that no function that would have set them to `true` was called.
I'm not sure how they get the values that you are seeing!

jwellbelove:
image

daniel-brosche:
What is your target system?

jwellbelove:
Although I don't get the unit test fail with CMake using GCC or clang, I do get it if I compile using meson (same version of GCC)

jwellbelove:
OK, I seem to have found the issue.
I moved `SetupFixture` into the anonymous namespace with the unit tests and the test now pass.

daniel-brosche:
strange issue, when I execute just the single test case (without moving SetupFixture) then the test also pass - just mystic

daniel-brosche:
have no double checked and it works also on my side when moving SetupFixture into the namespace

daniel-brosche:
is my source Ok to approval the workflows?

PR #463, opened 2021-11-15
let map/multimap/set/multiset return iterator on erase

mampfes:
Beginning with C++11, erase(iterator) and erase(const_iterator) returns an iterator following the removed element.

This PR improves C++11 compatibility for the STL container replacements in etl by adding the missing return value.

I didn't touch the flat_* and reference_flat_* containers in etl because I'm not very familiar with them. If necessary, I could try to add the return value there as well.

### DIFF
--- a/include/etl/map.h
+++ b/include/etl/map.h
@@ -993,10 +993,10 @@ namespace etl
     //*************************************************************************
     /// Erases the value at the specified position.
     //*************************************************************************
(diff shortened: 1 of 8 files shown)


jwellbelove:
I've pull these into a feature and I shall look at the flat_map/set containers, as they are a little more complicated to modify.

PR #470, opened 2021-11-20
Fix list templated insert interface

bolry:
Shall except a const_iterator


jwellbelove:
I been going through all of the containers after your previous pull requests and I've already done this change.

bolry:
OK, didn't see that on the development branch at the time of writing.

PR #480, opened 2021-12-10
Fix for double formatting with precision > 9

### LINKED ISSUE #479: etl::string_stream double formatting broken
I just upgraded release `20.2.2` to `20.21.0` and have noticed that writing a double to a string stream using `precision` has changed to a broken behavior.

I've narrowed it down to the following section of our code.

```c++
    WW_LOG_TRACE("value=%f", value);
    etl::string<20> data_string;
    etl::string_stream data_stream(data_string);
    etl::format_spec format;
    format.precision(10);
    data_stream << format << value;
    WW_LOG_TRACE("data_string=%s", data_string.c_str())
```

Produces the following logs

```
18: [2021-12-09 16:49:05.308] [firmware] [trace] [utility/system] __write_double:292 value=5.987000
18: [2021-12-09 16:49:05.308] [firmware] [trace] [utility/system] __write_double:304 data_string=5.139173455
```

The problem is that the value `5.987000` should be formatted to a string `5.9870000000`.

LuskeyNoah:
- Resolves #479

### DIFF
--- a/include/etl/private/to_string_helper.h
+++ b/include/etl/private/to_string_helper.h
@@ -56,8 +56,10 @@ namespace etl
   {
 #if ETL_NOT_USING_64BIT_TYPES
     typedef int32_t workspace_t;
+    typedef uint32_t uworkspace_t;
 #else
     typedef int64_t workspace_t;
+    typedef uint64_t uworkspace_t;
 #endif
 
     //***************************************************************************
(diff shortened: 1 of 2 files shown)


jwellbelove:
I was wondering whether 64 bit ints could be eliminated by storing the multiplier as a 'decimal point shift' index, instead of an actual multiplier value.

PR #487, opened 2021-12-27
Sync library name with arduino version

ivankravets:
There is a conflict between the two library manifests: the field "name" differs. As result, PlatformIO Crawler fails and can't fetch the latest version. See 

https://registry.platformio.org/packages/libraries/etlcpp/Embedded%20Template%20Library

### DIFF
--- a/library.json
+++ b/library.json
@@ -1,5 +1,5 @@
 {
-  "name": "ETL Embedded Template Library",
+  "name": "Embedded Template Library ETL",
   "version": "20.22.0",
   "author s": {
     "name": "John Wellbelove",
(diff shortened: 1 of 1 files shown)


ivankravets:
Thanks for the merge!

is it possible to bump a patch release/tag?

ivankravets:
@jwellbelove please ignore my comment above related to bumping a new release. I see we have a conflict in the registry:

- https://registry.platformio.org/packages/libraries/etlcpp/Embedded%20Template%20Library%20ETL - Arduino clone
- Original version https://registry.platformio.org/packages/libraries/etlcpp/Embedded%20Template%20Library


How could we resolve this issue? It is not possible to have 2 packages with the same name in the registry by the same owner.

PR #522, opened 2022-03-29
Update cmake version handling

robamu:
- Introduces a new version file `version.txt`. I could have also specified the version in the `CMakeLists.txt` file directly, but using an external .txt file might be useful to single-source the version number
- Parse that version.txt file in the `CMakeLists.txt` and assign the version number to the `project` call


robamu:
I actually saw that the version is already specified in the `library.json` file and in the `meson.build` file. I'm not fully sure what the best approach is now. Is it possible to specify an external version file for these config/build system files as well to reduce the number of files which need to be updated for each release?

robamu:
I managed to re-use the `version.txt` for both meson and CMake now.. Also includes a small fix for the GCC meson build now

robamu:
`git describe` does not really seem to work like expected in the CI/CD.. I tried fetching the tags, that does not help unfortunately.

jwellbelove:
It would also be nice to be able to automatically update `library.properties` and `library.json`.

robamu:
I think automating is kind of tricky.. what would be possible it to have a script which fetches the latest git tag and updates is at the 3(?) remaining spots.

Issue #521, opened 2022-03-28
CMake: Add package version file for installation


robamu:
Hi,

CMake provides the option to write a package version file. The documentation for this can be found [here](https://cmake.org/cmake/help/latest/module/CMakePackageConfigHelpers.html#command:write_basic_package_version_file).

I think this is really useful if `FetchContent` and `find_package` are used to find an installed ETL library. From what I have seen, I was not able to use the version feature, for example when providing a major version requirement like this:

```cmake
find_package(etl 20 QUIET)
```

Could be that I misused the feature or did something wrong so correct me if I am wrong. If I omit the version number, CMake is able to find the installed ETL library without issues.

Also I think it would be nice if installation instructions are provided for CMake.

I followed the following steps to install a specific version of ETL

```sh
git clone https://github.com/ETLCPP/etl
cd etl
mkdir build && cd build
git checkout <someVersion>
cmake .. && cmake --install .
```

jwellbelove:
I will take a look, but I'm no expert in cmake.

robamu:
I hope this is not already something you have been working on, but I opened a PR which introduces a little bit of version handling to the CMake file:
https://github.com/ETLCPP/etl/pull/522

jwellbelove:
Added 20.27.2

PR #523, opened 2022-03-31
Extended CMake installation handling

robamu:
I extended the installation handling following these instructions: https://dominikberner.ch/cmake-interface-lib/ 

Installation tested on my Ubuntu 21.10 OS and on Windows 11
New version feature was tested as well. Current Compatiblity settings is same major version.

Builds on top of #522 .


jwellbelove:
It looks like the Github actions have all failed.

```
CMake Error at /usr/local/share/cmake-3.22/Modules/WriteBasicConfigVersionFile.cmake:43 (message):
  No VERSION specified for WRITE_BASIC_CONFIG_VERSION_FILE()
```

robamu:
Ah. I actually handled the case where no version is found, but then of cours wiriting a config version file won't work. I'll see that I account for that case as well..

robamu:
Closes #521

PR #527, opened 2022-04-05
CMake install instructions

robamu:
Full set of CMake install instructions. Maybe this would be better situated on the ETL website (as well)?

Use of the major version requirement with `find_package` requires https://github.com/ETLCPP/etl/pull/523


jwellbelove:
Thank you.

PR #533, opened 2022-04-14
CMake: add a flag to enable version lookup outside the root folder

valentink-eroad:
Add ARGN  to `determine_version_with_git()` params list to allow specifying optional ALLOW_LOOKING_ABOVE_CMAKE_SOURCE_DIR flag.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,7 +5,7 @@ cmake_minimum_required(VERSION 3.5.0)
 
 include(${CMAKE_CURRENT_SOURCE_DIR}/cmake/helpers.cmake)
 
-determine_version_with_git()
+determine_version_with_git(${GIT_DIR_LOOKUP_POLICY})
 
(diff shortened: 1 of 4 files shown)


valentink-eroad:
We use ETL as a Git submodule and with the recent CMake changes in 20.27.2 and 20.27.3 it started failing with `Version string HEAD-HASH-NOTFOUND retrieved with git describe is invalid` error. Seems that if .git folder can't be located in ETL root folder it needs additional flag - `ALLOW_LOOKING_ABOVE_CMAKE_SOURCE_DIR`.

I also noticed that the `FetchContent_MakeAvailable(etl)` also fails, but it works with the proposed changes.

Another way to fix it could be just adding the flag permanently:
```cmake
get_git_head_revision(refspec hash "ALLOW_LOOKING_ABOVE_CMAKE_SOURCE_DIR")
```

PR #540, opened 2022-04-26
New alias target etl::etl and CMake instructions update

robamu:
When installing ETL with `find_package`, the actual import target is inside the `etl` namespace. This is the correction for the `CMake` instructions


robamu:
I think I will also add the alias `etl::etl` in the `CMakeLists.txt` so that users which use `add_subdirectory` or `FetchContent` can link against the same namespace target as well

PR #552, opened 2022-06-08
Feature/variant visit

doocman:
Implements etl::visit to emulate std::visit.

Currently tested with MSVC-2022 (17.2.3)

### DIFF
--- a/include/etl/private/variant_legacy.h
+++ b/include/etl/private/variant_legacy.h
@@ -95,6 +95,17 @@ namespace etl
   };
 
   //***************************************************************************
+  /// 'Bad variant access' exception for the variant class.
+  ///\ingroup variant
+  //***************************************************************************
+  class bad_variant_access : public variant_exception {
+  public:
+    bad_variant_access(string_type file_name_, numeric_type line_number_)
+      : variant_exception(ETL_ERROR_TEXT("variant:bad variant access", ETL_VARIANT_FILE_ID"A"), file_name_, line_number_)
+    {}
+  };
+
+  //***************************************************************************
   /// A template class that can store any of the types defined in the template parameter list.
   /// Supports up to 8 types.
   ///\ingroup variant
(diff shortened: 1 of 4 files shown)


jwellbelove:
That was a pretty quick PR, from your suggestion on 6th to the PR on 8th.
Did you write the visit code from scratch, or was a modification of something already existing?
I need to know because of copyright issues and possible conflicting licence requirements.

jwellbelove:
Is there a reason you used `static_cast<TCallable&&>` etc instead of `etl::move`?

doocman:
The code is written from scratch. I have however looket at the MSCV STL
implementation for other things that I've implemented (I have a
"named_tuple" and wanted to see details about get<...> and tuple element).
I've noticed that they (Microsoft) uses static_cast where a "std::forward"
may work just as well. I can change that if you want.
 I'll give push an update soon that also works with gcc and clang on Ubuntu.

// Robin

jwellbelove:
GCC is having an issue with this line at the moment.

`using type = typename visit_result_helper<typename visit_result_single_wrapper<TCallable>::ttype, make_index_sequence<variant_size<remove_reference_t<T1> >::value>, T1, Ts...>::type;`

```
variant_variadic.h:1358:181: error: type/value mismatch at argument 1 in template parameter list for ‘template<template<class ...> class<template-parameter-1-1>, class ...> struct etl::private_variant::visit_result_helper’
 1358 | esult_single_wrapper<TCallable>::ttype, make_index_sequence<variant_size<remove_reference_t<T1> >::value>, T1, Ts...>::type;
      |                                                                                                                     ^
variant_variadic.h:1358:181: note:   expected a class template, got ‘typename etl::private_variant::visit_result_single_wrapper<TCallable>::ttype’
```

doocman:
Yes. I have fix but it is not pushed yet. I'm currently on my job so I'll
push it in a couple of hours.

// Robin

doocman:
I've pushed and opened a pr again but I noticed too late that you had made
changes to the branch 😅. I tried to fix the merge conflicts but was unsure
about e.g. the "ETL_VARIANT_FILE" and formatting.
I also had a problem with "declval" being defined in both utility.h and
type_traits.h (I got a "ambiguous overload" error) so I removed the utility
one from the pr. If that was wrong then I happily fix it if you want it in
a different way. (I tested removing the type_traits version first to be
more conforming to the standard but that broke a bunch of things.
"utility.h" includes "type_traits.h" so I thought it would be a relatively
safe change.)

Best regards,
Robin

jwellbelove:
`ETL_VARIANT_FILE_ID` is defined as the string `"24"`
Each exception has a unique letter added to it to create a short version of the error identifier.
i.e.
```
ETL_VARIANT_FILE_ID"A" => "24A"
ETL_VARIANT_FILE_ID"B" => "24B"
```

Removing `declval` from `type_traits.h` is the best move, as `utility.h` requires definitions from `type_traits.h`.

doocman:
Ahh ok. Thanks for the explanation.

Hmm.. "declval" is used in type_traits.h for "nonvoid_convertable<...>".
Removing 'declval' from type_traits would require a different
implementation for "is_convertible".

// Robin

Den fre 10 juni 2022 kl 19:32 skrev John Wellbelove <
***@***.***>:

jwellbelove:
Ah, I didn't notice that one. It'll probably have to go in a separate private header that they both include.

jwellbelove:
Actually, Ii meant that the other way round. I've removed the `declval` from utility and left it in `type_traits`.

doocman:
That makes sense 🙂
Then I think I am content for now. What do you think?

// Robin

jwellbelove:
All of the tests seem to pass across Window/Linux, MSVC/GCC/Clang

Issue #227, opened 2020-05-24
hash function for enum


firefly1250:
I'd like to use unordered_map in which key is enum.
I can use it by defining hash function by myself, but it is inconvinient.
Could you add definition of hash function for enum like std?

jwellbelove:
I looked at this a while ago, but came up against the need to implement a cross platform ETL version of `std::is_enum`. Implementing `etl::is_enum` requires several other type traits to be defined, all of which can be tricky to implement without help from the compiler. I may be able to put together a solution that will work for compilers that support the `__is_enum(T)` intrinsic.

jwellbelove:
It can be tricky trying to track down which _version_ of each compiler started to define __is_enum(T).

jwellbelove:
https://www.etlcpp.com/bitset.html

PR #561, opened 2022-07-10
is_enum implementation

jesseli2002:
Discussion in #227 notes that a lack of implementation of `is_enum` prevents it from moving forward; not sure if it's because new type traits have been added since but I was able to implement `is_enum` by checking for `static_cast`ability into `int` with SFINAE, then ignoring other types that have this capability (arithmetic types, any class types with `operator int()` overloads, and reference types).

In the process I found that `is_reference` gave the wrong result for rvalue references, which this relies on, so I fixed that too.

Some other bugs which I noticed - I'll raise issues and submit PRs for these when I get the chance:
- `etl::void_t` suffers from a C++11 defect - see https://en.cppreference.com/w/cpp/types/void_t and search for CWG issue 1558
- The `message_parser` file has gotten out of sync with the generator for that file; the CMake should probably be setup to run all the generators as part of the build process...

I ran the unit tests with g++ 8.4 on Ubuntu 18.04, but don't have clang or MSVC so can't readily test those. https://www.etlcpp.com/unit_testing.html says that there should be CI for this?

### DIFF
--- a/include/etl/generators/type_traits_generator.h
+++ b/include/etl/generators/type_traits_generator.h
@@ -93,7 +93,7 @@ namespace etl
   //*****************************************************************************
   // Traits are defined by the ETL
   //*****************************************************************************
-  
+
   //***************************************************************************
   /// integral_constant
   template <typename T, const T VALUE>
@@ -426,17 +426,6 @@ namespace etl
 #endif
 
   //***************************************************************************
-  /// is_reference
-  template<typename T> struct is_reference_helper : false_type {};
-  template<typename T> struct is_reference_helper<T&> : true_type {};
-  template<typename T> struct is_reference : is_reference_helper<typename remove_cv<T>::type> {};
-
-#if ETL_USING_CPP17
-  template <typename T>
-  inline constexpr bool is_reference_v = is_reference<T>::value;
-#endif
-
-  //***************************************************************************
   /// is_lvalue_reference
   template<typename T> struct is_lvalue_reference_helper : false_type {};
   template<typename T> struct is_lvalue_reference_helper<T&> : true_type {};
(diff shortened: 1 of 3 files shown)


jwellbelove:
Thanks for the PR.
I run VS2019 on Windows and GCC and Clang scripts on WSL locally and Github actions run the same on  a push to `master`.

jwellbelove:
I will merge the PR to the feature branch to look to see what changes are necessary.

jesseli2002:
> It looks like the class and test code cannot cope with No STL option or pre-C++ 11 compilation.

Looking at it a bit closer - I didn't notice the part of `etl/type_traits.h` which just inherits from the STL if that's available, and I don't think I put in a `is_enum : std::is_enum`

For reference, my implementation requires C++11; in particular it uses `decltype`

jwellbelove:
It's becoming harder to add new features without using at least C++11 as the minimum standard.

Innokentiy-Alaytsev:
Oh, I see I missed the opportunity for proposing my version of `is_enum` type trait:
```c++
namespace internal {
	enum class is_enum_check_enum {};

	template < class TType >
	static constexpr etl::negation< etl::disjunction<
	    etl::is_arithmetic< TType >,
	    etl::is_reference< TType >,
	    etl::is_rvalue_reference< TType > > >
	    is_enum_check (decltype (static_cast< is_enum_check_enum > (
	        etl::declval< TType > ()))*);

	template < class TType >
	static constexpr etl::false_type is_enum_check (...);
}; // namespace internal


template < class TType >
struct is_enum : public decltype (internal::is_enum_check< TType > (0)) {
};


template < class TType >
static constexpr bool is_enum_v = is_enum< TType >::value;
```

jwellbelove:
Are there any advantages/improvements over the one submitted?

Innokentiy-Alaytsev:
I guess not. Just sharing another implementation that seems to work according to my tests.

jesseli2002:
@Innokentiy-Alaytsev 
For reference, StackOverflow's answer, which at a glance seems to be more complete: https://stackoverflow.com/a/11317409/10762409
I didn't go with it because there were some other type traits that would've needed to been implemented.

Innokentiy-Alaytsev:
@jesseli2002 IMHO, both your and mine implementations are valid and don't actually need all those traits: checks for `is_class`, `is_union`, etc., seem to be superseded by `static_cast`-based check, `is_integral` and `is_floating_point` are superseded by `is_arithmetic`.

jesseli2002:
> @jesseli2002 IMHO, both your and mine implementations are valid and don't actually need all those traits: checks for `is_class`, `is_union`, etc., seem to be superseded by `static_cast`-based check, `is_integral` and `is_floating_point` are superseded by `is_arithmetic`.

Right, didn't see the `static_cast` in my glance over your code

PR #567, opened 2022-07-20
error-handler: only return when the condition is false

benedekkupper:


jwellbelove:
I should probably create some additional CMake projects that test the various permutations of error handler configurations.

benedekkupper:
> I should probably create some additional CMake projects that test the various permutations of error handler configurations.

Thanks for following up on that idea!

PR #585, opened 2022-08-28
Fix etl::result<void> being unusable due to deleted default constructor

aBraM-aBraM:
`etl::result<void>` was unusable due to the lack of a default constructor. This resulted 🙈 in the result being unusable as functions with `etl::result<void>` could only return errors results

### DIFF
--- a/include/etl/result.h
+++ b/include/etl/result.h
@@ -227,9 +227,11 @@ namespace etl
   public:
 
     //*******************************************
-    /// Cannot be default constructed
+    /// Default Constructor
     //*******************************************
-    result() = delete;
+    result()
+    {
+    }
 
     //*******************************************
     /// Copy constructor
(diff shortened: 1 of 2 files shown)


aBraM-aBraM:
I've force pushed into my own branch to rephrase the commit.

hackerjones:
Maybe I don't understand how to use the void specialization. It is always returning an error so why not just return the error without wrapping it in this? I would expect I could return either (void) where both is_value() and is_error() return false. Or I could return an error and is_error() would return true. What am I not understanding here?

jwellbelove:
I think you're correct. For a default constructed `result<void, Error>`, `is_error()` should return `false` and only return `true` if an error has been set.
I've also just noticed that the rvalue reference member functions are not wrapped by a check for C++11.

Could you raise this as an issue, so that I don't forget fix this?
Thanks.

Issue #588, opened 2022-09-02
minmax_push.h/minmax_pop.h includes are missing in string_utilities.h


ilyapas:
The header string_utilities.h is using the etl::min function without including the headers "private/minmax_push.h" and "private/minmax_pop.h", as it is done in all other ETL modules where these functions are used. This leads to compile errors on Windows (VS 2022) if min/max macros are defined by the system.

jwellbelove:
Fixed 20.34.0

PR #589, opened 2022-09-02
Add minmax_push.h/minmax_pop.h includes

ilyapas:
Includes for minmax_push.h/minmax_pop.h are required due to usage of etl:min in this module. This prevents conflicts with system defined min/max macros.


ilyapas:
Fixes issue #588

PR #597, opened 2022-09-12
WIP: Add basic Sphinx documentation support

robamu:
I saw that the ETL documentation is built using Doxygen and hosted here: https://www.etlcpp.com/doxygen/

There is a way to use Sphinx to generate the documentation which is well documented here: https://devblogs.microsoft.com/cppblog/clear-functional-c-documentation-with-sphinx-breathe-doxygen-cmake/ .
Using Sphinx has many advantages which are also explained in the article above.

I added the support for Sphinx docs as specified in the article.
Using sphinx also allow to easily set a custom style. Maybe there is some style file which looks visually closer to https://www.etlcpp.com/ :)
It's still WIP because cross-references to all modules to be documented need to be added manually.
However, it is already possible to build and inspect it, so you can try building this and check whether you like it.

Building the docs requires `breathe` and `sphinx` in addition to `doxygen`:

```sh
python3 -m pip install breathe sphinx
```

The docs can then be built with the following command

```sh
mkdir build
cmake -DBUILD_DOCS=ON ..
cmake --build . -j
```

after that you can simply open them with `docs/sphinx/index.html`

```sh
firefox docs/sphinx/index.html
```


jwellbelove:
To be honest, I started to use Doxygen style markup in the ETL at first, as that's what I used in my previous job, but I haven't generated any Doxygen documentation from it for _years_. Also I haven't really bothered adding the markup for quite some time, so I'm not sure how useful the output would be now.

I'll merge it into a faature branch for the moment until I can look at this at some point in the future.

robamu:
I might add an additional PR with some example cross-references soon.

robamu:
The doxygen output looks good enough for the `etl::map` and `etl::multimap` from what I have seen.

Issue #591, opened 2022-09-06
Automatically define __STDC_LIMIT_MACROS for appropriate compilers


dhebbeker:
According to [the manual for using ETL with different compilers](https://www.etlcpp.com/compilers.html) for some compilers it is required to have `__STDC_LIMIT_MACROS` defined. I wonder if it would be an improvement if that macro would be defined automatically by `platform.h` or its dependencies in case such a compiler is detected and the appropriate limit macros are not already defined?

dhebbeker:
Yes, I am compiling for Extended [Embedded C++](http://www.caravan.net/ec2plus/language.html), which is a derivative from C++98.

dhebbeker:
737b0bf6f715941b088c41554d99ed8ff4a50619 is not sufficient to resolve this issue, as often `stdint.h` is included *before* `platform.h`.

jwellbelove:
Thanks, I'll look at that some more.

jwellbelove:
Fixed 20.34.0

PR #601, opened 2022-09-13
Move __STDC_LIMIT_MACROS and __STDC_CONSTANT_MACROS before stdint.h

dhebbeker:
Macros must be defined before first include of stdint.h. Else they have no effect.

This relates to #591.


dhebbeker:
Actually even this solution is susceptible as it can not be guaranteed, that `stdint.h` will not be included before by the user.

jwellbelove:
That is true. It may be wise to leave the current advice in [](https://www.etlcpp.com/compilers.html)

PR #603, opened 2022-09-14
Feature/add back insert iterator

evantill:
Add `back_insert_iterator` and `back_inserter()`


include/etl/iterator.h
```diff
@@ -604,7 +604,93 @@ namespace etl
...
+    }
+
+    /// Simply returns *this.  (This %iterator does not @a move.)
+    back_insert_iterator operator++(int)
+    {
+      return (*this);
+    }
+
+  protected:
+    TContainer* container_;
```
jesseli2002: https://en.cppreference.com/w/cpp/iterator/back_insert_iterator 
Strictly following the C++ STL would have `container_` be `container`

include/etl/iterator.h
```diff
@@ -604,7 +604,93 @@ namespace etl
...
+     *              presumably a reference-to-const T for container<T>.
+     * @return This %iterator, for chained operations.
+     */
+    back_insert_iterator& operator=(const typename TContainer::value_type& value)
+    {
+      container_->push_back(value);
+      return (*this);
+    }
+
+    back_insert_iterator& operator=(typename TContainer::value_type&& value)
```
jesseli2002: Should be wrapped in a check for C++11 (`ETL_HAS_CPP11` or something, don't quote me on that - there's examples of this basically wherever there's a move constructor, for example)

jesseli2002 (changes requested):

include/etl/iterator.h
evantill: It is already wrapped in a `ETL_USING_CPP11`.
see Line 435: 
https://github.com/ETLCPP/etl/blob/a84cdc6058bb7157fb22e8b06cd77cc218a241e1/include/etl/iterator.h#L435

evantill: but I will fix that to make it clearer (by closing the previous `ETL_USING_CPP11`check and reopening it).

include/etl/iterator.h
evantill: should be fixed with https://github.com/ETLCPP/etl/pull/603/commits/9499c3ef93c6abbf1f0aa27fd75a3bfdf4fa3ea6

include/etl/iterator.h
evantill: ~~should I wrap also the unit test ?~~

evantill: should be fixed by 37bf8b5018cc1cd444125a3b4d744f13b2c2ccb0

evantill: I also wrapped the unit test in d43098d0a22e1296eab118a9acc563aaf1ba32ed

jesseli2002: My bad - although, is there any reason why `back_insert_iterator` couldn't be made available to C++03? [The STL even had it pre-C++11.](https://en.cppreference.com/w/cpp/iterator/back_insert_iterator)

jwellbelove:
The all iterator tags should use the `ETL_OR_STD` namespace.
I think there should be also be a matching `etl::back_inserter` defined.

There is also the complementary `front_insert_iterator` & `etl::front_inserter` in the STL.

include/etl/iterator.h
evantill: @jesseli2002 the same is true for the `reverse_iterator` implementation.
Maybe someone else can answer your question.

jwellbelove: There's no reason that back_insert_iterator  couldn't be made available to C++03, the other iterator types are.
Just wrap any rvalue reference parameter funtions in `ETL_USING_CPP11` compilation guards.

jwellbelove: You should also include, for completeness, `back_inserter`, `front_insert_iterator` & `front_inserter`.

evantill: `front_inserter` added in 5eb8e4f6762e626a69d460f6de18930aa23bb259.

evantill: `back_insert_iterator` have been made available for C++03.
**But I can not test the code (I don't find how to configure the build to test in C++03)**

include/etl/iterator.h
```diff
@@ -604,7 +604,191 @@ namespace etl
...
+   * template parameter deduction, making the compiler match the correct types for you.
+   *
+   * @tparam TContainer The container type.
+   * @param container A container of arbitrary type.
+   * @return An instance of back_insert_iterator working on @p container.
+   */
+  template <class TContainer>
+  ETL_NODISCARD ETL_CONSTEXPR17 inline ETL_OR_STD::back_insert_iterator<TContainer> back_inserter(TContainer& container)
+  {
+    return ETL_OR_STD::back_insert_iterator<TContainer>(container);
```
evantill, line 694: Is this the correct way of using the `ETL_OR_STD`namespace ?

include/etl/iterator.h
```diff
@@ -604,7 +604,191 @@ namespace etl
...
+  template <class TContainer>
+  class back_insert_iterator : public iterator<output_iterator_tag, void, void, void, void>
+  {
+  public:
+    /// A nested typedef for the type of whatever container you used.
+    typedef TContainer container_type;
+
+    /// The only way to create this %iterator is with a container.
+    explicit ETL_CONSTEXPR17 back_insert_iterator(TContainer& c)
+      : container(addressof(c))
```
evantill: is it best to use a qualified name here for `addressof()` ?

include/etl/iterator.h
```diff
@@ -604,7 +604,191 @@ namespace etl
...
+     * This kind of %iterator doesn't really have a @a position in the
+     * container (you can think of the position as being permanently at
+     * the end, if you like).  Assigning a value to the %iterator will
+     * always append the value to the end of the container.
+     *
+     * @param value An instance of whatever type container_type::const_reference is;
+     *              presumably a reference-to-const T for container<T>.
+     * @return This %iterator, for chained operations.
+     */
+    ETL_CONSTEXPR17 back_insert_iterator& operator=(const typename TContainer::value_type& value)
```
evantill, line 644: I have used `ETL_CONSTEXPR17` but I just found there is also `ETL_CONSTEXPR20`, `ETL_CONSTEXPR`, `ETL_IF_CONSTEXPR`.
Are there some guidelines to help select the correct one ?

evantill (commented):
Some questions

include/etl/iterator.h
jwellbelove: `ETL_OR_STD` is used for a small set of functions and classes that may be expected when interacting with the the STL, especially when (Argument Dependent Lookup (ADL) is invoked in the ETL.
e.g.
std::swap()
std::begin()
std::end()

In the above case in think you should just be returning the ETL version.

include/etl/iterator.h
jwellbelove: Yes, use `etl::addressof()`.
I prefer to use qualified names, as it protects against weird errors if a header is missing and a global exists with the same name.

include/etl/iterator.h
jwellbelove: The `ETL_CONSTEXPRxx` macros are defined to `constexpr` or 'nothing' depending on the C++ language level.

For example `ETL_CONSTEXPR14` is only defined as `constexpr` for C++14 and above.

`ETL_IF_CONSTEXPR` is defined as `constexpr` for C++17 and above.
It is used in this context...
`if ETL_IF_CONSTEXPR(etl::is_signed_v<T>)`

jwellbelove: `ETL_CONSTEXPR` >= C++11
`ETL_CONSTEXPR14` >= C++14
`ETL_CONSTEXPR17` >= C++17
`ETL_CONSTEXPR20` >= C++20
`ETL_IF_CONSTEXPR` >= C++17

jwellbelove: A big difference between C++11 and C++14 was to do with what was allowed in a `constexpr` function.
I try to remember to declare new additions to the ETL as `constexpr` wherever I can.
If member functions in a class could be a mixture or `ETL_CONSTEXPR` or `ETL_CONSTEXPR14` I tend to stick to `ETL_CONSTEXPR14` for all of them.

include/etl/iterator.h
evantill: Let me resume the usage of `ETL_OR_STD` :

Definition of `ETL_OR_STD` depends on `ETL_NOT_USING_STL` itself dependings on `ETL_NO_STL`.

According to the [documentation](https://www.etlcpp.com/no_stl.html) :
> When `ETL_NO_STL` is defined then the ETL will use its own versions, otherwise the STL versions will be used.

In this case, I think that the actual code is the right think to do. **Is that correct ?**

include/etl/iterator.h
evantill: implemented in b01a9943c88296e579040dd8975003fb95754b30

jwellbelove:
Was there a particular reason that you used `ETL_CONSTEXPR17` instead of `ETL_CONSTEXPR14`?

evantill:
> Was there a particular reason that you used `ETL_CONSTEXPR17` instead of `ETL_CONSTEXPR14`?

I can not remember the link but I have adapted the source from the std implementation and the `constexpr` xas added since C++17.

PR #604, opened 2022-09-14
Update test sources in meson build

evantill:
- Update list of tests according to CMakeLists.txt content.
- Update compiler args :
	- Add same arguments as in CMakeLists.txt.
	- Remove non-vitual-dtor warnings (code need to be fixed before enabling this warning).
	- Enable sanatizer.
- Add unittest-cpp subproject as a wrap dependency.
- Add `meson test`target to start the `etl_unit_tests` program.


test/meson.build
```diff
@@ -225,33 +249,39 @@ etl_test_sources = files(
...
-if meson.get_compiler('cpp').get_id() == 'gcc'
-    etl_test_sources += files('test_atomic_gcc_sync.cpp')
+if meson.get_compiler('cpp').get_argument_syntax() == 'gcc'
+    compile_args += '-fsanitize=address,undefined'
     compile_args += '-fexceptions'
+    compile_args += '-Wall'
+    compile_args += '-Wextra'    
+    compile_args += '-Wno-non-virtual-dtor' #TODO remove and fix warning in code
+    compile_args += '-Werror'
+    link_args += '-fsanitize=address,undefined'
```
eli-schwartz, line 284: If it's not good enough to set those via `meson setup -DXXX=...` then you can force it per target with: 

```meson
executable(
    ...,
    override_options: [
        'werror=true',
        'warning_level=3',
        'b_sanitize=address,undefined'
    ],
)
```

evantill: I have added those lines to be in sync with the CMake build :

https://github.com/ETLCPP/etl/blob/181a438287b89b300d66a0facd2d85089560aa71/test/CMakeLists.txt#L342-L358

evantill: Do I need to change the code ?

PR #613, opened 2022-09-29
Fix bare assert in reference_counted_object.h

crazyscot:
This upgrades the use of bare assert() into the full ETL error handling mechanism.

Signed-off-by: Ross Younger <crazyscot@gmail.com>


jwellbelove (changes requested):
`ETL_MULTI_LOOP_FILE_ID"A"` is the wrong tag for this exception.
You will need to add `ETL_REFERENCE_COUNTED_OBJECT_FILE_ID` to `file_error_numbers.h`

crazyscot:
Ah, now I see what's going on there! Thanks, updated my branch.

PR #614, opened 2022-10-01
unaligned types: match the signedness of the storage type with the template type

benedekkupper:
The storage type should either be always unsigned, or signed only when the template type is signed. The latter is implemented by this change, not necessarily in the nicest way, so do review first.

### DIFF
--- a/include/etl/unaligned_type.h
+++ b/include/etl/unaligned_type.h
@@ -58,10 +58,10 @@ namespace etl
 
       static ETL_CONSTANT size_t Size = Size_;
 
-      typedef char*                                 pointer;
-      typedef const char*                           const_pointer;
-      typedef char*                                 iterator;
-      typedef const char*                           const_iterator;
+      typedef unsigned char*                        pointer;
+      typedef const unsigned char*                  const_pointer;
+      typedef unsigned char*                        iterator;
+      typedef const unsigned char*                  const_iterator;
       typedef etl::reverse_iterator<iterator>       reverse_iterator;
       typedef etl::reverse_iterator<const_iterator> const_reverse_iterator;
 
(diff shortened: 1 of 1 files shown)


jwellbelove:
You overlooked the comment for `class unaligned_type_common` 
_Contains all functionality that doesn't require the type._

You added a type to `class unaligned_type_common`, which defeats the purpose of the class.

benedekkupper:
Do you agree with my problem though? For me it's also acceptable to simply change the storage fro char to unsigned char, as it doesn't make sense to consider it signed (how would the sign extension be valid in this case?).

jwellbelove:
I'm not sure why signed/unsigned storage is seen as a problem. The storage is just an array of raw data bytes that simply get copied to/from the arithmetic type's memory. The arithmetic value is unaffected by the underlying storage type. There is no sign extension occurring.

i.e.
`uint16_t v = 4660;`
Underlying byte values (little endian)
`0x12 0x34`
It is irrelevant to the arithmetic type whether these bytes are signed or unsigned.

benedekkupper:
The signed storage type is a problem as the storage is publicly accessible via `operator[]` and `data()`. If someone does this for example, they are in for a surprise:
```
etl::be_uint16_t ut = 0x1234;
int firstbyte = ut[0];
assert(firstbyte == 0x12);
```

jwellbelove:
OK, I see what you are talking about now.

The unit tests are currently failing.
Example error
```
/unaligned_type.h:335:26: error: no matching member function for call to 'copy'
      unaligned_copy<T>::copy(this->storage, value);
```

jwellbelove:
I'll merge and fix the unit tests

benedekkupper:
Thanks for picking this up, much appreciated!

Issue #617, opened 2022-10-05
[bug] incorrect computation of find_next in bitset


Chiraffollo:
The method find_next returns erroneous result if the starting position is larger or equal the number of bits in the underlying type, e.g., for bitset<64, uint8> method find_next only works for starting position < 8.
Additionally, the class may not compile if starting position is larger than bits in unsigned.

Example:
```
etl::bitset<64> bs1(0xC000'0000'0000'0031ULL);
size_t bs1fnt1 = bs1.find_next(true, 10);
size_t bs1fnt2 = bs1.find_next(true, 59);
CHECK_EQUAL(62U, bs1fnt1);
CHECK_EQUAL(62U, bs1fnt2);
```
Result:
bs1fnt1: `Expected 62 but was 54`
bs1fnt2: `Expected 62 but was 18446744073709551615`

jwellbelove:
Fixed in next release

PR #618, opened 2022-10-05
fix bug in find_next of (new) bitset class

Chiraffollo:
fixes bug #617


jwellbelove:
Does the legacy bitset have the same issue?

Chiraffollo:
no, the legacy bitset does not have issue, it works fine

Issue #572, opened 2022-07-24
Use standard feature testing


benedekkupper:
ETL is doing a lot of its own detection on which language features are supported. The problem is, it's not very accurate. I got here because by updating ETL in my [C++20 project](https://github.com/benedekkupper/stm32-i2c-hid) the build has failed. The problem boils down to the `#define ETL_NO_SMALL_CHAR_SUPPORT ETL_CPP20_NOT_SUPPORTED` line. This condition is correct as it is spelled out. However, GCC 10 sets `__cplusplus` to `201709L` for `-std=gnu++20`, so the macro gets an incorrect value, compilation fails due to duplicate definition of `char8_t`.

A better way to deal with these issues (at least from C++20) is relying on the [C++ standard feature tests](https://en.cppreference.com/w/cpp/feature_test). These same macros could be defined by ETL itself for earlier C++ versions, as long as the `__cplusplus` numbering pitfall mentioned above is handled well. (A big oversight from the feature test design is not adding a macro to check if feature test macros are available at all or not.)

benedekkupper:
Any timeline for merging this branch? The later it gets done, the more additional work it generates.

jwellbelove:
Between holidaying in Malta and catch a bad cold, I've not looked at the ETL for a few weeks. I have several things to work on for the ETL, so I'm going to have to try to work through them over xmas as I get the free time.

rolandreichweinbmw:
Can this issue be considered as done now?

ETL is currently already heavily using feature test macros.

benedekkupper:
I'm sad to see the work put in the PR wasted, but realistically the effort to rename the macros to their specific feature names is not worth it, which is why this didn't go anywhere. I guess changes will be made on a case-by-case basis depending on what breaks.

It would have been beneficial if this approach was at least adopted for new code additions, especially for recent language and standard library features, as using a cutting-edge compiler version sometimes means the feature set is in between two standard releases, and the current way of using either all or nothing from a standard release is suboptimal.

PR #623, opened 2022-10-15
use feature test macros instead of version checks

benedekkupper:
I have started working on #572 myself, the branch is currently work in progress, so don't consider this a real PR at the moment, more of an indication that there's something being done in this direction.

My workflow is this:
1. `git grep ETL_USING_CPP`
2. Try to deduct what feature is needed from that version of C++.
3. Look up the macro name for the feature on https://en.cppreference.com/w/cpp/feature_test
4. If found, replace `ETL_USING_CPPxx` with the feature test macro, otherwise add a comment to `determine_compiler_language_support.h`


jwellbelove:
I'm not sure how these can be used before C++20.
e.g.
How can you use C++20's `__cpp_rvalue_references` in place of ETL_USING_CPP11?
In C++11 `__cpp_rvalue_references` will be undefined.

jwellbelove:
As your link mentions, these feature test macros are only defined for compilers that support C++20 or above.

Feature testing (**_since C++20_**) <<<<<<
Feature testing 
The standard defines a set of preprocessor macros corresponding to C++ language and library features introduced in C++11 or later. They are intended as a simple and portable way to detect the presence of said features.

Issue #516, opened 2022-03-23
Use after move


mampfes:
We are using Coverity static code analyzer and found an issue in etl::observer:

https://github.com/ETLCPP/etl/blob/5577170fd268f1618b05773a6bf4d4a25e76d28d/include/etl/observer.h#L250

`n` is effectively moved by std::forward and therefore only valid for the first matching observer. Consecutive observers get a default-initialized instance.

jwellbelove:
Fixed 20.27.0

Issue #407, opened 2021-07-29
observable::notify_observers() gives compiler error when the passed type is a reference


jovere:
There is a fairly confusing issue with the `notify_observers()` function that is not immediately obvious.  For example, I create an observer and observable with the following definition:

```C++
struct myStruct {};

typedef etl::observer<myStruct&> myObserver;

class anObservable : public etl::observable<myObserver, 1> {
 public:
  void notifySomebody() { notify_observers(data); }

  myStruct data;
};
```

When I call `notifySomebody()`, `data` is being passed by copy to the `notify_observers()` function, not reference. I came across this when I marked the copy/move constructors of `myStruct` as deleted. The immediate way to fix the issue is to call notify_observers with the following: `notify_observers<myStruct&>(data);`  However, it's not immediately apparent to the end user that this is necessary if they don't have the copy/move constructors deleted.

Changing the `notify_observers()` function so there are two versions, one taking a reference and the other an r-side reference seems to fix the issue in modern C++, but I am unsure how to fix it in C++03 and if there is a more elegant solution.

Thanks,
Jeremy

jwellbelove:
I ideally, I suppose observable would interrogate the observer types in the template parameter list and define a `notify_observers()` for each notification type that they expect. I have no idea whether that is possible.

Onthe other hand...
The `notify_observers()` function would not need two variants as it is a template function and therefore a parameter declared as `&&` would be a 'universal reference'. What the function would then `forward` the parameter to `notification()`.

I don't think there's much that can be done for C++03, apart from the fix you used. I'll have to make a note in the documentation.

PR #625, opened 2022-10-18
Revert "Fixed move parameter in notify_observers"

bretzlaff:
This reverts commit fa7f6dc9d7187fb246af920b3c3dcfe75bdb7bbf.

The functionality removed in the previous commit was using the universal reference and was not a move parameter.  This is important to keep because it allows having values that are const and still using the nofity_observer pattern with them.  See [https://www.justsoftwaresolutions.co.uk/cplusplus/rvalue_references_and_perfect_forwarding.html](https://www.justsoftwaresolutions.co.uk/cplusplus/rvalue_references_and_perfect_forwarding.html) for a better explanation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,6 +13,8 @@ Embedded Template Library (ETL)
 [![Build status](https://ci.appveyor.com/api/projects/status/b7jgecv7unqjw4u0/branch/master?svg=true)](https://ci.appveyor.com/project/jwellbelove/etl/branch/master)
 [![Codacy Badge](https://api.codacy.com/project/badge/Grade/3c14cd918ccf40008d0bcd7b083d5946)](https://www.codacy.com/manual/jwellbelove/etl?utm_source=github.com&amp;utm_medium=referral&amp;utm_content=ETLCPP/etl&amp;utm_campaign=Badge_Grade)
 
+[![](https://img.shields.io/static/v1?label=Sponsor&message=%E2%9D%A4&logo=GitHub&color=%23fe8e86)](https://github.com/sponsors/jwellbelove)
+
 ## Motivation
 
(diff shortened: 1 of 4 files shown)


jwellbelove:
Thanks, I knew about  universal references when the issue was raised, but obviously didn't notice!

jwellbelove:
I've managed to find the original issue #516.
The deletion of the function was necessary because the notification can be called multiple times from the observable's loop.
If the forwarding notification 'moves' the value, then all subsequent notifications could get invalid values.

jwellbelove:
This example gives the output...
```
Data = Hello
Data = ?
Data = ?
Data = ?
```

```
#include <iostream>
#include <utility>
#include <string>

class Data
{
public:

  Data(const std::string& s_)
    : s(s_)
  {
  }

  ~Data()
  {
    s = "?";
  }

  Data(Data&& data)
    : s(std::move(data.s))
  {
    data.s = "?";
  }

  Data& operator =(Data&& data)
  {
    s = std::move(data.s);
    data.s = "?";
  }

  std::string s;

  Data() = delete;
  Data(const Data&) = delete;
  Data& operator =(const Data&) = delete;
};

void Function(Data&& data)
{
  Data d = std::move(data);

  std::cout << "Data = " << d.s << "\n";
}

template <typename T>
void NotifyAll(T&& data)
{
  for (int i = 0; i < 4; ++i)
  {
    Function(std::forward<Data>(data));
  }
}

int main()
{
  Data data = { "Hello" };
  
  NotifyAll(std::move(data));
}
```

bretzlaff:
It appears that there was a previous previous issue #407 that prompted the universal reference.

Issue #448, opened 2021-10-18
Use ETL with Zug


pwhittin:
Zug (https://github.com/arximboldi/zug) is a Functional Programming header-only library that implements transducers.  I tried a quick test by replacing std::vector with etl::vector, and it will not compile.  

Any help would be much appreciated!

Here's the source code:
```cpp
#include <iostream>

#include <etl/vector.h>
#include <zug/into.hpp>

using namespace zug;

int main()
{
    auto v = etl::vector<int, 10>{1, 2, 3};
    auto res = etl::vector<int, 10>{};

    auto& res2 = into(res, identity, v);

    for (const auto& i : res2)
        std::cout << i << "\n";

    return 0;
}
```

Here's the compiler information:
```
$ g++ --version
g++ (Ubuntu 9.3.0-17ubuntu1~20.04) 9.3.0
Copyright (C) 2019 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

Here's the compiler output:
```
====================[ Build | zug_test | Debug ]================================
/home/pauwhi/bin/clion-2020.3.3/bin/cmake/linux/bin/cmake --build /home/pauwhi/Downloads/junk/zug-test/cmake-build-debug --target zug_test -- -j 9
[ 50%] Building CXX object CMakeFiles/zug_test.dir/main.cpp.o
In file included from /home/pauwhi/Downloads/junk/zug-test/zug/zug/reduce.hpp:11,
                 from /home/pauwhi/Downloads/junk/zug-test/zug/zug/transduce.hpp:11,
                 from /home/pauwhi/Downloads/junk/zug-test/zug/zug/into.hpp:13,
                 from /home/pauwhi/Downloads/junk/zug-test/main.cpp:4:
/home/pauwhi/Downloads/junk/zug-test/zug/zug/detail/is_non_empty.hpp: In instantiation of ‘bool zug::detail::is_non_empty(const RangeT&, const RangeTs& ...) [with RangeT = etl::vector<int, 10>; RangeTs = {}]’:
[...]

PR #627, opened 2022-10-20
Make tests build with C++14

jovere:
This is a fairly large change to make ETL tests build with C++14 versions of Clang and GCC.

I embarked on this because, while attempting to use the `message_packet`, I was getting compile errors when a `const&` message was passed to the `message_packet` constructor. It turns out, it was being tested, but only for the C++17 version of the code, not the C++14 enabled code. Below is a summary of what has changed.

Specifically for the build process, I made the following changes:
* Building with CMAKE defaults to using C++17, however, if -DCMAKE_CXX_STANDARD is passed on the command line, that is used instead.
* Added builds for all versions of Clang and GCC in C++14.
* Added the `-j` option in the `make` to use all available cores on the build servers.

Along with fixing the `message_packet`, I also have done the following code changes:
* Where template deduction is not specifically tested, the template parameters are specifically entered. Where deduction is tested, wrappers are put around these tests so they are only enabled for C++17 and newer.
* Because C++14 needs even `static constexpr` variables have a defined space, I used an anonymous `enum` to define those values (I found it in other places and ran with it).
* `inline constexpr` variables aren't supported in C++14, but `static constexpr` is supported. Therefore, many places that had those variable templates defined were copied and wrapped for C++14.
* Some uses of copy constructors needed to be created because the assignment operator is manually created. Auto-creation for some of the special operators/constructors and not others was deprecated in C++14.
* Where the the standard library did not have a function implemented in C++14 and did not appear to be used as a test against the ETL, I switched to using the ETL implementation.
* Added `ETL_OR_STD17` so the appropriate `size()` function can be called without name collisions.
* Some functions needed some fancy removal of the enum type to be used with `copy_n` and `fill_n`. These were changed with `#ifdef` wrappers.


include/etl/iterator.h
```diff
@@ -894,7 +894,6 @@ namespace etl
   }
 #endif
 
-#if ETL_NOT_USING_STL || ETL_CPP17_NOT_SUPPORTED
```
jwellbelove: You cannot remove this conditional statement as std::size and etl::size cannot exist concurrently.
This is due to that fact that any use of ADL with this function will raise an ambiguous function error when using the STL.

e.g.
using std::size;
size_t s = size(container);

jovere: Uggh. `using std::size;` is such a terrible idea. Do you have any alternative suggestions? Having these checks run on both C++ versions is a necessity.

jwellbelove: The `using` is necessary due to the way overloads of STL template functions have to be declared.
As you are not allowed to define them in the `std` namespace, you usually define them in the global namespace instead.

When they are used in a template function, the compiler doesn't know whether the STL version or a custom user overload will be the best fit for the type.
Therefore you must use the `using`.

All of the following STL template functions have this issue...
```
swap()
begin()
end()
size()
```

Therefore, in any template code you must do something like the following...

```
template <typename TObject>
void DoSomthing(const TObject obj)
{
  using std::size; // Allow the automatic use of std::size.

  size_t s = size(obj); // ADL lookup. May use std::size or a custom overload.

  // Etc...
}
```

If you want code that can be configured to use either the ETL or STL then you must do the following.

```
template <typename TObject>
void DoSomthing(const TObject obj)
{
  using ETL_OR_STD::size;

  size_t s = size(obj); // ADL lookup. May use std::size, etl::size (depending on configuration) or a custom overload.

  // Etc...
}
```
_Do you have any alternative suggestions?_
What issue were you trying to solve by removing the conditional compilation?

The ETL defines its version of `size()` if either you are not using the STL or you _are_ using it but you are not using C++17.
C++17 was the first version that defined `std::size()`.

jovere: The tests are using `std::size()` in many places. Because as you stated, C++17 is the first one that had `std::size()`, they won't compile when building the tests for C++14. `etl::size()` needs to be used.

jovere: I did not know of the `ETL_OR_STD`. I can replace all uses of `std/etl::size` with that.

jwellbelove: Yes, `ETL_OR_STD` will either be defined as `std` or `etl`, depending on the use of the STL and compiler standard.
As an example, `ETL_OR_STD::pair` is used throughout the unit tests.

jovere: I just tried `ETL_OR_STD` and it doesn't work with cpp14, because there isn't any `size` declared in `std` in cpp14. My first thought was to create a `size_for_etl_tests.h` with the following contents:

```c++
namespace etl_tests{
#if ETL_CPP17_NOT_SUPPORTED || ETL_NOT_USING_STL
using etl::size;
#else
using std::size;
#endif
};
```
Then use `etl_tests::size()` everywhere `std::size` is needed.

jovere: I do have a question. It would seem that the only time a collision would occur is if `using std::size;` was declared inside the etl namespace. If `etl::size` is already defined, and works exactly like the one defined in `std`, why would you use the one outside the etl library?

jwellbelove: The original issue was raised here in #448 
An external library was defining the following...
```
    using std::begin;
    using std::end;
    return begin(r) != end(r) && is_non_empty(rs...);
```
Argument dependent lookup was seeing both `std::begin` & `std::end` as well as `etl::begin` & `etl::end` due to the using directives.

jwellbelove: Getting the ETL to work seamlessly both with and without the STL has been a headache at times!

jwellbelove: Check the definitions in `etl/private/choose_namespace.h`

The are currently definitions for `ETL_OR_STD` and `ETL_OR_STD11` 
Maybe add a `ETL_OR_STD17`?

jovere: Yeah, I noticed that. I already copied the STD11 one and created a STD17 one. I did a push and all the tests appear to still be passing.

jovere:
Hi John,
What is the status of this PR?
I have removed etl::size in the C++17 configuration. Is there anything else that needs to be done?
Thanks,
Jeremy

jwellbelove:
I haven't been able to take a look at this recently, and I shall be away from home over the next couple of weeks (holiday in Malta) so I won't be able to give it proper attention until I get back.
In the mean time I will pull the code into a feature branch on Github.

jovere:
Hi John, 
What is the status of getting this into the main branch?
Thanks,
Jeremy

jwellbelove:
I haven't had a chance to look at it yet as I've been away in Malta for a while and since I've been back I've been ill with a chest infection. I'll try to look at it in the next couple of days, but I'm going to be away again from Sunday to Wednesday.

jovere:
I see this has still not been pulled into the main branch. The longer this sits here, the more it's going to deviate from the changes that are being added. How can I help to get this pulled in?

Thanks,
Jeremy

jwellbelove:
I've started to look at this. I merged in the newest master branch, but there are a few odd compilation errors that I need to figure out.
I'm away on holiday in Spain on Friday for a few weeks, but I may get a bit of time to look at this while I'm away.

PR #628, opened 2022-10-23
fix set of ETL_NO_STL flag

skorokhod:


jwellbelove (commented):
`ETL_NO_STL` is a defined/undefined macro. If it is set to either `0` or `1` it will be  defined and therefore select the No STL option.

skorokhod:
> `ETL_NO_STL` is a defined/undefined macro. If it is set to either `0` or `1` it will be defined and therefore select the No STL option in both cases.

I just made it so by fixing meson.build.
Do you plan to fix it in some other way?

jwellbelove:
Sorry, yes, you did fix it correctly. I'm not quite sure why I saw that as incorrect. I must have been half asleep at the time!

Issue #636, opened 2022-11-22
Optional with pod type does not compile when using emplace


Chiraffollo:
In the optional `emplace` method for pod types the `storage` is accessed by `get_reference` method. But this method is only supported by non-pod types optional.

jwellbelove:
Fixed 20.35.3

PR #637, opened 2022-11-22
Fix bug #636 in optional emplace for pod types

Chiraffollo:
Fixing bug #636

### DIFF
--- a/include/etl/optional.h
+++ b/include/etl/optional.h
@@ -767,7 +767,7 @@ namespace etl
       if (valid)
       {
         // Destroy the old one.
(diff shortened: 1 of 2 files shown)


include/etl/optional.h
```diff
@@ -767,7 +767,7 @@ namespace etl
       if (valid)
       {
         // Destroy the old one.
-        storage.template get_reference<T>().~T();
+        storage.~T();
```
jwellbelove, line 770: As it is a POD you can delete the whole conditional statement and destructor.

PR #645, opened 2022-12-20
Avoid 'missing return statement at end of non-void function' in `etl::visit<>()`.

dhebbeker:
For some definitions of `ETL_ASSERT()` there may be no return statement in case of an invalid type.
This results in undefined behavior.

> ```
> Warning[Pe940]: missing return statement at end of non-void function "etl::visit<TReturn,TVisitor,TVariant>(TVisitor &, TVariant const &) include\etl\private\variant_legacy.h 976
> ```


jwellbelove (commented):
You could also use ETL_ASSERT_FAIL_AND_RETURN_VALUE

dhebbeker:
Thank you, I was not aware of that macro. I changed it.

Issue #646, opened 2022-12-20
Improve default is_constructible, is_copy_constructible, is_move_constructible


aBraM-aBraM:
The default implementation (the ones that exist if the user only defined ETL_NO_STL) can be improved to check for more than pointers and integral values. They don't require intrinsic compiler functions and have implementations in the cpp standard

jwellbelove:
Do you want to raise a PR?

aBraM-aBraM:
> Do you want to raise a PR?

yeah I've already solved this just a minute :)

PR #648, opened 2022-12-20
Feature/#646 improve type traits

aBraM-aBraM:
add #646 requirements

I'll elaborate on the importance of this addition. this causes `variant` of classes/structs to have a move ctor that doesn't do anything and thus causes `result` to have a move ctor that does nothing but can change type from result to error. 

These are two undefined behaviors I've found more might be the result of this

### DIFF
--- a/include/etl/generators/type_traits_generator.h
+++ b/include/etl/generators/type_traits_generator.h
@@ -714,7 +714,7 @@ namespace etl
   //***************************************************************************
   /// is_enum
   ///\ingroup type_traits
-  /// Implemented by checking if type is convertable to an integer thru static_cast
+  /// Implemented by checking if type is convertible to an integer through static_cast
 
   namespace private_type_traits 
   {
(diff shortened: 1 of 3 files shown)


jwellbelove (changes requested):
Some of these traits are only available for C++11 or above.
You will need to surround the code with #if ETL_USING_CPP11 conditionals and make the C++03 compatible code available for that standard.

jwellbelove (changes requested):
The type traits are generated from `type_traits_generator.h`.
You must ensure that type_traits.h is correctly generated from it when the script `generate_type_traits.bat` is run.
The best way I found to do this is to manually change the _generated_ header to what you want, commit it, and then modify the _generator_ header until it generates a header with no differences to that committed.

aBraM-aBraM:
> By only available for C++11 do you mean they don't compile in C++03 or just not defined by the standard?

If they are just not defined I don't think there's a good reason not to define them (without them some containers have undefined / unexpected behavior).

If those traits are not compiling should I just keep it for C++03 to only work with pointers and integral values? Also I'm not sure if the documentation is part of the repository and if so, please tell me so I'd be able to update them.

jwellbelove:
What I mean is, that for any improved type trait that is implemented using features only available in C++11 and above, then the older C++03 implementation (if one exists) should be available if C++11 is not supported.

```
#if ETL_USING_CPP11
  // C++11 implementation
#else
  // C++03 implementation
#endif

#if ETL_USING_CPP11
  // C++11 only implementation
#endif
```

aBraM-aBraM:
> What I mean is, that for any improved type trait that is implemented using features only available in C++11 and above, then the older C++03 implementation (if one exists) should be available if C++11 is not supported.
> 
> ```
> #if ETL_USING_CPP11
>   // C++11 implementation
> #else
>   // C++03 implementation
> #endif
> 
> #if ETL_USING_CPP11
>   // C++11 only implementation
> #endif
> ```

done

Issue #653, opened 2022-12-30
Warning when used with CMake


flaviut:
When I use this as a dependency in my CMake project (with CPM/FetchContent), I get the following warning: "Version string HEAD-HASH-NOTFOUND retrieved with git describe is invalid".

This is because etl [tries to get the version number with Git first, before falling back to `version.txt`](https://github.com/ETLCPP/etl/blob/master/CMakeLists.txt#L9-L12).

Would it be reasonable to wrap [this in a `PROJECT_IS_TOP_LEVEL` flag](https://cmake.org/cmake/help/latest/variable/PROJECT_IS_TOP_LEVEL.html)? If ETL is being built standalone or on older versions of CMake, first try and get the version from git. If ETL is being build as a subdirectory, first try and get the version `version.txt`. In both cases, fall back to the other on failure.

jwellbelove:
That sounds like a reasonable change. Are you able to do the change and create a pull request for this?

jwellbelove:
Fixed 20.35.9

PR #655, opened 2022-12-31
Don't warn on tag missing when subproject (#653)

flaviut:
Different solution than proposed in the issue, since that proposed solution would given unexpected results when an intermediate (untagged) commit is checked out.

This change simply skips warning about a missing git version when this is a subproject, and uses the original version calculation logic.

I've also renamed `determine_version` to `determine_version_with_file`. I'd originally done this in an intermediate version of this PR, but I think that keeping the renaming is clearer code.


flaviut:
I ended up taking a bit of a different approach, but I'd be happy to switch to the original proposed approach if you think it was better.

PR #659, opened 2023-01-06
emplace member functions return reference to emplaced value

### LINKED ISSUE #656: vector emplace_back not returning a reference
Since C++17, `std::vector::emplace_back` returns a reference to the constructed element.

I find this rather handy, what do you think about adding this to etl as well?

W4RH4WK:
Resolves #656


W4RH4WK:
Note that I did not touch `pvoidvector` and `ivectorpointer` as these are in the `private` subdirectory.

I also did not modify the queue wrappers as this would impose additional constraints on the underlying container.

PR #670, opened 2023-02-09
Feature/span equality operators

KurtisT:
Added equals and not equals operator for spans.

### DIFF
--- a/include/etl/span.h
+++ b/include/etl/span.h
@@ -346,6 +346,15 @@ namespace etl
     }
 
     //*************************************************************************
+    /// Compare two spans for equality.
+    //*************************************************************************
+    template <typename Type2, size_t N2, etl::enable_if_t<etl::is_same<etl::remove_cv_t<Type2>, value_type>::value, bool> = true>
+    ETL_NODISCARD ETL_CONSTEXPR bool operator==(const etl::span<Type2, N2>& other) const ETL_NOEXCEPT
+    {
+      return ((pbegin == other.pbegin) && (Extent == other.size()));
+    }
+
+    //*************************************************************************
     /// Obtains a span that is a view over the first COUNT elements of this span.
     //*************************************************************************
     template <size_t COUNT>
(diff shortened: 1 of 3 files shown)


jwellbelove:
I gave the accepted changes some thought last night and I see a problem.
What does it mean to compare a span?
Should we compare the pointers to the range, or the values contained within that range (shallow or deep comparison)?
Both comparisons could be argued as being 'correct'.

This may be the reason why comparisons were not included for `std::span`.

I think that a deep comparison is more valid in that it reflects what a comparison of a container would do.

A better implementation may be to first compare the pointers/size, and, if they do not match, compare the contents of the spans.

jwellbelove:
I've updated the pulled code to do a deep comparison.

KurtisT:
I had the same question about doing a shallow vs. deep comparison and came to the opposite conclusion. My reasoning was just that if the underlying containers are different, the spans would be operating on different containers and would not be considered the same. Even if the data in the containers happen to be the same. I agree the both comparisons can be thought of as correct so if you think a deep compare is more valid that's good with me.

jwellbelove:
Maybe additional explicit comparison member functions could be useful?

Compare internal pointers
`bool is_equal_span_range(const etl::span<U>& other) const`

Compare pointed to data
`bool is_equal_span_data(const etl::span<U>& other) const`

jwellbelove:
Deep comparison was originally in `std::span`, but was removed.
`https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p1085r2.md`

Maybe some sort of overloaded `etl::equal` could be more relevant?

```
template <typename T1, size_t N1, typename T2, size_t N2>
etl::enable_if_t<etl::is_same_v<etl::remove_cv_t<T1>, etl::remove_cv_t<T2>>, bool>
etl::equal(const etl::span<T1, N1>& lhs, const etl::span<T2, N2>& rhs)
{
  return (lhs.empty() && rhs.empty()) ||
         ((lhs.begin() == rhs.begin()) && (lhs.size() == rhs.size())) ||
         etl::equal(lhs.begin(), lhs.end(), rhs.begin(), rhs.end());
}
```

jwellbelove:
There is the question of whether 'empty' spans should compare as equal.

```
int i;
etl::span<int> span1;
etl::span<int> span2(&i, &i);

bool equal = span1 == span2; // ?
```

PR #688, opened 2023-04-26
Update delegate_cpp11.h

egg-art:
Add get method for invocation element stub member.

In current project we have a need to track a content of a container of delegates on a particular event.
I propose this change as it will allow a safe way to access a function pointer behind delegate and it will allow to store and identify what delegate was in the container later during device memory examination.

### DIFF
--- a/include/etl/private/delegate_cpp11.h
+++ b/include/etl/private/delegate_cpp11.h
@@ -94,6 +94,10 @@ namespace etl
   template <typename TReturn, typename... TParams>
   class delegate<TReturn(TParams...)> final
   {
+  private:
+
+    using stub_type = TReturn(*)(void* object, TParams...);
+
   public:
 
     //*************************************************************************
@@ -425,9 +429,12 @@ namespace etl
       return is_valid();
     }
 
-  private:
+    stub_type get_invocation_element_stub() const
+    {
+      return invocation.stub;
+    }
 
-    using stub_type = TReturn(*)(void* object, TParams...);
+  private:
 
     //*************************************************************************
     /// The internal invocation object.


jwellbelove:
You are happy that two delegates that call the same static/member function will return the same value?

jwellbelove:
Would the current functionality not do what you want?
The equality operator compares invocations, which compares the stubs and objects.

```
/// Checks equality.
ETL_CONSTEXPR14 bool operator == (const delegate& rhs) const
{
  return invocation == rhs.invocation;
}

//***********************************************************************
ETL_CONSTEXPR14 bool operator ==(const invocation_element& rhs) const
{
  return (rhs.stub == stub) && (rhs.object == object);
}
```

jwellbelove:
Example unit test.

```
Test test1;
Test test2;

auto d1 = etl::delegate<void(int, int)>::create<free_int>();
auto d2 = etl::delegate<void(int, int)>::create<Test, &Test::member_int>(test1);
auto d3 = etl::delegate<void(int, int)>::create<Test, &Test::member_int>(test2);

etl::delegate<void(int, int)> d4;

using Delegate_List = std::vector<etl::delegate<void(int, int)>>;

Delegate_List delegate_list = { d1, d2, d3 };

Delegate_List::const_iterator itr;

d4 = d1;

itr = std::find(delegate_list.begin(), delegate_list.end(), d4);
CHECK(*itr == d1);
CHECK(*itr != d2);
CHECK(*itr != d3);

d4 = d2;

itr = std::find(delegate_list.begin(), delegate_list.end(), d4);
CHECK(*itr != d1);
CHECK(*itr == d2);
CHECK(*itr != d3);

d4 = d3;

itr = std::find(delegate_list.begin(), delegate_list.end(), d4);
CHECK(*itr != d1);
CHECK(*itr != d2);
CHECK(*itr == d3);
```

PR #689, opened 2023-05-01
feat(span): Add 'as_bytes' and 'as_writable_bytes'

devjoa:
These functions emulate the std C++20 functions std::as_bytes and std::as_writable_bytes.

I made an attempt to recreate 'as_bytes' and 'as_writable_bytes'. I have not any wider experience with metaprogrmaming, but I think this is correct.

### DIFF
--- a/include/etl/span.h
+++ b/include/etl/span.h
@@ -41,6 +41,7 @@ SOFTWARE.
 #include "integral_limits.h"
 #include "memory.h"
 #include "array.h"
+#include "byte.h"
 
 #include "private/dynamic_extent.h"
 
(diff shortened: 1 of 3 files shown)


include/etl/span.h
```diff
@@ -938,6 +939,28 @@ namespace etl
     }
   };
 #endif
+
+  //*************************************************************************
+  /// Obtains a view to the object representation of the elements of the span s.
+  //*************************************************************************
+
+#if ETL_USING_CPP11
```
jwellbelove: Are you sure this is only implementable in C++11?
You could make it compatible with C++03 by replacing `constexpr` with `ETL_CONSTANT`, and `static_assert` with `ETL_STATIC_ASSERT`.

devjoa: No, I'm not sure...
I'm an old C programmer that has just started to program C++ for about 1 year, and not really familiar with its history. I made it C++11 due to I thought 'auto' as a return parameter was not available prior C++11.

I will create a new version without the condition, and the other fixes.

jwellbelove: Yes, `auto` is not available to C++03, you'll have to spell out the return type in full.
A great many embedded programmers are still updating legacy devices/development platforms that do not use any standard later than C++03. I try to allow C++03 compatibility wherever it is technically possible.

jwellbelove: I was a C programmer for about 12 years until somewhen around 2000, when I moved jobs and had the opportunity to look at C++. I was amazed to realise that I'd been implementing a lot of C++ functionality, in C, for years!

devjoa:
Ooops, I closed this PR by mistake. I was too fast on my keyboard and made a ```  git push -d ...`   instead of an `  git push -f`  ...
Can you please open this PR again?

devjoa:
I found how to fix it after a google. I had to re-push my old commit.

PR #690, opened 2023-05-05
Add rvalue accessors to expected and optional, and update constructors.

msmorgan:
- Add rvalue methods to optional and expected.
- Fix constructors in expected.h.

The constructor for `unexpected` was not working properly with the implicit constructor `expected(unexpected<E>)` (probably due to CPP23 overload resolution rules), so it has been replaced with copy- and move-versions of constructors that take `TError`.


jwellbelove:
The code does not pass the the C++ standard sanity checks.
Please check by running the bash script `./runsanitychecks.sh`
You will need both GCC and CLang compilers.
I am currently testing against GCC 12.1.0 and Clang 14.0.0

msmorgan:
Finally got them both installed on WSL. Looking into it.

msmorgan:
@jwellbelove I believe I've fixed the issues you mentioned. Could you run the analyzers?

Issue #722, opened 2023-06-30
Bitset accesses memory out of bounds


Chiraffollo:
The methods set, reset, and test do not check whether the data accessed is in the valid range of used bits. While accessing the most significant byte is just messy, accessing a byte outside the array can cause a fatal error.

jwellbelove:
Thanks, I'll take a look at that.

jwellbelove:
Fixed 20.37.0

PR #724, opened 2023-06-30
Fix issue #722 Bitset accesses memory out of bounds

Chiraffollo:


jwellbelove:
I'm not sure that an access outside of the bounds of the active bitset should fail 'silently'.
I'm thinking that it should emit an `etl::bitset_overflow`.

Chiraffollo:
I agree completely.
But currently `flip` has exactly the same 'silent' behavior and the comment of `test` method states `Positions greater than the number of configured bits will return <b>false</b>`.
In my opinion, all methods should have the same mechanism. All should ignore it silently or `etl::bitset_overflow` should be emitted consistently. Therefore, I oriented myself by the other 'silent' methods or comments.
Should we change all to `etl::bitset_overflow` ?

jwellbelove:
In retrospect I think that the functions should raise an `ETL_ASSERT...`.
The only exception would be the `[]` operator. I tend to stick with the STL convention.
e.g. `std::vector`'s `[]` vs `at()`

Chiraffollo:
Okay, changed it to exceptions (without `[]`). Additionally, I used `ETL_ASSERT_OR_RETURN_VALUE` to allow usage with exceptions disabled .

PR #740, opened 2023-08-10
add static assert with auto message

dhebbeker:
Suggested changes:

- Add overload to function-like macro `ETL_STATIC_ASSERT()` for only one argument (condition). That overload does either call `static_assert()` if C++17 is used or calls `ETL_STATIC_ASSERT()` where the second (message) argument is generated by stringifying the condition.
- In order to perform overload resolution with function-like macros a helper macro has been added to resolve overloads of 1 or 2 arguments.

[...]

PR #741, opened 2023-08-10
add stringify macro

dhebbeker:
These simple C-preprocessor macros may be handy.

In particular I plan to suggest a feature to ETL which would be using these (see [here](https://github.com/ETLCPP/etl/pull/740/files#diff-b7795b84f7b88e318d9a560760118019ff00dcc86ce86ca36774a913bd94f8e4R55)).


jwellbelove:
As this has been taken from the Linux kernel code, doesn't this have a GPL licence?
A derivative work from GPL code cannot be merged into the ETL.

jwellbelove:
Actually, I've just remembered that there is an `ETL_STRINGIFY` in `macros.h`, through not currently a var-args version.

dhebbeker:
As the code is quite trivial and very common in different ecosystems under different licenses I assumed that Copyright protection does not apply. Also the suggested change is barely similar to the original file.

I did overlook that `ETL_STRINGIFY` is available, sorry! It is a much better option to use that instead.

jwellbelove:
```
#define ETL_STRINGIFY_1(...) #__VA_ARGS__
#define ETL_STRINGIFY(...) ETL_STRINGIFY_1(__VA_ARGS__)
```

Is still an improvement on the current 

```
#define ETL_STRINGIFY2(X)  #X
#define ETL_STRINGIFY(X)   ETL_STRINGIFY2(X)
```

dhebbeker:
Officially `__VA_ARGS__` is only available since C++11. But most compilers seem to implement it even before.

Issue #749, opened 2023-08-28
Code change causes warning with armcc compiler


CurtisAMU:
https://github.com/ETLCPP/etl/blob/b38afd7fc98f50a00b207b184506049f05ab7120/include/etl/private/vector_base.h#L171

The extra semicolon causes a warning if `ETL_DEBUG_COUNT` is not defined:
```
..\..\..\Lib\ETL\include\etl/private/vector_base.h(171): warning:  #381-D: extra ";" ignored
      ETL_DECLARE_DEBUG_COUNT;   ///< Internal debugging.
```

We have a zero warning policy for our project, which treats this as an error. It would be okay for debugging to have a warning to ignore but preferably not on release builds.

jwellbelove:
Does it warn with `void;`

jwellbelove:
Forget that last suggestion

jwellbelove:
What about `void(0);`

CurtisAMU:
`ETL_DECLARE_DEBUG_COUNT;` is a declaration and `(void)0;` does not work there; it's an expression.
The former
`ETL_DECLARE_DEBUG_COUNT   ///< Internal debugging.`
without the semicolon worked. What was the problem with that? It looks odd without the semicolon but should work.

This also works with armcc:
`#define ETL_DECLARE_DEBUG_COUNT struct SDebugCountDummy {}`

CurtisAMU:
A dummy typedef would also do.

jwellbelove:
There will also be this warning if `ETL_NO_CHECKS` is defined for `ETL_ASSERT` etc.

jwellbelove:
Do the other macros cause a warning?
`ETL_INCREMENT_DEBUG_COUNT` etc

CurtisAMU:
Actually no. Empty statements are okay, it seems.

CurtisAMU:
> Do the other macros cause a warning? `ETL_INCREMENT_DEBUG_COUNT` etc

Not as far as I can see.

CurtisAMU:
They are not used in a class declaration.

jwellbelove:
If I try an empty statement in a Keil project using ARM compiler V6 I get

`warning: empty expression statement has no effect; remove unnecessary ';' to silence this warning [-Wextra-semi-stmt]`

CurtisAMU:
One of those complementary differences between compiler V5 and V6. There `(void)0;` would help in both cases, I'd expect.

[...]

PR #764, opened 2023-09-21
Reduce compiler warning when using `ETL_DECLARE_DEBUG_COUNT`.

dhebbeker:
Replaced:

    (ETL_DECLARE_DEBUG_COUNT\s*?);+

with

    \1

In order to avoid warning as such

    Warning[Pe381]: extra ";" ignored etl\include\etl\private\vector_base.h 171

When

    !defined(ETL_DEBUG_COUNT)

⚠ Partially reverts c92ab94220b29b4f196c7de44a391fd3a0a1916c.

For `ETL_DECLARE_DEBUG_COUNT` it can be assumed, that it will not be part of an expression.
For the other `ETL_\w*_DEBUG_COUNT` this can not be assumend.

### DIFF
--- a/include/etl/circular_buffer.h
+++ b/include/etl/circular_buffer.h
@@ -170,7 +170,7 @@ namespace etl
     size_type buffer_size;
     size_type in;            ///< Index to the next write.
     size_type out;           ///< Index to the next read.
-    ETL_DECLARE_DEBUG_COUNT;  ///< Internal debugging.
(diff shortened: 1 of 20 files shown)


jwellbelove:
I fixed this in #749, but it looks like I forgot to merge the changes in the last release!

dhebbeker:
I can confirm, this issue is resolved. A different solution has been applied.

PR #773, opened 2023-10-12
Return raw pointer for `etl::multi_span::operator->()`.

### LINKED ISSUE #772: `etl::multi_span::operator ->()` defect
The signature of `etl::multi_span::operator->()` (`const` and non-`const`) has as return type `pointer` which is `T*`. This is correct.

But the member functions return `*p_value` which results in type `T`. This can not work.

dhebbeker:
Resolves #772.

- using `operator*()` in case getting the current element is changed one day
- using `&` to get the address

Simply returning `p_value` may break in case the internal type of `p_value` would change. This way it is more robust to changes.


jwellbelove:
Fixed 20.38.4

PR #777, opened 2023-10-25
Allow Building Tests as Submodule and GCC < 11

jmartens-42:
moved BUILD_TESTS check outside of root cmake check to allow building tests when ETL is a submodule

added check for GNUC >= 11 for -Wstringop-overread ignore pragma since it isn't introduced until gcc/g++-11

PR #784, opened 2023-11-09
Fix library error for Github Actions for Clang C++20

dhebbeker:
I observed that the GitHub Actions fail. Through analysis since which commit the build fails, I to applied a fix to make the workflow running successfully again.

I am not sure why this fixes the workflow. 🤷 I merely re-applied f208eb4073c98b3c190b17a73db3b61aa35fae25. I think that change accidentally got lost in the merge 95c7f44cc61d51406a70ccb9048724f9fbdd78f9.

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -361,6 +361,7 @@ endif ()
 
 if (CMAKE_CXX_COMPILER_ID MATCHES "Clang")
     message(STATUS "Using Clang compiler")
+	set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -stdlib=libc++")
 endif ()
 
 if ((CMAKE_CXX_COMPILER_ID MATCHES "GNU") OR (CMAKE_CXX_COMPILER_ID MATCHES "Clang"))


jwellbelove:
Unfortunately the changes cause clang compilation to fail on my local CI scripts.

```
-- Looking for C++ include pthread.h
-- Looking for C++ include pthread.h - not found
CMake Error at /usr/share/cmake-3.22/Modules/FindPackageHandleStandardArgs.cmake:230 (message):
  Could NOT find Threads (missing: Threads_FOUND)
Call Stack (most recent call first):
  /usr/share/cmake-3.22/Modules/FindPackageHandleStandardArgs.cmake:594 (_FPHSA_FAILURE_MESSAGE)
  /usr/share/cmake-3.22/Modules/FindThreads.cmake:238 (FIND_PACKAGE_HANDLE_STANDARD_ARGS)
  UnitTest++/CMakeLists.txt:45 (find_package)
```

dhebbeker:
Sorry, I wouldn't know how to resolve that. From a quick research it seems that this error message may be a symptom for anything else (i.e. https://github.com/danvratil/qcoro/issues/22#issuecomment-974924300).

jwellbelove:
If I comment out the line `set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -stdlib=libc++")` then everything is fine.

I'm running 
Ubuntu 22.04
Clang 15.0.7

dhebbeker:
Somehow, without that line it does not work on the GitHub runner. I guess this is related to to workaround you applied in https://github.com/ETLCPP/etl/commit/95c7f44cc61d51406a70ccb9048724f9fbdd78f9#diff-24e820d109223f3d266ca9b6edc626ef33158ab4140adec4c207b7e15bbbbc33R20-R25 and https://github.com/ETLCPP/etl/pull/777#issuecomment-1781714104 respectively https://github.com/actions/runner-images/issues/8659.

Would it help if you would update your local clang to version 17? At least that seams to be the workaround applied for the runner...

jwellbelove:
I was hoping that Github would sort the issue out very soon, so that workarounds wouldn't have to be applied.

Unfortunately clang-17 doesn't seem to be available for Ubuntu 22.04 (`E: Unable to locate package clang-17`). (I'm not a Linux expert)

jwellbelove:
After a web search I tried adding `find_package(Threads)`, but that just causes **all** STL header includes to fail.

dhebbeker:
Would it be possible to apply the same manual installation commands as in the workaround: https://github.com/ETLCPP/etl/commit/95c7f44cc61d51406a70ccb9048724f9fbdd78f9#diff-24e820d109223f3d266ca9b6edc626ef33158ab4140adec4c207b7e15bbbbc33R20-R25
That is essentially to download a script which can be used to install a selectable clang version.

The runner also uses Ubuntu 22.04. Thus those steps should be applicable to your local environment as well.

I am not a Linux expert either. I am just hoping the steps applied in the GitHub Action should work for your local setting as well, especially as it is based on the same operating system. 🤞

jwellbelove:
It looks like the Github Actions is using Clang 14, whilst I am using Clang 15.

dhebbeker:
But the script .github/workflows/clang.yml installs clang v17 isn't it?

       wget https://apt.llvm.org/llvm.sh
       chmod +x ./llvm.sh
       sudo ./llvm.sh 17

jwellbelove:
_"But the script .github/workflows/clang.yml installs clang v17 isn't it?"_

The fix didn't work. The compiler was not updated. clang identified as v14,

dhebbeker:
🤔 I have been checking the build logs.

221cc42f22f4fa2f811fcbdd3a0dbf05d70345d8 has been used by the GitHub Actions. That commit does [contain the temporary fix](https://github.com/ETLCPP/etl/blob/221cc42f22f4fa2f811fcbdd3a0dbf05d70345d8/.github/workflows/clang.yml#L86-L91). [The run for Clang C++20 Linux - STL (ubuntu-22.04)](https://github.com/ETLCPP/etl/actions/runs/6658851825/job/18096628442#logs) did fail. After a presumably successful installation of clang 17, the compiler called by the build process identifies itself as clang 14.

I think a step in the installation process is missing. Or the workflow should instruct to use that specific clang version. This has been mentioned in https://github.com/actions/runner-images/issues/8659#issuecomment-1780570741

I can not reproduce the exact workaround on my machine as my Ubuntu version is outdated. But maybe you can try this for your local installation:

After the manual installation of clang 17 (using `llvm.sh`), instruct the build process to specifically use that clang version. The change for the GitHub workflow should be like follows:

```diff
-        export CC=clang
-        export CXX=clang++
+        export CC=clang-17
+        export CXX=clang++-17
```

Or select the default version to be used on your machine:

```bash
sudo update-alternatives --config clang
sudo update-alternatives --config clang++
```

It would be good if both, the hosted workflows and your local installation would be running without errors.

A workaround for the GitHub Action would be to remove GCC13 (like https://github.com/sourcemeta/jsontoolkit/commit/caa66fb2cc9d86beb5dfc5dfa15823c0e42005ee) or to use a different runner like https://github.com/actions/runner-images/issues/8659#issuecomment-1824069260.

PR #804, opened 2023-12-13
variant varadic default constructed state

fermentedfly:
Hi,

It seems variant_varadic still contains valid type_id when default constructed, 
initializing to `variant_npos` should resolve the issue.

Regards,
Manuel

### DIFF
--- a/include/etl/private/variant_variadic.h
+++ b/include/etl/private/variant_variadic.h
@@ -531,7 +531,7 @@ namespace etl
 
       default_construct_in_place<type>(data);
       operation = operation_type<type, etl::is_copy_constructible<type>::value, etl::is_move_constructible<type>::value>::do_operation;
-      type_id   = 0U;
+      type_id   = variant_npos;
     }
 #include "diagnostic_pop.h"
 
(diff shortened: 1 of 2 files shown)


Chiraffollo:
Hi,
when I upgraded the etl library in my local project to the version 2.38.11 from 2.38.10, I stumbled across this ticket.
For std::variant, it is specified that it contains an element of the first type if default constructed and thus the index must be 0. Is there a reason why the etl must differ here? If so, the creation of the first type would have to be removed in the default constructor and it would also be helpful if a note were added to the documentation that this deviates from std::variant and that the class therefore does not behave as expected.

jwellbelove:
_Is there a reason why the etl must differ here?_

You are right, this change does look incorrect.

Chiraffollo:
Should I create a pull request?

jwellbelove:
No need, I'm doing a simple revert, plus a couple of extra tests to check the value initialisation.

Chiraffollo:
Okay, thank you very much! Have a nice weekend.

Issue #811, opened 2023-12-29
Implement SAE J1850 CRC8


drewr95:
I've recently needed to use the j1850 crc8 checksum and figured this library might find use for it as well.

When it comes to SAE, finding detailed public information (free) isn't really easy to come by.  [This post on SO can give some insight.](https://stackoverflow.com/questions/38639423/understanding-results-of-crc8-sae-j1850-normal-vs-zero)

jwellbelove:
Pull request merged

PR #812, opened 2023-12-29
Implement SAE J1850 CRC8 and "Zero" version

drewr95:
As stated in [#811](https://github.com/ETLCPP/etl/issues/811), I've recently needed to use the SAE J1850 CRC8 algorithm.  I figured this would be a nice place to put it.

There is apparently a "Zero" version of the algorithm as well, of which I've included.


jwellbelove:
Thanks, I'll merge that in shortly.

PR #840, opened 2024-02-10
Adding constexpr to intrusive containers

ZachOB:
Adding `constexpr` to the intrusive containers. I was able to build and run tests for c++11, 14, and 17. c++20 had some issues with `char8_t` so I was unable to run those. c++03 was very broken for me but `ETL_CONSTEXPR` should be pre-compiled away  so I don't think this could have broken anything.

### DIFF
--- a/include/etl/intrusive_forward_list.h
+++ b/include/etl/intrusive_forward_list.h
@@ -55,6 +55,7 @@ namespace etl
   {
   public:
 
+    ETL_CONSTEXPR
     intrusive_forward_list_exception(string_type reason_, string_type file_name_, numeric_type line_number_)
       : exception(reason_, file_name_, line_number_)
     {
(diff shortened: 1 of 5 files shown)


jwellbelove:
You need to create new unit tests to test that the use of the `constexpr` changes compile and work correctly.

ZachOB:
Looks like my use of constexpr was a little more hopeful than useful.

Both `intrusive_list` and `intrusive_forward_list` are not literal due to the `clear()` in their destructors. `intrusive_queue` and `intrusive_stack` have no destructors, so can be constexpr, but maybe they should call `clear()` in their destructors?

The only use I can think of for a constexpr queue/stack is data processing in constexpr functions, as having a standalone constexpr queue or stack seems entirely pointless.

Issue #842, opened 2024-02-14
etl::span subspan could statically check extents


mike919192:
Hi, I noticed that the etl::span subspan function does not currently statically check extents.  std::span does check this.

Here is a code demonstration:
```
std::array<int, 10> array {0};

//std::span std_span1 = std::span(array).subspan<15, 1>(); //compile time error
//std::span std_span2 = std::span(array).subspan<5, 8>(); //compile time error

etl::span etl_span1 = etl::span(array).subspan<15, 1>(); //currently no compile time error
etl::span etl_span2 = etl::span(array).subspan<5, 8>(); //currently no compile time error
```

Seems like it should be fairly easy to add.  I tested a modification to add the static checking, at least for c++17 and newer.  Not sure if there is a way to add it to 11 and 14.

```
#if ETL_USING_CPP11
    //*************************************************************************
    /// Obtains a span that is a view from OFFSET over the next COUNT elements of this span.
    //*************************************************************************
    template <size_t OFFSET, size_t COUNT = etl::dynamic_extent>
    ETL_NODISCARD ETL_CONSTEXPR
    etl::span<element_type, COUNT != etl::dynamic_extent ? COUNT : Extent - OFFSET> subspan() const ETL_NOEXCEPT
    {
//ADDED THESE LINES
#if ETL_USING_CPP17
      if constexpr(extent != etl::dynamic_extent)
      {
        static_assert(OFFSET <= extent);
        if constexpr(COUNT != etl::dynamic_extent)
        {
          static_assert(COUNT <= extent);
          static_assert(COUNT <= (extent - OFFSET));
        }
      }
#endif
//END ADDED LINES		
      return (COUNT == etl::dynamic_extent) ? etl::span<element_type, COUNT != etl::dynamic_extent ? COUNT : Extent - OFFSET>(pbegin + OFFSET, (pbegin + Extent))
                                            : etl::span<element_type, COUNT != etl::dynamic_extent ? COUNT : Extent - OFFSET>(pbegin + OFFSET, pbegin + OFFSET + COUNT);
    }
#else
    //*************************************************************************
    /// Obtains a span that is a view from OFFSET over the next COUNT elements of this span.
    //*************************************************************************
    template <size_t OFFSET, size_t COUNT>
    etl::span<element_type, COUNT != etl::dynamic_extent ? COUNT : Extent - OFFSET> subspan() const
    {
      if (COUNT == etl::dynamic_extent)
      {
[...]

PR #843, opened 2024-02-14
In subspan function, add static checks on extents

mike919192:
Potential fix for issue #842 

etl::span subspan function does not currently statically check extents. std::span does check this.

Code demonstration:
```
std::array<int, 10> array {0};

//std::span std_span1 = std::span(array).subspan<15, 1>(); //compile time error
//std::span std_span2 = std::span(array).subspan<5, 8>(); //compile time error

etl::span etl_span1 = etl::span(array).subspan<15, 1>(); //currently no compile time error
etl::span etl_span2 = etl::span(array).subspan<5, 8>(); //currently no compile time error
```

This PR should enable c++11 and newer to have compile error when the extent check fails.

### DIFF
--- a/include/etl/span.h
+++ b/include/etl/span.h
@@ -296,8 +296,8 @@ namespace etl
     template <size_t COUNT>
     ETL_NODISCARD ETL_CONSTEXPR etl::span<element_type, COUNT> first() const ETL_NOEXCEPT
     {
-      //if extent is static, check that original span contains at least COUNT elements
-      ETL_STATIC_ASSERT((extent != etl::dynamic_extent) ? COUNT <= extent : true, "Original span does not contain COUNT elements");
+      //if Extent is static, check that original span contains at least COUNT elements
+      ETL_STATIC_ASSERT((Extent != etl::dynamic_extent) ? COUNT <= Extent : true, "Original span does not contain COUNT elements");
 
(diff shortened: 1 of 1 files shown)


jwellbelove (changes requested):
`static_assert` should be replaced with `ETL_STATIC_ASSERT`.

jwellbelove (changes requested):
The static asserts have no message fields

mike919192:
I also realized that the first and last functions also behave similarly, so I have also added the checks for those functions.  I believe those are all the functions where static checks are possible.

```
std::array<int, 10> array {0};

//std::span std_span1 = std::span(array).first<15>(); //compile time error
//std::span std_span2 = std::span(array).last<15>(); //compile time error

etl::span etl_span1 = etl::span(array).first<15>(); //currently no compile time error
etl::span etl_span2 = etl::span(array).last<15>(); //currently no compile time error
```

I don't know if it makes sense to also add the static assert for the pre c++11 subspan function?  I'll await @jwellbelove feedback.

jwellbelove:
The `ETL_STATIC_ASSERT` will work for pre-C++11 as the macro defines a C++03 static assert implementation.

mike919192:
Thanks.  I duplicated the static asserts for the pre c++11 subspan function.  I should be finished with changes unless there are other issues.

jwellbelove (approved):

Issue #848, opened 2024-02-22
[`string_stream`] Missing initializers for const format specifiers `etl::left` and `etl::right`


KnoerleMaTLS:
The const variables for the format specifiers `etl::left` and `etl::right` which are declared in the file `basic_format_spec.h` which is used by the file `string_stream.h` do not have any initializers.
Therefore some exotic compilers (e.g. TASKING TriCore) raise an error when including the `string_stream.h` file with the following error messages:
- cptc E0811: [".../include/etl/basic_format_spec.h" 176] const variable "etl::left" requires an initializer -- class "etl::private_basic_format_spec::left_spec" has no user-provided default constructor
- cptc E0811: [".../include/etl/basic_format_spec.h" 180] const variable "etl::right" requires an initializer -- class "etl::private_basic_format_spec::right_spec" has no user-provided default constructor

The error message is already pretty descriptive and explains the requirements which are defined by the C++ standard.
Since we are default initializing a const object the following applies:
_"A class type T is const-default-constructible if default-initialization of T would invoke a user-provided constructor of T"_ ([Source](https://en.cppreference.com/w/cpp/language/default_initialization#:~:text=A%20class%20type%20T%20is%20const%2Ddefault%2Dconstructible%20if%20default%2Dinitialization%20of%20T%20would%20invoke%20a%20user%2Dprovided%20constructor%20of%20T))

All other format specifiers in the `namespace private_basic_format_spec` do provide a user defined constructor which initializes their member.
Since the `left_spec` and `right_spec` structs do not have any member they do not provide a constructor, which is the correct solution.

Anyway the standard does not allow to create a const object without initializing it, since it can not be mutated afterwards and therefore will not be in any useful state. But since the struct is empty there is no invalid state.

There is already an open issue about this point of the standard ([CWG 253](https://cplusplus.github.io/CWG/issues/253.html)).
This issue states the following reported behavior:
_"default-initialization of a const object could not call an implicitly declared default constructor"_
The correct behavior would be:
_"allowed if all subobjects are initialized"_

It seems like there are compilers which already implemented their own solution to handle this issue, while there is no official fix of the standard yet.
Therefore there were no issues with the GCC or clang compiler when using the `string_stream.h` file.
Anyway the compiler we are using at our project does not accept this.

There are two ways to fix this which came to my mind:
1) Implement an empty user provided default constructor in `struct left_spec` and `struct right_spec` (not preferred, rule of zero)
[...]

PR #849, opened 2024-02-22
fix(string_stream): add missing initializations of const format specifiers

KnoerleMaTLS:
This pull request introduces a possible fix for the issue #848 by adding the missing initializers.

### DIFF
--- a/include/etl/basic_format_spec.h
+++ b/include/etl/basic_format_spec.h
@@ -172,10 +172,10 @@ namespace etl
   static ETL_CONSTANT private_basic_format_spec::base_spec hex(16U);
 
   //*********************************
-  static ETL_CONSTANT private_basic_format_spec::left_spec left;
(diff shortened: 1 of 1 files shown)


jwellbelove:
Were you getting warning from the compiler?

KnoerleMaTLS:
Warning about what/where?
Regarding the propsed fix?

jwellbelove:
Did you get errors or warnings about 'missing initializers'?
`left_spec` & `right_spec` have default constructors, so they shouldn't need explicit initialisation.

KnoerleMaTLS:
Yeah, i got compiler errors about missing initializers for `left` and `right`.
Please check #848 for further details.

Yes, the structs `left_spec` and `right_spec` have compiler generated default constructors, but regarding the C++ standard this is not enough to initialize a const object (_"default-initialization of a const object could not call an implicitly declared default constructor"_).
I also tried to explain this in more detail in the mentioned issue.

Issue #844, opened 2024-02-15
etl::type_def - constexpr instance


dfalk-solve:
My question here is, if it would be possible to enhance etl::type_def to allow an constexpr instance that just allows reading the value, which could be helpful for a private member definition within a class.

Currently I do the definitions using it the following way, e.g.

Definition in a common header file:
ETL_TYPEDEF(uint32_t, milliSecond_t);   ///< [ms]   Time measurement (1ms = 0.001s)

And in a private section of a class:
static constexpr milliSecond_t::value_type DefaultTimeout { 10 };

So I define a constant for the "::value_type" which is the underlaying 
data type, which is uint32_t And what I would prefer is to define a constant value for the strong type, in this case "milliSecond_t"

static constexpr milliSecond_t DefaultTimeout { 10 };

Would this be possible if defining a constexpr constructor and a "get()" function? If yes, could this be a possible enhancement for one of the next versions of ETL?

drewr95:
I planned on working on this (or if John or someone else beat me to it), see my issue here about the same subject: https://github.com/ETLCPP/etl/issues/838

jwellbelove:
Fixed 20.38.11

Issue #838, opened 2024-02-07
Make strong typedefs able to be constexpr


drewr95:
I noticed that the strong typedefs _ETL_TYPEDEF_ doesn't use constexpr when I think it could.  I think this could additionally be applied to the custom enum macro and user type macros as well.

Example:
```c++
ETL_TYPEDEF(float, seconds);
constexpr seconds duration{1.0F};
``` 

Trying to constexpr a strong typedef results in a compilation error:

> error: constexpr variable cannot have non-literal type 'const seconds' (aka 'const type_def<seconds_tag, float>')

jwellbelove:
Fixed 20.38.11

PR #861, opened 2024-03-12
Implement Constexpr Strong Typedef

drewr95:
For https://github.com/ETLCPP/etl/issues/838 and https://github.com/ETLCPP/etl/issues/844, added the ability to make constexpr strong types.

### DIFF
--- a/include/etl/type_def.h
+++ b/include/etl/type_def.h
@@ -58,38 +58,38 @@ namespace etl
         typedef TIdType id_type;
 
         //*********************************************************************
-        type_def()
+        ETL_CONSTEXPR type_def()
             : value(TValue())
         {
         }
 
         //*********************************************************************
-        type_def(TValue value_)
+        ETL_CONSTEXPR type_def(TValue value_)
             : value(value_)
         {
         }
 
         //*********************************************************************
-        type_def(const type_def& other)
+        ETL_CONSTEXPR type_def(const type_def& other)
             : value(other.value)
         {
         }
 
         //*********************************************************************
-        operator TValue() const
+        ETL_CONSTEXPR operator TValue() const
         {
             return value;
         }
 
         //*********************************************************************
-        type_def& operator ++()
+        ETL_CONSTEXPR type_def& operator ++()
         {
             ++value;
             return *this;
         }
 
         //*********************************************************************
-        type_def operator ++(int)
+        ETL_CONSTEXPR type_def operator ++(int)
         {
             type_def temp(*this);
             type_def::operator ++();
@@ -97,14 +97,14 @@ namespace etl
         }
 
         //*********************************************************************
-        type_def& operator --()
+        ETL_CONSTEXPR type_def& operator --()
         {
             --value;
             return *this;
         }
 
         //*********************************************************************
-        type_def operator --(int)
+        ETL_CONSTEXPR type_def operator --(int)
         {
             type_def temp(*this);
             type_def::operator --();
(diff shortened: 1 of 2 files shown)


include/etl/type_def.h
```diff
@@ -238,63 +238,71 @@ namespace etl
...
-        type_def& operator =(const type_def& rhs)
+        ETL_CONSTEXPR type_def& operator =(const type_def& rhs)
         {
           value = rhs.value;
           return *this;
         }
 
         //*********************************************************************
-        TValue& get()
+        ETL_CONSTEXPR TValue& get()
```
drewr95, line 255: @jwellbelove I believe I may have mistaken this part.  I think for C++11 users this will make the TValue reference be const?  In C++14 they loosened the restrictions on what constexpr meant, so methods didn't have to be necessarily constexpr to have the keyword for it.  I can change this quickly if you suspect this is the case.

jwellbelove: Yes, it is normal to not have `constexpr` on both the const and non-const functions.
I'm pretty sure you don't need the `#if ETL_USING_CPP14`, unless I'm missing something.
I'm also sure that the `operators` can also be constexpr.

drewr95: What would you like?  I thought for CPP11 constexpr was strict?  I can leave the constexpr off the get methods.  As far as operators, I left the assignment operators (*=, +=, -=, ++, --, etc) not constexpr because there wasn't a way for me to test them.  In C++14 it would have been okay to make them constexpr because of the more lax rules.  But I'm pretty sure that would have failed the C++11 actions.  I think i'll just leave the gets as non-constexpr and open a new PR

jwellbelove: I found I get a compile error if I've set both const and non-const versions of a member function `constexpr`.

The operators can be tested by making a unit test that is only enabled for C++14 and above.
The CI that I run locally, and on Github Actions, test with C++11, C++14, C++17 & C++20

drewr95: I will update the tests for C++14 and above, thanks.

jwellbelove: Github CI has flagged some issues.

https://github.com/ETLCPP/etl/actions/runs/8262992076/job/22603500042?pr=862

drewr95: I'm not quite sure how they can be constexpr without specifying `#if ETL_USING_CPP14`, can you clarify?  It's not making sense how the C++11 builds would pass since all the operator *= += -= tests will be non-constexpr.  As you pointed out, I would have to create separate tests for C++14 and above - which is "ETL_USING_CPP14" if I'm not mistaken.  Why would the tests need differentiated but not the implementation?

drewr95: Part of me thinks the assignment operators shouldn't be marked constexpr regardless because you can't manipulate  a constexpr variable.  So I almost feel like the PR should stay the way it is (I removed the constexpr keywords from the get and assignment operators).

jwellbelove: You can able to use `constexpr` operators inside a `constexpr` function that returns a `constexpr` object.

Example
[https://godbolt.org/z/rPqW17xPx](https://godbolt.org/z/rPqW17xPx)

There is more than one `constexpr` macro. 
`ETL_CONSTEXPR` or `ETL_CONSTEXPR11` Defined as `constexpr` for C++11 and above
`ETL_CONSTEXPR14` Defined as `constexpr` for C++14 and above
`ETL_CONSTEXPR17` Defined as `constexpr` for C++17 and above
`ETL_CONSTEXPR20` Defined as `constexpr` for C++20 and above

You use the one that is appropriate for the code.

drewr95: Thank you, much appreciated!  I will get right on that

drewr95: I'm still getting build errors.  I'm kind of stumped.  The lambdas were nice to keep the typedefs local to the test functions.  If I move the lambdas out to constexpr free functions like your godbolt-linked example, then the typedef will need to exist at a scope outside of the tests (which will conflict with the tests).  Unless you're okay with that or have another suggestion

Eplankton: It seems that with ARM GCC 13 and C++23, the compiler will always complain about something like `error: 'ETL_CONSTEXPR14' does not name a type` or like: 
```
/etl/algorithm.h:1777:3: note: the macro 'ETL_NODISCARD' had not yet been defined
./app/./etl/platform.h:290: note: it was later defined here
  290 |   #define ETL_NODISCARD    [[nodiscard]]
```

What should we put in the `etl_profile.h` to fix that?

PR #863, opened 2024-03-13
Add eq and ne operators for compare utility.

Benzinnos:


jwellbelove (changes requested):
Maybe add the '<=>' operator for C++20 too?

jwellbelove:
_So, are we really need spaceship operator? What the use case? We have already all 6 comparing operators._
Well, the struct is called 'compare' and I thought that the three way compare would be the only one missing from the set after this change.

Issue #830, opened 2024-01-25
gcc warning: potential null pointer dereference [-Wnull-dereference] using etl::map::insert


Forlautboy:
Hi,

I have a question regaring an occuring compiler warning.

Compiling with -O2 the gcc warns about potential null pointer dereference [-Wnull-dereference] using the function etl::map::insert.

As far is I understand the warning is raised from Line 1500 of map.h:
::new (&node.value) value_type(value);
Disabeling the warning arount the insert statement has no effect since the warning arises from an instantiated template.

Similar behaviour occures using flat_map but in lien 1032 of flat_map.h:
::new (pvalue) value_type(etl::forward<TValueType>(value));

I just wanted to ask if sombody faced the same effect and maybe has some tips for coming around.
Finally I am just not sure if its really a problem of if gccs warning is just a false positive.

We are using gcc 12.2.0 and etl v20.38.10 with ETL_THROW_EXCEPTIONS disabled.

Kind Regards,

Tobias

jwellbelove:
When exceptions are not used then GCC is expecting that the error handler could return, thereby propagating a `nullptr`. As `insert` can return a indication that the value was not inserted the code should probably be refactored to allow a failed allocation (i.e. pool empty) to result in valid code with the appropriate return value.

Forlautboy:
Ok, I understood. So you suggest that if the error_handler returns, instead of producing undefined behaviour by accessing nullpointer, map::insert to return a nulliterator and false for inserted? That sounds promising. The Standard function std::map::insert returns inserted = false for the case that the key is already present in the map. So the combination of nulliterator and inserted = false for a insert on a full map would still distinguish from the case that the key is already prensent.
I think that could fix our problem. I will try this on a locally and provide further feedback.

Forlautboy:
I tested your suggestion. Although it eliminates the undefined behavior the warnings sill arise. I am still net certain which "pointer" the compiler thinks of dereferencing. I am even a bit suspicious that the warning considers some internal magic of the new-statement.

/usr/include/etl/map.h:1520:7: warning: potential null pointer dereference [-Wnull-dereference]
 1520 |       ::new (&node.value) value_type(value);
      |       ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

=> Line 1500 in unmodified code

/usr/include/etl/map.h:1545:7: warning: potential null pointer dereference [-Wnull-dereference]
[...]

PR #873, opened 2024-03-29
Validate allocated DataNode is not null to prevent gcc -Wnull-dereference (issue #830)

brawner:
gcc appears to warn about not checking for null when allocating these pointers. It is probably a false positive because clang does not issue the same warning, and optimizations can change how gcc issues this warning. Nonetheless, this is one approach to resolving the issue.

Addresses https://github.com/ETLCPP/etl/issues/830

### DIFF
--- a/include/etl/map.h
+++ b/include/etl/map.h
@@ -935,7 +935,9 @@ namespace etl
         ETL_ASSERT(!full(), ETL_ERROR(map_full));
 
         // Get next available free node
-        Data_Node& node = allocate_data_node_with_key(etl::move(key));
+        Data_Node* node = allocate_data_node_with_key(etl::move(key));
+
+        ETL_ASSERT(node != ETL_NULLPTR, ETL_ERROR(map_full));
 
         // Obtain the inserted node (might be ETL_NULLPTR if node was a duplicate)
         inserted_node = insert_node(root_node, node);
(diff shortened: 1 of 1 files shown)


jwellbelove:
There are errors in the CI
Example
```
/home/runner/work/etl/etl/test/../include/etl/map.h:1226:32: error: comparison of distinct pointer types ('etl::map_base::Node *' and 'etl::imap<std::basic_string<char>, int, std::greater<std::basic_string<char>>>::Data_Node **')
      inserted = inserted_node == &node;
```

jwellbelove:
I've investigated this some more and found a pattern to the warning.
Many ETL containers use a function called `allocate_data_node()` to allocate a node from an `etl::ipool`. Only in the `set`/`multiset`/`map`/`multimap` containers does it return a reference; in all the others it returns a pointer. In all of the containers the function that call it return a reference (they don't seem to exhibit the warning), so the solution appears to be to just modify the `allocate_data_node()` in `set`/`multiset`/`map`/`multimap`. I've already applied this change to my issue branch.

brawner:
@jwellbelove Did you want me to fix this PR, or are you handling it on your branch?

jwellbelove:
I've tested my changes with my CI scripts here, so I'll merge them. Thanks for your PR though. All inputs are always appreciated.

brawner:
I can confirm that your fixed version appears to work with my code as well.

Issue #823, opened 2024-01-22
Multiple variable shadowing warnings in private/bitset_new.h


digimokan:
When compiling with -Wshadow, there are multiple warnings in private/bitset_new.h.

There is one warning for the `lsb` variable shadowing:

https://github.com/ETLCPP/etl/blob/master/include/etl/private/bitset_new.h#L643

And the rest of the warnings come from `value` variable shadowing:

https://github.com/ETLCPP/etl/blob/master/include/etl/private/bitset_new.h#L190

https://github.com/ETLCPP/etl/blob/master/include/etl/private/bitset_new.h#L532

https://github.com/ETLCPP/etl/blob/master/include/etl/private/bitset_new.h#L766

https://github.com/ETLCPP/etl/blob/master/include/etl/private/bitset_new.h#L974

https://github.com/ETLCPP/etl/blob/master/include/etl/private/bitset_new.h#L1038

https://github.com/ETLCPP/etl/blob/master/include/etl/private/bitset_new.h#L1048

https://github.com/ETLCPP/etl/blob/master/include/etl/private/bitset_new.h#L1855

https://github.com/ETLCPP/etl/blob/master/include/etl/private/bitset_new.h#L1926

https://github.com/ETLCPP/etl/blob/master/include/etl/private/bitset_new.h#L2642

https://github.com/ETLCPP/etl/blob/master/include/etl/private/bitset_new.h#L2651

https://github.com/ETLCPP/etl/blob/master/include/etl/private/bitset_new.h#L2760

https://github.com/ETLCPP/etl/blob/master/include/etl/private/bitset_new.h#L3530

https://github.com/ETLCPP/etl/blob/master/include/etl/private/bitset_new.h#L3520

https://github.com/ETLCPP/etl/blob/master/include/etl/private/bitset_new.h#L3641

jwellbelove:
I'm currently refactoring `bitset_new.h`, so I'll look at these changes while I'm at it.

jwellbelove:
Fixed 20.38.11

PR #881, opened 2024-04-11
Rename successor to avoid shadowing

senceryazici:
Some of the constructors in [message_router.h](https://github.com/ETLCPP/etl/blob/master/include/etl/message_router.h) are using arguments named `successor`, resulting in following warning/error message:

```
etl/message_router.h:296:5: error: declaration of ‘successor’ shadows a member of ‘etl::message_producer’
```

This PR renames arguments to `successor_` to prevent shadowing.

### DIFF
--- a/include/etl/message_router.h
+++ b/include/etl/message_router.h
@@ -200,8 +200,8 @@ namespace etl
     }
 
     //********************************************
-    null_message_router(etl::imessage_router& successor)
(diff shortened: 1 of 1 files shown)


jwellbelove:
Thanks for the PR, but this has already been addressed in commit 77513730e09b742a2fc9ad5ecb0ebb7b78b54a7f in response to issue #823

This change will be released soon.

PR #905, opened 2024-06-10
etl/type_traits.h: Add support for type_identity

tigran2008:


jwellbelove:
Did you add a unit test?

tigran2008:
Sorry, I overlooked that part :/

I'm away from my computer right now, but would the following be enough if I added it later?

Given this function:

```cpp
template <typename T>
T add(T first, typename etl::type_identity<T>::type second) {
    return first + second;
}
```

The result of `add(2.5, 1)` would be checked if it was equal to `3.5`

jwellbelove:
That looks OK.

test/test_type_traits.cpp
```diff
@@ -1320,4 +1327,9 @@ namespace
     CHECK_FALSE(bool(etl::is_base_of_all<Base, D1, D2, D3, D4>::value));
 #endif
   }
+
+  //*************************************************************************
+  TEST(test_type_identity) {
+    CHECK(type_identity_test_add(1.5f, 2) == 3.5f);
```
jwellbelove: You should never test floating point numbers for equality.
Use `CHECK_CLOSE` with a small tolerance value, such as `0.01f`.

tigran2008: Why not? I used 0.5 as the fraction, which is a negative power of two, so no rounding is involved.

tigran2008:
Well, I changed the CHECK into a CHECK_CLOSE.

test/test_type_traits.cpp
jwellbelove: It's just standard good practice.
It protects against code breaking if someone were to change a value in the future without realising the implications.

PR #912, opened 2024-07-02
Improve ci

### LINKED ISSUE #911: Speed up builds
Something I noticed is that current GHA workflows simply call `make` - which defaults to running the jobs sequentially. Meanwhile, GitHub's public runners [have more than one CPU core](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories).

Your builds could probably be significantly sped up by changing the `make` call to something like `make -j4`, or more universally `make -j $(nproc)`. It could also be possible to switch over to ninja (which defaults to one job per CPU thread), but I don't think it's installed by default in GHA images.

jaskij:
1. Remove duplication in workflow definitions by using matrices. Should make the inline display in PR more readable too.
2. Tell `make` to use all the CPUs. Seems to roughly halve the build time.

Closes #911 

Also: due to the nature of the changes, my branch will have a lot of noise in commits, so I *highly* recommend squash merging.

-----

Subtasks:

- [X] Linux builds - done
- [X] macOS builds - done - needs discussion
- [X] Windows - done
- [ ] caching - left for future
- [X] explore `cmake --build` - won't gain anything with it

----

Before merging:

- [ ] `-Wno-vla-cxx-extension`
- [X] restore branch filters


jaskij:
Example run: https://github.com/jaskij/etl/actions/runs/9768833717

@jwellbelove questions:

1. For C++03, the old workflows compiled, but didn't run tests, was this intentional?
2. C++98 didn't seem to even be compiled at all?
3. In the current version, all the GCC workflows run on Ubuntu 24.04, with GCC 14, is this ok? I could add Unbuntu 22.04 for all tests except C++23, but that would add another eighteen tests and I'm not sure it's necessary. If anything, I'd prefer running tests on the minimum GCC version you support.

jwellbelove:
The C++03 workflows are only active for the 'syntax-checks'.
The actual unit tests are only valid for C++11 and above.
C++98 & C++03 are treated as being the same. The CMakeLists.txt file uses the C++98 flag for both options.
I just run with whatever GCC/Clang version supports all of the features in the ETL.

jaskij:
> The C++03 workflows are only active for the 'syntax-checks'.
> The actual unit tests are only valid for C++11 and above.

That's interesting, because I actually enabled running tests for C++03 and it seems to work. See for example [this run](https://github.com/jaskij/etl/actions/runs/9769193030/job/26968182170). Only did that because it was easier than special-casing 03 to not run the tests.

>  C++98 & C++03 are treated as being the same. The CMakeLists.txt file uses the C++98 flag for both options.

Okay, so separate C++03 and C++98 checks are not necessary.

>  I just run with whatever GCC/Clang version supports all of the features in the ETL.

It could be worthwile for some users to test on older GCC or clang versions, but since you currently do not specify minimum supported compiler version, I'll leave it running as is until a change is necessary. It will be much easier in the future as well.

----

Do you mind if I upgrade the MacOS runner from `macos-12` to `macos-14`?

jaskij:
Hmm, it seems that LLVM 18 [defaults to warning on VLAs in C++](https://github.com/llvm/llvm-project/issues/62836), which leads to [some tests not compiling](https://github.com/jaskij/etl/actions/runs/9769384134/job/26968716731). What do you want me to do with that?

jwellbelove:
I'm not sure how the unit tests compile for C++03 as, for example, the tests for `etl::foreward_list` compares against the output of `std::forward_list`, which was only available from C++11.

If I get an issue raised about a compatibility issue with an older compiler, then I address it at that point. I've not had very many raised over the years. The most problems that have come up have been with people using STLPort.

I have no issue with upgrading the MacOS version.

jaskij:
> I'm not sure how the unit tests compile for C++03 as, for example, the tests for `etl::foreward_list` compare against the output of `std::forward_list`, which was only available from C++11.

They don't. The matrix definition in the workflow treats `03` as a number, and drops the leading zero. Then it goes through the [`if`-`elseif` tree in tests `CMakeLists.txt`](https://github.com/ETLCPP/etl/blame/c959c932a6edab7f9de59c4eb4d114f76f36ad3d/test/CMakeLists.txt#L309-L327) falling through all the way to C++20.

jaskij:
Right now, I all the Linux and MacOS builds in a single file, you can [view a run here](https://github.com/jaskij/etl/actions/runs/9770557035), although MacOS seems to be having conniptions right now.

That's the state I'm leaving it in, I'll take a stab at MSVC on a different day. Might be a touch harder, since I haven't used it in a decade.

The current workflow file is much more complicated than it used to be, but I believe should still be quite readable even without actually looking up the documentation. The one exception are [matrices](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs) but the documentation for them seems good enough.

jaskij:
I accidentally dropped syntax checks, which are now restored.

Meanwhile, I added a [somewhat hacky fix in `platform.h`](https://github.com/ETLCPP/etl/pull/912/commits/6f3b4687e411975c8638cd36a8b0cbcebf6548c4) to get C++03 syntax checks to pass on MacOS.

Current run: https://github.com/jaskij/etl/actions/runs/9771327648/job/26973921406

jaskij:
Sorry for the comment span, but that's probably it for now.

@jwellbelove can you take a look at this? Seems like `multiset.h` is picking `std::` improperly here, with `NO_STL=ON`. https://github.com/jaskij/etl/actions/runs/9771327648/job/26973923356

jaskij:
Thinking on this somewhat more, I'm inclined to have macOS tests as a separate workflow, and just let them fail for now.

jaskij:
Okay, I believe this is as far as this PR should go. I did essentially two things:

- deduplicated job definitions using matrices
- sped up the builds by running them in parallel

The part about running in parallel is only try for Linux and macOS - to do the same for Windows/MSVC, I would need to change the build generator/executor from MSBuild.exe to make or something else, and that isn't a change I wanted to make right now.

Overall, execution times do seem to have fallen reasonably. The results will also be easier to view in GitHub's PRs.

What I didn't do - as it should be a separate PR - is C++ 23 enablement. With the jobs deduplicated as they are now, enabling C++23 builds should be a very simple change.

Regarding specific OSes:

- **Linux**: this went more or less without issues, as it is also the environment I'm most familiar with.
- **macOS**: it's similar enough to Linux for the CI job to work, but macOS clang is troublesome with C++03 and in no-STL builds, so I have left them disabled
- **Windows**: surprisingly painless, but MSBuild does not support parallelizing within a project, so there was no speed gain

Future possibilities:

- C++ 23 enablement in build system and tests
- switching the MSVC job from MSBuild to something else
- caching

jaskij:
@jwellbelove one last thing: are you fine with disabling the `-Wvla-cxx-extension` warning in tests? I did it because some tests did not compile under newer clang with it enabled, but I'm not sure if this is something you want permanently.

Commit: https://github.com/ETLCPP/etl/pull/912/commits/38fbc84a1a1211367f6d41f841fd5593702e4d0c

jwellbelove:
Is that a warning that you are using VLAs?
What code is it warning about?

jaskij:
Specifically, it's a warning that using VLAs in C++ is a non-standard extension. The code it warns about is stuff like this, in several tests:

```cpp
    TEST(test_rotate_left8_constexpr)
    {
      char temp[etl::rotate_left(uint8_t(0xAAU))];

      CHECK_EQUAL(etl::rotate_left(uint8_t(0xAAU)), sizeof(temp));
    }
```

Which, now that I look at it, seems to have been a mistake on my part - the warning is absolutely correct, as this test should not be compiled for C++ 11.

jaskij:
Do you want me to expand the guards for those tests so they are only compiled if `ETL_USUING_cpp14` is defined?

jaskij:
Here's the CMake command line from a [run that fails on this](https://github.com/jaskij/etl/actions/runs/9770208143/job/26971045143):

```
  cmake .. \
    -DBUILD_TESTS=${TESTS} \
    -DNO_STL=${NO_STL} \
    -DETL_USE_TYPE_TRAITS_BUILTINS=OFF \
    -DETL_USER_DEFINED_TYPE_TRAITS=OFF \
    -DETL_FORCE_TEST_CPP03=${FORCE_03} \
    -DETL_CXX_STANDARD=${CXX_VER}
  shell: /usr/bin/bash -e {0}
  env:
    ASAN_OPTIONS: alloc_dealloc_mismatch=0,detect_leaks=0
    CC: clang
    CXX: clang++
    CXX_VER: 11
    FORCE_03: OFF
    NO_STL: OFF
    TESTS: YES
```

jwellbelove:
Is there any way to replace the 'ON, OFF' in the Jobs summary with something more descriptive?

jaskij:
The easiest way to do so would be to replace that with `NO_STL=ON`/`NO_STL=OFF` and `ETL_FORCE_TEST_CPP03_IMPLEMENTATION=ON`/`ETL_FORCE_TEST_CPP03_IMPLEMENTATION=OFF`, but those would probably be too long.

It can be more freeform, for example `NO_STL`/`STL` and `FORCE_O3`/`NO_FORCE_03`, or whatever else. Just needs two values. This would make the job definition mildly more complicated, in expanding the `include:` part of the job matrix definition. No new GitHub workflow features would be used.

jaskij:
@jwellbelove can you let me know which of the two options for making the summary more readable you prefer?

Issue #922, opened 2024-07-16
Does ETL need a `nullptr_t` type?


tigran2008:
Seeing issue #921, I noticed that ETL doesn't have a `nullptr_t`. Are you for or against having an `etl::nullptr_t` type, and if you wouldn't mind having it, what do you think of implementing <etl/nullptr.h> like this instead?

```cpp
namespace etl
{
#if ETL_CPP11_NOT_SUPPORTED
  // Use the old style C++ NULL definition.
  typedef enum { _nullptr = 0 } nullptr_t;
  #define ETL_NULLPTR (etl::_nullptr)
#else
  // Use the new style nullptr.
  typedef decltype(nullptr) nullptr_t;
  #define ETL_NULLPTR nullptr
#endif
} // namespace etl
```

PR #924, opened 2024-07-17
Add an etl::nullptr_t type to <etl/nullptr.h>

tigran2008:
Well, the title is self-explanatory. This is a tiny change anyway, and the code is basically what's proposed in #922.

Edit: As of this moment, I changed the pre-C++11 code from an enum into a  custom class

### DIFF
--- a/include/etl/memory.h
+++ b/include/etl/memory.h
@@ -1393,7 +1393,7 @@ namespace etl
     }
 
     //*********************************
-    void reset(pointer p_ = pointer()) ETL_NOEXCEPT
+    void reset(pointer p_) ETL_NOEXCEPT
     {
       assert(p_ != p);
 
@@ -1404,6 +1404,12 @@ namespace etl
       {
         deleter(value);
       }
+    }
+
+    void reset(etl::nullptr_t p_ = ETL_NULLPTR) ETL_NOEXCEPT
+    {
+      (void)p_; // unused argument
+      reset(pointer());
     }
 
     //*********************************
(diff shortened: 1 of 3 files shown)


tigran2008:
I'm so sorry for the mistakes, I'm a total beginner to C++ and I'm trying my best :/

tigran2008:
After I'm done, I promise I'll squash the commits 😅

jwellbelove:
You can run the unit tests locally if you want to check before you push, though it does take some time to do all C++11/14/17/20 variants.

tigran2008:
Do you think this implementation of etl::nullptr_t is acceptable? It instantiates a `_nullptr` instance beforehand, and it occupies one byte of storage, unlike the previous one, which didn't occupy any. I could also do `#define ETL_NULLPTR (etl::nullptr())` but I thought it wouldn't be good to instantiate a new object everywhere a nullptr is used..

Also, if you do think it's an acceptable solution, do you think I should add a `(void)_nullptr; // silence unused warning`? I didn't get an unused variable warning on my computer (gcc version 13.3.1 20240522 (Red Hat 13.3.1-1) (GCC) on Fedora 39) but I just noticed that it warns on https://cpp.sh

tigran2008:
Do you think I should also delete etl::addressof(ETL_NULLPTR)?

tigran2008:
Well, I'll be waiting for a comment from you before squashing the commits

jwellbelove:
I'll have a look tomorrow.

include/etl/memory.h
```diff
@@ -1393,7 +1393,7 @@ namespace etl
     }
 
     //*********************************
-    void reset(pointer p_ = pointer()) ETL_NOEXCEPT
+    void reset(pointer p_) ETL_NOEXCEPT
```
jwellbelove: According to [cppreference](https://en.cppreference.com/w/cpp/memory/unique_ptr/reset), the signature for this member function is `void reset( pointer ptr = pointer() ) noexcept;`

tigran2008: Right. I'll fix that later today

PR #929, opened 2024-07-19
Fix run-tests.sh script

Chiraffollo:
- Fix ordering of parameters in help info
- Change spaces to tabs in indentation of methods (unique usage of indentation)
- Fix check for number of parameters to avoid empty string as default parameter value

### DIFF
--- a/test/run-tests.sh
+++ b/test/run-tests.sh
@@ -142,7 +142,7 @@ fi
 #******************************************************************************
 # Set the number of concurrent processes to use.
 #******************************************************************************
-if [ $# -le 3 ]; then
+if [ $# -le 2 ]; then
   export CMAKE_BUILD_PARALLEL_LEVEL=4
 else
   export CMAKE_BUILD_PARALLEL_LEVEL=$3
--- a/test/run-tests.sh
+++ b/test/run-tests.sh
@@ -21,7 +21,7 @@ NoColour='\033[0m'
 
 ParseGitBranch() 
 {
-    git rev-parse --abbrev-ref HEAD
+	git rev-parse --abbrev-ref HEAD
 }
 
 SetConfigurationName()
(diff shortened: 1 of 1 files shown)


test/run-tests.sh
```diff
@@ -142,7 +142,7 @@ fi
 #******************************************************************************
 # Set the number of concurrent processes to use.
 #******************************************************************************
-if [ $# -eq 2 ]; then
+if [ $# -le 3 ]; then
```
jwellbelove: This does not work.
`./run-tests.sh 11 0 7` sets `CMAKE_BUILD_PARALLEL_LEVEL` to `4`

jwellbelove: I've found this seems to cover all parameter lengths

```
#******************************************************************************
# Set the language standard. Default C++20
#******************************************************************************
if [ $# -eq 0 ]; then
  cxx_standard="20"
elif [ "$1" = "20" ]; then
  cxx_standard="11"
elif [ "$1" = "14" ]; then
  cxx_standard="14"
elif [ "$1" = "17" ]; then
  cxx_standard="17"
elif [ "$1" = "20" ]; then
  cxx_standard="20"
else
  PrintHelp
  exit
fi

#******************************************************************************
# Set the optimisation level. Default -O0
#******************************************************************************
if [ $# -eq 1 ]; then
  opt="-O0"
elif [ "$2" = "1" ]; then
  opt="-O1"
elif [ "$2" = "2" ]; then
  opt="-O2"
elif [ "$2" = "3" ]; then
  opt="-O3" 
else
  opt="-O0"
fi

#******************************************************************************
# Set the number of concurrent processes to use. Default 4
#******************************************************************************
if [ $# -le 2 ]; then
  export CMAKE_BUILD_PARALLEL_LEVEL=4
else
  export CMAKE_BUILD_PARALLEL_LEVEL=$3
fi
```

Chiraffollo: Your right. I should not try to improve my fix after testing... I will fix it.

Issue #937, opened 2024-07-30
ETL_USING_LEGACY_VARIANT Is always getting evaluated to 1


drewr95:
It seems that ETL_USING_LEGACY_VARIANT is always getting evaluated to 1 for me.  I do not have ETL_USE_LEGACY_VARIANT defined, so the !ETL_USING_CPP11 must be getting evaluated to true.  I believe this is because these macros are defined before determining compiler versions.
 
![image](https://github.com/user-attachments/assets/bdf67fd0-c56f-45b6-af63-df31e214db51)

drewr95:
Looks like VSCode is confirming this:
![image](https://github.com/user-attachments/assets/4d16b4dd-c36f-4d8c-a66c-fb9cf9f86fa0)


If we don't determine compiler version til line 153 (profiles/determine_compiler_version.h), how can ETL_USING_CPP11 be evaluated?

drewr95:
Yeah, can confirm, if I move that macro below determining compiler version my builds work again

jwellbelove:
Fixed 20.39.3

PR #938, opened 2024-07-30
fix: determine legacy variant after determining compiler version

drewr95:
Fixes the issue reported here: https://github.com/ETLCPP/etl/issues/937

### DIFF
--- a/include/etl/platform.h
+++ b/include/etl/platform.h
@@ -139,6 +139,15 @@ SOFTWARE.
 #endif
 
 //*************************************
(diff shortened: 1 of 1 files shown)


drewr95:
@jwellbelove @jwellbelove what's the process for hotfixes like these?  Will it get directly back into master or will it not be available until next release and sit in a feature branch?  I don't like pointing at my own fork but will if it's needed

jwellbelove:
I am currently gathering up all of the features and hotfixes to release under a new release label.
If there are no last minute issues or bugs, then I will officially create a new release tomorrow morning.

Issue #941, opened 2024-08-02
Add templated ticker type for callback timers


Sedictious:
Currently the callback timers support a fixed timer type of `uint32_t` which fixes the resolution. I think it would be useful to add a template parameter to allow for configurable timer size.

I was thinking about adding something like `timer_size` along the lines of what is supported for `etl::queue` with `memory_model` which could default to `uin32_t`, thus not breaking any backwards compatibility. Would that be something of interest or is there any limiting factor I am missing?

jwellbelove:
I'll take a look and see if it's viable.

Sedictious:
@jwellbelove I am dealing with the change locally so it would be possible for me to clean up and and prepare a PR  following the style of `memory_model `  if it would help

jwellbelove:
OK, give it a go. I'll be interested in what you come up with.
Don't forget there are multiple timer classes.

PR #943, opened 2024-08-02
[WIP] implementation of templated timer tick

Sedictious:
Addresses #941 

Opening this now before continuing with the other timer services because I feel the approach is perhaps breaking too many things.

This adds `timer_model.h` (name up to change, just wanted to keep it aligned with memory model) that is then used to determine the appropriate type for the ticker, to not limit the resolution to 32bits. I implemented this only for `message_timer.h` as a proof of concept but the concept can be applied to the rest of the services 

The main issue I have with this approach, is that it requires templating `timer` since `interval` and `state` depend on `TIMER_MODEL`. I couldn't find a way to abstract this without some major refactoring and without not templating `timer`. If this is a breaking change and there are no proposed alternative and more elegant solutions I'm missing, feel free to reject and close this PR

Otherwise, I can work on cleaning this up and working on the rest of the timers

### DIFF
--- a/examples/ArmTimerCallbacks - C++/main.cpp
+++ b/examples/ArmTimerCallbacks - C++/main.cpp
@@ -36,10 +36,10 @@ const int N_TIMERS = 4;
 
 etl::callback_timer<N_TIMERS> callback_timer;
 
-etl::timer::id::type short_toggle;
-etl::timer::id::type long_toggle;
-etl::timer::id::type start_timers;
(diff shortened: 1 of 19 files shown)


Sedictious:
@jwellbelove Small bump for this because I'm still up for making the changes for all the timers if needed. If you think this is breaking backwards compatibility in an unacceptable way and don't have any clearly alternative in mind I could cancel the PR  :/

PR #946, opened 2024-08-07
Make include paths to private files relative

dhebbeker:
Analog to #879 

By removing `etl/` from include paths (in `#include` statements), the path to the directory `include/` doesn't need to be provided to the preprocessor as an include path if the files in `include/etl/` are included by other means. This has no disadvantages.
Actually the form `#include "..."` is intended to be used for relative paths in the first place.

This is relevant if one wants to include the source files from `include/etl` only indirectly.

For example we use special generated header files which wrap the include statement of the vanilla header files with diagnostic commands. Those commands disable some compiler diagnostics for ETL's files and re-enable them after the file inclusion. Wrapper files are generated for every non-private header file. We provide the directory with the wrapper files as include path to the preprocessor instead of the normal `include/` directory. Thus include statements like `#include "etl/private/..."` are invalid in that case.

This may be beneficial only in special cases like ours. But it should have no disadvantages. In contrary, it uses the `#include "..."` form as intended.

### DIFF
--- a/include/etl/base64.h
+++ b/include/etl/base64.h
@@ -26,13 +26,13 @@ SOFTWARE.
 #ifndef ETL_BASE64_INCLUDED
 #define ETL_BASE64_INCLUDED
 
-#include "etl/platform.h"
-#include "etl/static_assert.h"
-#include "etl/exception.h"
-#include "etl/error_handler.h"
-#include "etl/type_traits.h"
-#include "etl/enum_type.h"
-#include "etl/integral_limits.h"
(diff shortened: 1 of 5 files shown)


jwellbelove:
There are 622 lines in 63 files where this file path is used.

dhebbeker:
I added 08365d0ec7088bedbe9fe9aadc5ca7b0851ffb2c to accommodate all findings of `"etl/"` in all files which are in `include/`. For other files this should not be (as) relevant.

PR #964, opened 2024-10-01
Use constexpr for basic_string

ebmmy:
This pull request intends to align the use of `constexpr` with how it is defined for `std::basic_string' in the STL.

### DIFF
--- a/include/etl/basic_string.h
+++ b/include/etl/basic_string.h
@@ -173,7 +173,7 @@ namespace etl
     /// Gets the current size of the string.
     ///\return The current size of the string.
     //*************************************************************************
-    ETL_CONSTEXPR20 size_type size() const
+    ETL_CONSTEXPR14 size_type size() const
     {
       return current_size;
     }
@@ -182,7 +182,7 @@ namespace etl
     /// Gets the current size of the string.
     ///\return The current size of the string.
     //*************************************************************************
-    ETL_CONSTEXPR20 size_type length() const
+    ETL_CONSTEXPR14 size_type length() const
     {
       return current_size;
     }
@@ -191,7 +191,7 @@ namespace etl
     /// Checks the 'empty' state of the string.
     ///\return <b>true</b> if empty.
     //*************************************************************************
-    ETL_CONSTEXPR20 bool empty() const
+    ETL_CONSTEXPR14 bool empty() const
     {
       return (current_size == 0);
     }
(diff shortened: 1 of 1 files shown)


jwellbelove:
I'm pretty sure that this should work for C++14 constexpr, unless you know of any C++17/20 features that are required.

Also, you have not created any unit tests to check that all of the constexpr functions can be used in a constexpr expression.

ebmmy:
> I'm pretty sure that this should work for C++14 constexpr, unless you know of any C++17/20 features that are required.

Right, `ETL_CONSTEXPR14` should work fine, I was not sure whether your policy was to conform to the C++ standard or use the first possible version for constexpr.

> Also, you have not created any unit tests to check that all of the constexpr functions can be used in a constexpr expression.

What would be the purpose of such tests? This wouldn't really test a functionality, either the function isn't marked as `constexpr', the expression wouldn't compile, or it is marked as `constexpr' and it can be used inside a `constexpr' expression. What do you think?

jwellbelove:
The purpose of the `constexpr` tests is to make sure that the compiler is happy that a function marked as `constexpr` can actually work in a `constexpr` expression.

ebmmy:
> The purpose of the `constexpr` tests is to make sure that the compiler is happy that a function marked as `constexpr` can actually work in a `constexpr` expression.

Isn’t that granted by the c++ standard directly? I cannot really think of a case this isn’t true.

jwellbelove:
Certain compilers will only analyse the template code for what you actually invoke. For example, Microsoft's compiler will not alert that a template function cannot be used in a `constexpr` unless you actually create code that uses it in a `constexpr`.

Example:

```
template <typename T>
constexpr int GetValue(T i)
{
  static int si = 2;

  return i + si;
}

int main()
{
  constexpr int x = GetValue<int>(1); // This is a compile error for C++14 with VS2022

  int y = GetValue<int>(1);           // This is not a compile error for C++14 with VS2022

  return 0;
}
```

jwellbelove:
What about the constexpr support for the top level classes `etl::string`, `etl::wstring`, `etl::u16string`, `u32string`?

jwellbelove:
The unit tests fail for C++11.

ebmmy:
Hello!
Thanks a lot for the explanation about VS compiler, I'm never using it. Also sorry for the long delay I was away.

> What about the constexpr support for the top level classes etl::string, etl::wstring, etl::u8string, etl::u16string, u32string?

I think I need to have a deeper look for this, becquse the `ibasic_string` type is not litteral hence we cannot declare a constexpr variable for those derived classes. I'll try to think on how to rework the PR for this.

jwellbelove:
I had issues trying to make `etl::bitset` `constexpr`.
In the end, I had to discard using `etl::ibitset` and pass the buffer with every top-level member function to a private implementation class. I don't want to have to do that with the general ETL containers.

PR #965, opened 2024-10-01
etl/delegate: fix accident creation of a delegate to an rvalue delegate when copying/assigning from delegate with mismatching signature

VladimirP1:
Original problem, how to reproduce:
```
#include <etl/delegate.h>

namespace ctx {
struct os {};
struct any {
  any() = default;
  any(os) {}
};
} // namespace ctx

struct Test {
    etl::delegate<void(ctx::os)> GetCb() {
      return etl::delegate<void(ctx::any)>::create<Test,
      &Test::CbAnyCtx>(
          *this);
    }

private:
  void CbAnyCtx(ctx::any ctx) {}
};

int main() {
  Test().GetCb()(ctx::os{});
  etl::delegate<void(void)> f = []() {};
  return 0;
}
```

This compiles and does not look suspicious at the first glance but results in a segmentation fault at runtime.

What happens here:
1) A delegate is created `etl::delegate<void(ctx::any ctx)>::create<Test, &Test::CbAnyCtx>(*this);`
2) But we need a etl::delegate<void(ctx::os ctx)>, the compiler selects the constructor from `const TLambda&`.
3) Note that `etl::is_same<etl::delegate<TReturn(TParams...)>, TLambda>::value` is false because the signatures of the delegates do not match
4) The newly created delegate is returned and the first delegate is destroyed. Now the second delegate holds a dangling pointer

This PR attempts to fix this issue by using a reliable method for checking if a functor is a delegate

### DIFF
--- a/include/etl/private/delegate_cpp03.h
+++ b/include/etl/private/delegate_cpp03.h
@@ -168,6 +168,23 @@ namespace etl
     }
   };
 
+  //*****************************************************************
+  /// The tag to identify an etl::delegate.
+  ///\ingroup delegate
+  //*****************************************************************
+  struct delegate_tag
+  {
+  };
+
+  //***************************************************************************
+  /// is_delegate
+  //***************************************************************************
+  template <typename T>
+  struct is_delegate
+  {
+    static const bool value = etl::is_base_of<delegate_tag, T>::value;
+  };
+
   //*************************************************************************
   /// Declaration.
   //*************************************************************************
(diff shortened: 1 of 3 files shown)


jwellbelove:
Thanks, I'll take a look at that when I get back from Spain later this week.

jwellbelove (commented):
Does class delegate<void> have any use apart from being a base class 'tag'?
It's mainly a style thing, but when I've done something like this before I've usually just created a simple 'tag' class.

`struct delegate_tag {};`

`class delegate<void>` sort of looks like it should do something useful as a delegate type.

VladimirP1:
No other use, I simply did not want to pollute the namespace with additional classes

jwellbelove:
Yes, I think so.

jwellbelove:
It may be good to make a traits class.

template <typename T>
etl::is_delegate;

jwellbelove:
Just thought of another addition.

`etl::is_delegate_v` definition. 
For C++17 only.

PR #985, opened 2024-12-03
Add uncopyable.h, class uncopyable

rolandreichweinbmw:


rolandreichweinbmw:
I don't see a way to add a unit test for this feature. It's a convenience feature which being applied incorrectly would not compile.

rolandreichweinbmw:
If you prefer a unit test including a dummy use of this class (non-functional feature), I can add it.

include/etl/uncopyable.h
```diff
@@ -0,0 +1,56 @@
...
+  /// Can be used to make a class uncopyable by deriving from it.
+  //***************************************************************************
+  class uncopyable
+  {
+  public:
+    uncopyable(uncopyable const&)            = delete;
+    uncopyable& operator=(uncopyable const&) = delete;
+
+  protected:
+    uncopyable() = default;
```
jwellbelove, line 50: I've see comments online that Scott Meyers also adds the destructor in his 'Effective C++' book.

rolandreichweinbmw: I.e. a non-virtual public "=default" destructor?

Is there a reasoning behind it?

jwellbelove: It looks like it is intended to make the destructor `protected`, rather than default `public`.

jwellbelove:
I have sometimes added a commented out unit test that contains 'uncompilable' code, with a surrounding comment that the test should fail to compile if uncommented.
It can be useful if you want to check that the disabled functionality is _still_ disabled after a code change.

jwellbelove:
I also sometimes do this for `static_assert` checks.

rolandreichweinbmw:
I just added the destructor as discussed, and unit tests, including disabled code that intentionally wouldn't compile.

include/etl/uncopyable.h
rolandreichweinbmw: Added the destructor.

PR #990, opened 2024-12-07
Add contains() method to etl::unordered_map and etl::unordered_set

adejewski:
Hi,

I've noticed we are missing contains method in etl::unordered_map and etl::unordered_set, this PR fixes that. What is more, I think we are missing transparent comparators in both of those classes, I can look into that later if you want.


jwellbelove:
Are you able to add the same changes to `etl::unordered_multimap` and `etl::unordered_multiset`?

jwellbelove:
I am already working on transparent comparitors for the unordered containers.

adejewski:
Added `contains()` method to other structures also fixed the parameter type and used the same as `find()` method uses. I think now it is correct.

PR #1006, opened 2024-12-31
Fix arm64 signed char

rolandreichweinbmw:
On arm64, char is unsigned.

That's why etlcpp tests don't compile on arm64, see e.g. https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1091157

This pull request fixes the issue by making chars explicitly signed where necessary. Tested on Debian sid arm64.

### DIFF
--- a/test/test_bit_stream.cpp
+++ b/test/test_bit_stream.cpp
@@ -930,8 +930,8 @@ namespace
     //*************************************************************************
     TEST(put_get_multiple_variable_size)
     {
-      char c1 = 26;             // 6 bits
-      char c2 = -10;            // 7 bits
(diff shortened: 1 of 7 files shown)


jwellbelove:
I've just noticed that you've changed instances of `int8_t` to `signed char`.
As both are signed types, what was the purpose of the change?

rolandreichweinbmw:
Which lines are you referring to for change of `int8_t `to `signed char`?

I just changed `char `to `signed char` as described initially.

jwellbelove:
OK, I'll check tomorrow to see if I made that change myself at some point since the last merge into master.

Issue #666, opened 2023-02-03
Signals And Slots Mechanism


drewr95:
If this is already implemented and I didn't find it in the ETL documentation then please forgive me.  

I think it would be nice and I'm interested in implementing a signals and slots mechanism utilizing ETL's delegates.  Would you be willing to have that as part of the ETL library?

I was thinking of utilizing ETL's list mechanism as well and basing it similarly to [nod's](https://github.com/fr00b0/nod) implementation of it.

Thanks

jwellbelove:
There isn't a signals & slots mechanism in the ETL at the moment. The closest thing to it at the moment is `etl::observer` which allows you to subscribe to a notification.
I will take a look at the implementation that you linked.

jwellbelove:
Added in 20.44.0

PR #1012, opened 2025-01-12
Implement Signal

drewr95:
Initial implementation of signal for https://github.com/ETLCPP/etl/issues/666

I implemented something very similar to this on other software and thought this would be useful for this project.  I would like to use this version over my other implementation.  

Some other things that could be added at some point is being able to pass in an aggregator or some kind of module that can act on all the signals (think sumnation, product, etc).

### DIFF
--- a/include/etl/signal.h
+++ b/include/etl/signal.h
@@ -106,7 +106,7 @@ namespace etl
     template <typename... TArgs>
     ETL_CONSTEXPR14 explicit signal(TArgs&&... args) ETL_NOEXCEPT
       : _slots{ETL_OR_STD::forward<TArgs>(args)...}
-      , _end{std::next(_slots, sizeof...(args))}
+      , _end{_slots + sizeof... (args)}
     {
     };
 
--- a/include/etl/signal.h
+++ b/include/etl/signal.h
@@ -39,8 +39,10 @@ SOFTWARE.
 #include "platform.h"
 
 #if ETL_USING_STL
+  #include <algorithm>
   #include <iterator>
 #else 
+  #include "algorithm.h"
   #include "iterator.h"
 #endif // ETL_USING_STL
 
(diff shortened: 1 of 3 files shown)


include/etl/signal.h
```diff
@@ -0,0 +1,250 @@
...
+  /// ability to store in ROM.
+  ///
+  ///\tparam T: Callback signature.
+  ///\tparam LENGTH: Maximum number of slots that can be connected to the signal.
+  ///\tparam TSlot: Function-object type or container type that can be invoked.
+  ///
+  ///\todo Support for return types other than void (aggregate etc.)
+  //***************************************************************************
+  template <typename T, size_t LENGTH, typename TSlot = etl::delegate<T>>
+  class signal final
```
jwellbelove, line 97: If this class is meant to support C++03 then you should use `ETL_FINAL`.

You can check basic compatibility from C++03 to C++20 by running the script `etl/test/run-syntax-checks.sh` on Linux. You need both GCC and clang installed.

drewr95: My primary goal was to be c++14 compliant since that is what I use, but I tried to be as lower compliant as possible.  I didn't know that script existed, thanks, I'll run it and see what I find :)

jwellbelove:
I'll pull what you currently have to a branch, so I can have a play with it.

drewr95:
> I'll pull what you currently have to a branch, so I can have a play with it.

Yeah that's fine, thank you for looking at this.  Sorry, work has gotten busy and so I've been putting this off.  Feel free to separate via 03 vs 11+ if you'd like or however you see fit.  I know some places have a *_legacy header.

drewr95:
> I know for sure that people will be asking how to use this across threads. Have you tried doing this yet? With the ETL messaging framework you can use thread safe queues to manage this. I'm not sure yet how it would be managed with signals.

@jwellbelove No we are still single process here though we are about to go multi-core at some point.  Would you mind keeping me in the loop?  I'd like to stay informed as that is not my expertise :)

PR #1014, opened 2025-01-15
Added missing rational arithmetic functions from <ratio>

rolandreichweinbmw:


include/etl/ratio.h
```diff
@@ -97,6 +100,116 @@ namespace etl
 
   /// An approximation of e.
   typedef ratio<326, 120> ratio_e;
+
+#if ETL_USING_CPP14
```
jwellbelove: These are available in C++11 in the STL.

rolandreichweinbmw: I.e. you are asking to provide a C++11 implementation?

In this case, we would need to replace usage of gcd and lcm since they are not implemented in ETL for C++11.

jwellbelove: _"they are not implemented in ETL for C++11"_
See:-
gdc.h lines 87-140
lcm.h lines 84-137

rolandreichweinbmw: You are right. :-)

The reason I initially limited to >=C++14 is slightly different: In gcd.h, etl::gcd is ETL_CONSTEXPR14 which means I can only use it as constexpr and therefore in compile time computation of gcd for ratio_subtract etc. in C++14. 

Or am I missing sth. here?

We would need the equivalent of etl::gcd and etl::lcm as constexpr implementation in C++11?

jwellbelove: Ah, I see.

jwellbelove: I think I can come up with a bit of template metaprogramming that will solve that for integral template parameters. Making the current ETL `gdc` and `lcm` implementations `constexpr` for C++11 forces them to be recursive, which I'm not keen on. I try to keep recursive functions out of the ETL due to the possibility of blowing the stack on resource limited platforms.

jwellbelove: If you add these templates into your internal namespace, then you can calculate GCD and LCM for C++11.

```
// Primary template for GCD calculation
template <typename T, T A, T B, bool = (B == 0)>
struct const_gcd;

// Specialisation for the case when B is not zero
template <typename T, T A, T B>
struct const_gcd<T, A, B, false>
{
  static constexpr T value = const_gcd<T, B, A % B>::value;
};

// Specialisation for the case when B is zero
template <typename T, T A, T B>
struct const_gcd<T, A, B, true>
{
  static constexpr T value = A;
};

// Primary template for LCM calculation
template <typename T, T A, T B>
struct ratio_lcm
{
private:

  static constexpr T product = ((A * B) < 0) ? -(A * B) : A * B;

public:

  static constexpr T value = product / ratio_gcd<T, A, B>::value;
};
```

Use them like this:-

```
template<class R1>
struct _ratio_reduce
{
private:
  static ETL_CONSTEXPR11 intmax_t gcd = etl::ratio_gcd<intmax_t, R1::num, R1::den>::value;
public:
  using value = ratio<R1::num / gcd, R1::den / gcd>;
};

```
On a style note, the code doesn't quite fit with the style I adopted for the rest of the ETL.

I name internal namespaces after the class that they support, to reduce name clashes with other 'non-user' code

```
namespace etl
{
  namespace private_ratio
  {
    template <typename R1>
    struct ratio_reduce
    {
    };
  }
}

```
The ETL uses this convention for template classes

`template <typename R1>`

I don't use leading '_' as this convention is normally reserved for compiler libraries.

rolandreichweinbmw: Thanks for the guide. :-)

Please have a look at the updated PR.

I think I addressed all your notes.

PR #1016, opened 2025-01-16
Allow compile time CRC calculation

Zob314:
Adds constexpr to CRC, allowing compile time computation of CRC values.

I also added constexpr specific tests, but I'm really not sure if every CRC type needs a constexpr test. I'm open to suggestions.

Tested in c++11, c++14, c++17, and c++20 with gcc

All the tests were added via Regex pattern matching, so they're all nearly identical.


jwellbelove:
The tests fail for C++03.

include/etl/frame_check_sequence.h
```diff
@@ -107,7 +107,7 @@ namespace etl
     //*************************************************************************
     /// Default constructor.
     //*************************************************************************
-    frame_check_sequence()
+    ETL_CONSTEXPR14 frame_check_sequence() : frame_check{}
```
jwellbelove: The CI fails for C++03 syntax check.
You will have to change `frame_check{}` to `frame_check()` for C++03 compatibility.

include/etl/private/crc_implementation.h
```diff
@@ -241,16 +241,21 @@ namespace etl
     struct crc_table<TAccumulator, Accumulator_Bits, Chunk_Bits, Chunk_Mask, Polynomial, Reflect, 4U>
     {
       //*************************************************************************
+#if !ETL_USING_CPP11 || defined(ETL_FORCE_NO_ADVANCED_CPP)
```
Zob314: Is `ETL_FORCE_NO_ADVANCED_CPP` supported? It's only in profile.h and some tests, and it seems to break a few things when defined.

I can remove the extra check if it's not supported.

jwellbelove: It looks like `ETL_FORCE_NO_ADVANCED_CPP` is a redundant macro left over from some early compile options. It's not even documented.
I will mark them for removal.

PR #1023, opened 2025-02-03
Alignment typed storage

rolandreichweinbmw:
I have a recurring use case of globally allocated class instances that I need to explicitly construct in a more defined order than implicit construction of global C++ objects can offer.

Therefore, I have a convenience wrapper etl::typed_storage for aligned_storage_as for instantiating an object explicitly at a defined time.

Previously, I tried with etl::optional<T>, which basically supports the same interface I need, but I can't use this one because it always references ~T. But I want the compiler to optimize away the unneeded dtors because the global objects won't get destroyed anyway until power-off in many embedded scenarios.

(You can still explicitly destroy() if needed.)

### DIFF
--- a/include/etl/alignment.h
+++ b/include/etl/alignment.h
@@ -405,7 +405,7 @@ namespace etl
     /// \returns the instance of T which has been constructed in the internal byte array.
     //***************************************************************************
     template<typename... Args>
-    reference emplace(Args&&... args)
+    reference create(Args&&... args)
     {
       ETL_ASSERT(!has_value(), ETL_ERROR(etl::typed_storage_error));
       valid = true;
--- a/include/etl/alignment.h
+++ b/include/etl/alignment.h
@@ -36,6 +36,7 @@ SOFTWARE.
 #include "static_assert.h"
 #include "error_handler.h"
 #include "exception.h"
+#include "utility.h"
 
 #include <stdint.h>
 
@@ -67,6 +68,19 @@ namespace etl
 
     alignment_error(string_type file_name_, numeric_type line_number_)
       : alignment_exception(ETL_ERROR_TEXT("alignment:error", ETL_ALIGNMENT_FILE_ID"A"), file_name_, line_number_)
+    {
+    }
+  };
+
+  //***************************************************************************
+  /// Typed storage exception.
+  //***************************************************************************
+  class typed_storage_error : public alignment_exception
+  {
+  public:
+
+    typed_storage_error(string_type file_name_, numeric_type line_number_)
+      : alignment_exception(ETL_ERROR_TEXT("typed_storage:error", ETL_ALIGNMENT_FILE_ID"B"), file_name_, line_number_)
     {
     }
   };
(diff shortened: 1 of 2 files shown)


include/etl/alignment.h
```diff
@@ -71,6 +72,19 @@ namespace etl
...
 
+  //***************************************************************************
+  /// Typed storage exception.
+  //***************************************************************************
+  class typed_storage_error : public alignment_exception
+  {
+  public:
+
+    typed_storage_error(string_type file_name_, numeric_type line_number_)
+      : alignment_exception(ETL_ERROR_TEXT("typed_storage:error", ETL_ALIGNMENT_FILE_ID"A"), file_name_, line_number_)
```
jwellbelove: This should be `ETL_ALIGNMENT_FILE_ID"B"` as it's the second `alignment_exception`

include/etl/alignment.h
```diff
@@ -333,6 +347,109 @@ namespace etl
...
+    /// asserts if has_value() is true before calling emplace().
+    ///
+    /// \returns the instance of T which has been constructed in the internal byte array.
+    //***************************************************************************
+    template<typename... Args>
+    reference emplace(Args&&... args)
+    {
+      ETL_ASSERT(!has_value(), ETL_ERROR(etl::typed_storage_error));
+      valid = true;
+      return *new (data.template get_address<char>()) value_type(etl::forward<Args>(args)...);
```
jwellbelove: I would use `::new` here to explicitly reference the global `new`.

include/etl/alignment.h
```diff
@@ -333,6 +347,109 @@ namespace etl
...
+    }
+
+    //***************************************************************************
+    /// Constructs the instance of T forwarding the given \p args to its constructor and
+    /// asserts if has_value() is true before calling emplace().
+    ///
+    /// \returns the instance of T which has been constructed in the internal byte array.
+    //***************************************************************************
+    template<typename... Args>
+    reference emplace(Args&&... args)
```
jwellbelove: For the other classes in the ETL that have a member function that constructs an object, I have used _create(Args&&... args)_.
I can see the reason for using `emplace`, but it would be inconsistent naming.
Also, _create_ is the common antonym of `destroy`.

jwellbelove:
It says my requested review changes are 'Pending'.

rolandreichweinbmw:
I have the feeling I'm missing sth. here. :-) Can you please point me to your change requests? I currently can't see any in

https://github.com/ETLCPP/etl/pull/1023

or

https://github.com/ETLCPP/etl/pull/1023/files

?

include/etl/alignment.h
rolandreichweinbmw: Adjusted accordingly.

include/etl/alignment.h
rolandreichweinbmw: Adjusted accordingly.

include/etl/alignment.h
rolandreichweinbmw: Are you sure? emplace() is already being used in more ETL classes than create() is:

```
$ grep -r 'create(Args' include/etl/ -m1 |wc -l
5
$ grep -r 'emplace(Args' include/etl/ -m1 |wc -l
13
$
```

rolandreichweinbmw:
In the last commit in the PR, I replaced emplace() with create(). So depending how you decide regarding this name question above, you can take this one or leave this last commit out.

rolandreichweinbmw:
The feature is already available in the feature extended version of ETL at https://github.com/rolandreichweinbmw/etlplus

PR #1024, opened 2025-02-04
etl::span: Add advance(), copy(), reinterpret_as()

rolandreichweinbmw:
This commit adds to etl::span:
- advance()
- copy()
- reinterpret_as()

including tests.


rolandreichweinbmw:
Right. I will add this.

include/etl/span.h
```diff
@@ -746,6 +775,30 @@ namespace etl
            etl::equal(lhs.begin(), lhs.end(), rhs.begin(), rhs.end());
   }
 
+  //*************************************************************************
+  /// Copy complete element data from one span to another. If the destination
+  /// span is bigger than the source span, only the initial part of
+  /// destination span is overwritten.
+  ///\param src Source
+  ///\param dst Destination
+  ///\return true, iff copy was successful. Reasons for unsuccessful operation:
```
jwellbelove: iff

include/etl/span.h
```diff
@@ -746,6 +775,30 @@ namespace etl
...
+  /// 1. Source span is empty.
```
jwellbelove: Are you sure that copying an empty span is really an _unsuccessful_ operation?

include/etl/span.h
```diff
@@ -746,6 +775,30 @@ namespace etl
...
+  /// 2. They both point to the same data.
```
jwellbelove: Isn't this just a null-op, as opposed to a failure?

include/etl/span.h
rolandreichweinbmw: iff == if and only if

https://en.wikipedia.org/wiki/If_and_only_if

But I agree that it regularly leads to misunderstandings :-)

include/etl/span.h
rolandreichweinbmw: Similar to the null-op in copying to same location, you could consider it useful to be notified that no actual copying is taking place.

But if you prefer, we can make both cases valid, i.e. returning true.

include/etl/span.h
rolandreichweinbmw: Adjusted accordingly.

include/etl/span.h
rolandreichweinbmw: Update: Adjusted accordingly.

include/etl/span.h
rolandreichweinbmw: I changed it.

include/etl/span.h
```diff
@@ -781,17 +820,20 @@ namespace etl
...
-  /// 2. They both point to the same data.
-  /// 3. The destination span is shorter than the source span.
+  ///\return true, if copy was successful (including empty source span, or
+  ///        spans pointing to the same address)
+  ///\return false, if the destination span is shorter than the source span.
   //*************************************************************************
   template <typename T1, size_t N1, typename T2, size_t N2>
   typename etl::enable_if<etl::is_same<typename etl::remove_cv<T1>::type, typename etl::remove_cv<T2>::type>::value &&
                           !etl::is_const<T2>::value, bool>::type
     copy(const etl::span<T1, N1>& src, const etl::span<T2, N2>& dst)
```
jwellbelove, line 966: This is a class member function, but it copies between two other spans.
This would be better implemented as a separate standalone function, or maybe as a member `copy_to(dst)`

include/etl/span.h
```diff
@@ -710,15 +745,19 @@ namespace etl
...
 
     //*************************************************************************
     /// Reinterpret the span as a span with different element type.
     //*************************************************************************
     template<typename TNew>
-    ETL_NODISCARD ETL_CONSTEXPR etl::span<TNew, etl::dynamic_extent> reinterpret_as() const ETL_NOEXCEPT
+    ETL_NODISCARD ETL_CONSTEXPR14 etl::span<TNew, etl::dynamic_extent> reinterpret_as() const
     {
+#if ETL_USING_CPP11
+      ETL_ASSERT(etl::is_aligned<alignof(TNew)>(pbegin), ETL_ERROR(span_alignment_exception));
```
jwellbelove: You can use `etl::alignment_of` instead of `alignof` which works for C++03.

include/etl/span.h
```diff
@@ -710,15 +745,19 @@ namespace etl
...
     }
 
     //*************************************************************************
     /// Reinterpret the span as a span with different element type.
     //*************************************************************************
     template<typename TNew>
-    ETL_NODISCARD ETL_CONSTEXPR etl::span<TNew, etl::dynamic_extent> reinterpret_as() const ETL_NOEXCEPT
+    ETL_NODISCARD ETL_CONSTEXPR14 etl::span<TNew, etl::dynamic_extent> reinterpret_as() const
     {
+#if ETL_USING_CPP11
```
jwellbelove, line 758: This does not check that the value of `pend - pbegin` is sized to a whole number of `TNew` elements.

If `pend - pbegin` is 3 bytes and `TNew` is an `int32_t` then the size of the new span is 0.

If `pend - pbegin` is 6 bytes and `TNew` is an `int32_t` then the size of the new span is 1, with 2 bytes before `pend`.

An advance of the `int32_t` `span` `pbegin` pointer would never equal `pend`, meaning loops of the return `span` would never terminate.

include/etl/span.h
rolandreichweinbmw: Are you sure?

As far as I can see, this free function etl::copy(), actually an overload, is outside the class, similar to the directly above etl::equal().

What is strange here is the line numbering. When I merge it on top of current master, it ends up in l.968, not l.830.

include/etl/span.h
rolandreichweinbmw: Thanks for the hint! Updating it accordingly.

rolandreichweinbmw:
Regarding your remark about odd-numbered spans where you get rounding errors when reinterpreting: This is intended, and considered a feature. Consider the case of having a buffer from some protocol data, and only a part of it is the respective span, i.e. value array. Then the size is automatically rounded down to the next correct size, and we can only correctly assume the initial whole elements to be valid, and discard the rest:

* "If pend - pbegin is 3 bytes and TNew is an int32_t then the size of the new span is 0." - Yes, as expected.
* "If pend - pbegin is 6 bytes and TNew is an int32_t then the size of the new span is 1, with 2 bytes before pend." - Yes, as expected
* "An advance of the int32_t span pbegin pointer would never equal pend, meaning loops of the return span would never terminate." - No, because the expected new size 1 leads to pbegin and pend to be set correctly, with pend discarding the bytes you mentioned in the first 2 cases.

I added further unit tests for reinterpret_as() to show what I mean.

jwellbelove:
Yes, I realised that after I wrote the comment. I deleted the comment.

rolandreichweinbmw:
After rebasing onto latest ETL with new unaligned_type, I encountered failing unit tests for reinterpret_as() on Windows.

It was caused by the new multiple inheritance of unaligned_type because sizeof() now results in additional bytes counted for the second base class unaligned_copy which contributed an additional byte. (Even though empty class - but looks like 1 byte is reserved for that.)

My solution in the branch is the change to single inheritance, moving unaligned_copy to a class member typedef. Including tests for that.

include/etl/span.h
jwellbelove: Sorry, I see that it is already a standalone function. I got confused looking at diff view late last night. It made the copy function look like a member of the class.
I looked at the full file view this morning.

PR #1026, opened 2025-02-07
Intrusive forward list add remove by pointer

rolandreichweinbmw:
Besides remove() by reference, I provide this remove() by pointer (overload).

Scenario is as follows:

Current remove() by reference uses the comparison operator, comparing actually different instances for equality. In my use case, I have objects in the intrusive_forward_list which don't have a comparison operator defined but I have a pointer to them. This way, I can intentionally remove the actual object from the list pointed to by the pointer.

### DIFF
--- a/include/etl/intrusive_forward_list.h
+++ b/include/etl/intrusive_forward_list.h
@@ -992,6 +992,28 @@ namespace etl
     }
 
     //*************************************************************************
+    // Removes the element specified by pointer.
+    //*************************************************************************
+    void remove(const_pointer element)
+    {
+      iterator i_item = begin();
+      iterator i_last_item = before_begin();
+
+      while (i_item != end())
+      {
+        if (&i_item == element)
+        {
+          i_item = erase_after(i_last_item);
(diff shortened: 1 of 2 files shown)


jwellbelove:
Your unit test `test_remove_by_pointer` does not remove by pointer.

`data0.remove(*element);` is removing by _reference_.

rolandreichweinbmw:
Thanks for finding and the correct fix. I can confirm it works now also without the "*".

jwellbelove:
I've also copied the change to `etl::intrusive_list`.

jwellbelove:
Actually, as I was editing the intrusive lists, I realised that that there was already an `node_type* erase(node_type& node)` that basically does what your `remove(const_pointer element)` does.

It's about 8 years since I wrote the intrusive lists!

jwellbelove:
I will add a `node_type* erase(const node_type* node)` and maybe change the `node_type* erase(node_type& node)` to `node_type* erase(const node_type& node)`

Issue #1032, opened 2025-02-22
Tasking Compiler Failure For parameter_pack.h


drewr95:
I have the compiler flag set to treat warnings as errors in the Tasking compiler and am getting the following error:

> cptc E0001: ["\dependencies\thirdparty\etl\include\etl/private/../parameter_pack.h" 150] last line of file ends without a newline

jwellbelove:
Fixed 20.40.0

PR #1033, opened 2025-02-22
Add newline to end of parameter_pack.h

drewr95:
Fixes: https://github.com/ETLCPP/etl/issues/1032

This was causing a Tasking compiler error when the treat-warnings-as-errors flag is enabled.


jwellbelove:
Added this in 20.40.0

drewr95:
Oh you're right, sorry about that.  I forgot I had a later version pegged, I usually CMakeFetch on the master branch.  Thanks again

PR #1036, opened 2025-02-25
Add contains() and contains_node() to etl::intrusive_forward_list

rolandreichweinbmw:


jwellbelove:
For completeness, there ought to be a version for `etl::intrusive_list`.

rolandreichweinbmw:
Agree. Just updated accordingly.

PR #1044, opened 2025-03-05
Add traits to type_list

rolandreichweinbmw:
This change adds several traits to type_list.h:

* type_list_contains
* type_list_index_of
* type_list_max_sizeof_type (long name to prevent confusion with type_list_size)
* type_list_max_alignment

Further: test_type_list.cpp including tests for already existing traits.

### DIFF
--- a/include/etl/type_list.h
+++ b/include/etl/type_list.h
@@ -30,7 +30,10 @@ SOFTWARE.
 #define ETL_TYPE_LIST_INCLUDED
 
 #include "platform.h"
+
+#include "algorithm.h"
 #include "nth_type.h"
+#include "integral_limits.h"
 #include "static_assert.h"
 #include "type_traits.h"
 #include "utility.h"
@@ -38,6 +41,9 @@ SOFTWARE.
 #if ETL_USING_CPP11
 namespace etl
 {
+
+  static ETL_CONSTEXPR size_t type_list_npos = etl::integral_limits<size_t>::max;
+
   //***************************************************************************
   /// Type list forward declaration.
   //***************************************************************************
@@ -61,6 +67,18 @@ namespace etl
     type_list& operator =(const type_list&) ETL_DELETE;
   };
 
+  namespace private_type_list
+  {
+
+    // helper to solve the issue that recursed-rest can't be put directly in type_list::tail definition
+    template <typename... TTypes>
+    struct recursion_helper
+    {
+      using type = type_list<TTypes...>;
+    };
+
+  }
+
   //***************************************************************************
   /// Recursive type list implementation for multiple types.
   //***************************************************************************
@@ -68,6 +86,7 @@ namespace etl
   struct type_list<THead, TTail...> : type_list<TTail...>
   {
     using type = THead;
+    using tail = typename private_type_list::recursion_helper<TTail...>::type;
 
     static constexpr size_t size = sizeof...(TTail) + 1U;
 
@@ -87,6 +106,7 @@ namespace etl
   struct type_list<THead> : type_list<>
   {
     using type = THead;
+    using tail = typename private_type_list::recursion_helper<>::type;
 
     static constexpr size_t size = 1U;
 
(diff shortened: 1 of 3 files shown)


include/etl/type_list.h
```diff
@@ -180,6 +197,56 @@ namespace etl
   {
     using type = typename type_list_cat<etl::type_list<TTypes1..., TTypes2...>, TTail...>::type;
   };
+
+  template<typename TL, typename T>
```
jwellbelove, line 201: This can be replaced with a slightly simpler version.

```
template <typename TypeList, typename T>
struct type_list_contains
  : public etl::integral_constant<bool, etl::is_one_of<T, typename TypeList::type, typename TypeList::tail>::value>
{
};
```

rolandreichweinbmw: The simplification via etl::is_one_of is only valid if etl::is_one_of would do the recursion step into TypeList::tail (with another type and tail inside). But as far as I can see, it doesn't.

Unit testing confirms my point.

jwellbelove:
Removed the review comment.
I thought I'd had it compiling OK with `etl::is_one_of`, but I hadn't.

include/etl/type_list.h
```diff
@@ -180,6 +197,56 @@ namespace etl
...
+    : public etl::integral_constant<size_t, etl::is_same<typename TL::type, T>::value ? 0 : (type_list_index_of<typename TL::tail, T>::value + 1)>
+  {
+  };
+
+  template<typename T>
+  struct type_list_index_of<type_list<>, T>
+    : public etl::integral_constant<size_t, 0>
+  {
+  };
+
```
jwellbelove, line 234: A corresponding 'type from index' class should be implemented.
```
template<typename TypeList, size_t Index>
struct type_list_type_at
```
i.e.
```
using TypeList = etl::type_list<char, short, int>;
using second_type = etl::type_list_type_at<TypeList, 1>::type; // short
```

rolandreichweinbmw: Isn't that available via the already implemented etl::nth_type<>?

include/etl/type_list.h
jwellbelove: I would prefer `TypeList` over `TL`

include/etl/type_list.h
```diff
@@ -180,6 +197,56 @@ namespace etl
...
+  template<>
+  struct type_list_max_alignment<type_list<>>
+    : public etl::integral_constant<size_t, 1>
+  {
+  };
+
 }
```
jwellbelove, line 274: For C++17, add `_v` definitions for types that return a value
i.e.
```
#if ETL_USING_CPP17
  template<typename TL, typename T>
  inline constexpr bool type_list_contains_v = etl::type_list_contains<TL, T>::value;
#endif
```

include/etl/type_list.h
```diff
@@ -180,6 +197,56 @@ namespace etl
...
+  {
+  };
+
+  template<typename TL, typename T>
+  struct type_list_index_of
+    : public etl::integral_constant<size_t, etl::is_same<typename TL::type, T>::value ? 0 : (type_list_index_of<typename TL::tail, T>::value + 1)>
+  {
+  };
+
+  template<typename T>
```
jwellbelove, line 229: Doesn't this give the same index for the first type in a non-empty `type_list`, and an empty `type_list`?

```
using NonEmpty = etl::type_list<char, int>;
using Empty    = etl::type_list<>;

size_t nonEmptyIndex = etl::type_list_index_of<NonEmpty, char>::value;

size_t emptyIndex = etl::type_list_index_of<Empty, char>::value;
```

`nonEmptyIndex == emptyIndex`

Maybe add `static constexpr size_t npos = etl::integral_limits<size_t>::max;` to `etl::type_list<>` and return `etl::type_list<>::npos` for `emptyIndex`.

For the furture, I'm thinking that an ETL generic `etl::npos` may be useful

include/etl/type_list.h
rolandreichweinbmw: Adjusted accordingly.

include/etl/type_list.h
rolandreichweinbmw: Adjusted accordingly.

include/etl/type_list.h
rolandreichweinbmw: Adjusted accordingly.

The "npos" is implemented as etl::type_list_npos. Reasoning:
* etl::type_list<>::npos causes linker errors on application level / unit tests since the symbol needs to be defined in a single compile unit and in ETL we only work header-only. This is different from plain global symbols, or in an abstract template.
* etl::npos causes compile warnings and therefore errors because etl::basic_string_view::npos shadows etl::npos

include/etl/type_list.h
jwellbelove: Well, yes, but that's a much older class, originally designed for a narrower use case.
I think a class named consistently with other `type_list_x` classes could be less confusing.
There is another `etl::type_list` style class called `etl::parameter_pack` which will eventually be deprecated, as `etl::type_list` covers everything it does, and more.

PR #1046, opened 2025-03-06
Support etl::inplace_function

jiangyilism:
Provide similar api to std::function.

Current use case is mainly to hold a r-value lambda object, which etl::delegate is unsuitable since etl::delegate does not extend the lifetime of r-value lambda object.

Store a callable inside the inplace_function object's member field in a type-erasure way.
The member field, i.e. storage or buffer, has a compile-time fixed size.
The size is specified either by the macro ETL_INPLACE_FUNCTION_DEFAULT_CAPACITY or a non-type template parameter.

The implementation is inspired by:
1. SG14 inplace_function
2. folly::Function
3. function2

### DIFF
--- a/include/etl/private/inplace_function_cpp14.h
+++ b/include/etl/private/inplace_function_cpp14.h
@@ -1,0 +1,234 @@
+///\file
+
+/******************************************************************************
+The MIT License(MIT)
+
+Embedded Template Library.
+https://github.com/ETLCPP/etl
+https://www.etlcpp.com
+
+Copyright(c) 2025 BMW AG
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files(the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and / or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions :
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
+******************************************************************************/
+
+#ifndef ETL_INPLACE_FUNCTION_CPP14_INCLUDED
+#define ETL_INPLACE_FUNCTION_CPP14_INCLUDED
+
+#include "../platform.h"
+#include "../type_traits.h"
+#include "../utility.h"
+
+#ifndef ETL_INPLACE_FUNCTION_DEFAULT_CAPACITY
+  #define ETL_INPLACE_FUNCTION_DEFAULT_CAPACITY 32
+#endif
+namespace etl
+{
+  namespace private_inplace_function
+  {
+    template <typename T>
+    struct wrapper
+    {
+      using type = T;
+    };
+    template <typename R, typename... Args>
+    struct vtable_t ETL_FINAL
+    {
+      using storage_ptr_t = void*;
+      using invoke_func_t = R (*)(storage_ptr_t, Args&&...);
+      using copy_or_move_func_t = void (*)(storage_ptr_t, storage_ptr_t);
+      using dtor_func_t = void (*)(storage_ptr_t);
+
+      const invoke_func_t       invoke_func{[](storage_ptr_t, Args&&...) -> R
+                                      { return R(); }};
+      const copy_or_move_func_t copy_func{[](storage_ptr_t, storage_ptr_t) -> void {}};
+      const copy_or_move_func_t move_func{[](storage_ptr_t, storage_ptr_t) -> void {}};
+      const dtor_func_t         dtor_func{[](storage_ptr_t) -> void {}};
+
+      ~vtable_t() = default;
+
+      ETL_CONSTEXPR vtable_t() ETL_NOEXCEPT = default;
+
+      template <typename T>
+      explicit ETL_CONSTEXPR vtable_t(wrapper<T>) ETL_NOEXCEPT
+        : invoke_func{[](storage_ptr_t storage_ptr, Args&&... args) -> R
+                      { return (*static_cast<T*>(storage_ptr))(
+                          etl::forward<Args&&>(args)...); }},
+          copy_func{[](storage_ptr_t dst_ptr, storage_ptr_t src_ptr) -> void
+                    { ::new (dst_ptr) T{*static_cast<T*>(src_ptr)}; }},
+          move_func{[](storage_ptr_t dst_ptr, storage_ptr_t src_ptr) -> void
+                    {
+                      ::new (dst_ptr) T{etl::move(*static_cast<T*>(src_ptr))};
+                      static_cast<T*>(src_ptr)->~T();
+                    }},
+          dtor_func{[](storage_ptr_t src_ptr) -> void
+                    { static_cast<T*>(src_ptr)->~T(); }}
+      {
+      }
+
+    private:
+      vtable_t(const vtable_t&) ETL_DELETE;
+      vtable_t(vtable_t&&) ETL_DELETE;
+
+      vtable_t& operator=(const vtable_t&) ETL_DELETE;
+      vtable_t& operator=(vtable_t&&) ETL_DELETE;
+    };
+
+    template <class R, class... Args>
+    ETL_INLINE_VAR ETL_CONSTEXPR17
+        vtable_t<R, Args...>
+        default_vtable{};
+  }  // namespace private_inplace_function
+
+  template <
+    typename Signature,
+    size_t Capacity = ETL_INPLACE_FUNCTION_DEFAULT_CAPACITY>
+  class inplace_function;
+
+  namespace private_inplace_function
+  {
+    template <typename>
+    struct is_inplace_function : etl::false_type
+    {
+    };
+    template <typename Sig, size_t Cap>
+    struct is_inplace_function<inplace_function<Sig, Cap>> : etl::true_type
+    {
+    };
+  }  // namespace private_inplace_function
+
+  template <
+    typename R,
+    typename... Args,
+    size_t Capacity>
+  class inplace_function<R(Args...), Capacity> ETL_FINAL
+  {
+    using vtable_t = private_inplace_function::vtable_t<R, Args...>;
+    using vtable_ptr_t = const vtable_t*;
+
+    template <typename, size_t>
+    friend class inplace_function;
+
+  public:
+    using capacity = etl::integral_constant<size_t, Capacity>;
+
+    inplace_function() ETL_NOEXCEPT = default;
+
+    template <
+      typename T,
+      typename C = etl::decay_t<T>,
+      typename = etl::enable_if_t<
+        !private_inplace_function::is_inplace_function<C>::value && etl::is_invocable_r<R, C&, Args...>::value>>
+    inplace_function(T&& closure)
+    {
+      ETL_STATIC_ASSERT(etl::is_copy_constructible<C>::value,
+                        "cannot be constructed from non-copyable types");
+
+      ETL_STATIC_ASSERT(sizeof(C) <= Capacity,
+                        "internal storage too small");
+
+      static const vtable_t vt{private_inplace_function::wrapper<C>{}};
+
+      vtable_ptr = &vt;
+      ::new (storage) C{etl::forward<T>(closure)};
+    }
+
+    template <size_t Cap>
+    inplace_function(const inplace_function<R(Args...), Cap>& other)
+      : inplace_function(other.vtable_ptr, other.vtable_ptr->copy_func, other.storage)
+    {
+      ETL_STATIC_ASSERT(Capacity >= Cap,
+                        "internal storage too small");
+    }
+
+    template <size_t Cap>
+    inplace_function(inplace_function<R(Args...), Cap>&& other) ETL_NOEXCEPT
+      : inplace_function(other.vtable_ptr, other.vtable_ptr->move_func, other.storage)
+    {
+      ETL_STATIC_ASSERT(Capacity >= Cap,
+                        "internal storage too small");
+    }
+
+    inplace_function(const inplace_function& other)
+      : vtable_ptr{other.vtable_ptr}
+    {
+      vtable_ptr->copy_func(
+        storage,
+        other.storage);
+    }
+
+    inplace_function(inplace_function&& other) ETL_NOEXCEPT
+      : vtable_ptr{other.vtable_ptr}
+    {
+      vtable_ptr->move_func(
+        storage,
+        other.storage);
+    }
+
+    inplace_function& operator=(etl::nullptr_t) ETL_NOEXCEPT
+    {
+      vtable_ptr->dtor_func(&storage);
+      vtable_ptr = ETL_NULLPTR;
+      return *this;
+    }
+
+    inplace_function& operator=(inplace_function other) ETL_NOEXCEPT
+    {
+      vtable_ptr->dtor_func(storage);
+      vtable_ptr = other.vtable_ptr;
+      vtable_ptr->move_func(
+        storage,
+        other.storage);
+      return *this;
+    }
+
+    ~inplace_function()
+    {
+      vtable_ptr->dtor_func(storage);
+    }
+
+    R operator()(Args... args) const
+    {
+      return vtable_ptr->invoke_func(
+        storage,
+        etl::forward<Args>(args)...);
+    }
+
+    explicit ETL_CONSTEXPR operator bool() const ETL_NOEXCEPT
+    {
+      return vtable_ptr != &private_inplace_function::default_vtable<R, Args...>;
+    }
+
+  private:
+    vtable_ptr_t vtable_ptr{&private_inplace_function::default_vtable<R, Args...>};
+    mutable char storage[Capacity];
+
+    inplace_function(
+      vtable_ptr_t                           vtable,
+      typename vtable_t::copy_or_move_func_t copy_or_move_func,
+      typename vtable_t::storage_ptr_t       storage_ptr)
+      : vtable_ptr{vtable}
+    {
+      copy_or_move_func(storage, storage_ptr);
+    }
+  };
+
+}  // namespace etl
+
+#endif
--- a/test/test_inplace_function.cpp
+++ b/test/test_inplace_function.cpp
@@ -1,0 +1,652 @@
+/******************************************************************************
+The MIT License(MIT)
+
+Embedded Template Library.
+https://github.com/ETLCPP/etl
+https://www.etlcpp.com
+
+Copyright(c) 2025 BMW AG
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files(the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and / or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions :
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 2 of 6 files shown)


include/etl/private/inplace_function_cpp14.h
```diff
@@ -0,0 +1,234 @@
...
+    template <typename R, typename... Args>
+    struct vtable_t ETL_FINAL
+    {
+      using storage_ptr_t = void*;
+      using invoke_func_t = R (*)(storage_ptr_t, Args&&...);
+      using copy_or_move_func_t = void (*)(storage_ptr_t, storage_ptr_t);
+      using dtor_func_t = void (*)(storage_ptr_t);
+
+      const invoke_func_t       invoke_func{[](storage_ptr_t, Args&&...) -> R
+                                      { return R(); }};
```
jiangyilism: `std::function` throws `std::bad_function_call` if operator() is called on an empty object while `etl::inplace_function` safely returns a default-constructed object in this case.
That means type `typename R` is required to be default-constructible. 
`void` type is fine here ( https://timsong-cpp.github.io/cppwp/std14/expr.type.conv )
But this will not compile since `std::is_default_constructible_v<Type_a>` is false:
```
  struct Type_a
  {
    Type_a(int32_t val) : val_(val) {}

    int32_t val_;
  };

etl::inplace_function<Type_a()> func0;
```

This deviation from `std::function` is by designed and also tested in `test_inplace_function.cpp`

rolandreichweinbmw (approved):

include/etl/private/inplace_function_cpp14.h
jwellbelove: In `etl::delegate` an ETL error is raised.
`ETL_ASSERT(is_valid(), ETL_ERROR(delegate_uninitialised));`

Also in `etl::delegate` there are `call_if` and `call_or` member functions that handle uninitialised objects.

include/etl/private/inplace_function_cpp14.h
```diff
@@ -0,0 +1,234 @@
+///\file
+
+/******************************************************************************
+The MIT License(MIT)
+
+Embedded Template Library.
+https://github.com/ETLCPP/etl
+https://www.etlcpp.com
+
+Copyright(c) 2025 BMW AG
```
jwellbelove: How close a copy is this of the implementations that 'inspired' you?
There may be copyright implications.

include/etl/private/inplace_function_cpp14.h
```diff
@@ -0,0 +1,234 @@
...
+  public:
+    using capacity = etl::integral_constant<size_t, Capacity>;
+
+    inplace_function() ETL_NOEXCEPT = default;
+
+    template <
+      typename T,
+      typename C = etl::decay_t<T>,
+      typename = etl::enable_if_t<
+        !private_inplace_function::is_inplace_function<C>::value && etl::is_invocable_r<R, C&, Args...>::value>>
```
jwellbelove: This class has limited usefulness as is restricted to STL and C++14 and above.
Many of the users of the ETL do not/cannot use the STL.

jiangyilism: `etl::is_invocable_r` depends on `ETL_USING_CPP11` but not STL.
I updated to make `etl::inplace_function` compatible to c++11.

include/etl/private/inplace_function_cpp14.h
jiangyilism: The closest open source implementation is SG14 inplace_function.
`etl::inplace_function` compiled object memory layout will be almost identical (except alignment).
The main differences:
1. etl is compatible with c++11 while others, including SG14, typically requires c++14
2. etl one has no member/non-member function `swap()`, `operator==(nullptr_t)`, `operator!=(nullptr_t)` .
3. etl does not utilize aligned storage, which will be deprecated in c++23.
4. etl renames most of the class and variable names, except the name "inplace_function", which can be renamed too.
5. etl simplifies some template meta programming with etl type_traits header
6. move almost all stuffs in private namespace into private sections of classes.

include/etl/private/inplace_function_cpp14.h
jiangyilism: Updated to make it raise `etl::bad_inplace_function_call` when an empty `etl::inplace_function` is invoked with `operator()` .    (Similar to `std::bad_function_call`)
`std::function` does not support `call_if()` and `call_or()`. Do you suggest adding these 2 member func?

denravonska:
The missing R-value support blocks us from upgrading ETL. This will be great :)

include/etl/private/inplace_function_cpp11.h
```diff
@@ -0,0 +1,245 @@
...
+      const dtor_func_t         dtor_func{[](char*) -> void {}};
+
+      ~vtable_t() = default;
+      vtable_t() = default;
+
+      template <typename T>
+      explicit ETL_CONSTEXPR vtable_t(etl::type_identity<T>) ETL_NOEXCEPT
+        : invoke_func{[](char* pobj, Args&&... args) -> R
+                      { return (*reinterpret_cast<T*>(pobj))(
+                          etl::forward<Args&&>(args)...); }},
```
BartolomeyKant: I'm confused about this construction. Why `etl::forward<Args&&>` instead of `etl::forward<Args>` is used here?

include/etl/private/inplace_function_cpp11.h
```diff
@@ -0,0 +1,245 @@
...
+    {
+      vtable_ptr->dtor_func(obj);
+    }
+
+    R operator()(Args... args) const
+    {
+      ETL_ASSERT(vtable_ptr->invoke_func, ETL_ERROR(bad_inplace_function_call));
+      return vtable_ptr->invoke_func(
+        obj,
+        etl::forward<Args>(args)...);
```
BartolomeyKant: Does conversion from `Args...` to `Args&&...` is working here?

For example I want to pass some `struct Foo` by value into inplace_function. Would it be copied three times?

First in call to this `operator()`, second in `invoke_func` call, and third in call to actual functor inside the `invoke_func` lambda?

I'm not an expert I'm just wondering how this should work.

BartolomeyKant: Maybe `std::move(args)` should be used here, and then `std::forward<Args>(args)` in `invoke_func` lambda?

include/etl/private/inplace_function_cpp14.h
denravonska: @jiangyilism I suggest adding them. It avoids having to check if the object is assigned anything, and helps moving from `delegate`.

denravonska:
Out of curiosity, what's using all the space? I had to bump the size to 48 bytes for an `inplace_function<void(void *)>` on Cortex-M33.

rolandreichweinbmw:
John, can you please enable the CI workflows for this PR?

denravonska:
> > Out of curiosity, what's using all the space? I had to bump the size to 48 bytes for an `inplace_function<void(void *)>` on Cortex-M33.
> 
> Depends on what you save inside. If it is just a "void f1(void*){}" it maybe fits?
> 
> But if it's a lambda capturing a number of arguments by value inside, it may be much more.

Actually, I accidentally compiled for host so it's 48 bytes on a 64 bit system which makes a bit more sense.

BartolomeyKant:
> 64 bit system

What if ETL_INPLACE_FUNCTION_DEFAULT_CAPACITY would be calculated based on pointer size, like `sizeof(void*)*4`?

include/etl/private/inplace_function_cpp11.h
```diff
@@ -0,0 +1,245 @@
...
+#define ETL_INPLACE_FUNCTION_CPP11_INCLUDED
+
+#include "../platform.h"
+#include "../error_handler.h"
+#include "../exception.h"
+#include "../type_traits.h"
+#include "../utility.h"
+
+#ifndef ETL_INPLACE_FUNCTION_DEFAULT_CAPACITY
+  #define ETL_INPLACE_FUNCTION_DEFAULT_CAPACITY 32
```
denravonska: Can this be based on the target architecture? When building for 64 bit the bare minimum (`void()`) seems to be 40 bytes.

rolandreichweinbmw:
> > 64 bit system
> 
> What if ETL_INPLACE_FUNCTION_DEFAULT_CAPACITY would be calculated based on pointer size, like `sizeof(void*)*4`?

Similar to etl::delegate and etl::function, inplace_function has one single global size.

This one needs to store all possible closures in a whole project that *uses* ETL for a inplace_function template instance type. I.e. it depends.

E.g. for a simple etl::inplace_function<void(void *)> as mentioned by Denravonska, you might save different closure implementations, e.g.:
- void f1(void*){}
- [=]<void(void *)>(){a(b);c(d);e(f);g(h);}

I.e. needs to be adjusted if necessary for every project, to not waste memory.

I'm not sure if you can find a way to automatically calculate a global maximum size over all user locations. Finally, you can only find at link time (or after compiling all comile units) if you have captured all necessary source locations.

rolandreichweinbmw:
Hi @BartolomeyKant and @denravonska , I think now I got your point. :-) And I agree that different defaults for e.g. 32 and 64 bit platforms can be provided.

Even though an individual ETL_INPLACE_FUNCTION_DEFAULT_CAPACITY will typically be part of a project's etl_profile.h anyway.

The original default of 32 covered our initial use cases in a 32 bit project.

Luckily, the compiler will tell you at build time if it doesn't fit, as shown in @denravonska 's example.

denravonska:
> Hi @BartolomeyKant and @denravonska , I think now I got your point. :-) And I agree that different defaults for e.g. 32 and 64 bit platforms can be provided.
> 
> Even though an individual ETL_INPLACE_FUNCTION_DEFAULT_CAPACITY will typically be part of a project's etl_profile.h anyway.
> 
> The original default of 32 covered our initial use cases in a 32 bit project.
> 
> Luckily, the compiler will tell you at build time if it doesn't fit, as shown in @denravonska 's example.

I thought about this a bit more. My original concern was that the default size was valid for 32 bit but failed to compile on 64 bit, so maybe the default should be the smallest size required to fit the simplest inplace function given the target architecture.

_However_ this only applies to the default function. As soon as you do anything fancier you need to handle the size differences in the application anyway.

include/etl/private/inplace_function_cpp14.h
denravonska: `clear()` would also be nice to have for `delegate` compatibility.

jiangyilism:
Thanks for all the feedback.
@jwellbelove As you said in the slack channel discussion, your have a plan for a new fixed-storage function/delegate . Do you think I should drop this PR?

denravonska:
> Thanks for all the feedback. @jwellbelove As you said in the slack channel discussion, your have a plan for a new fixed-storage function/delegate . Do you think I should drop this PR?

He also said

> I may decided to go with the etl::inplace_function instead.

I have no major quarrel with the implementation. The only request I have so far is a default size that works for both 32- and 64 bit systems.

jwellbelove:
@jiangyilism No, keep it active for the time being. I'm in Greece at the moment, so I'll look at this functionality again when I get back next week.

include/etl/private/inplace_function_cpp11.h
jiangyilism: Thanks. Fixed to `etl::forward<Args>` .

BartolomeyKant (approved):

include/etl/private/inplace_function_cpp11.h
jiangyilism: It is copied twice under gcc/clang -O0 , -O1 and -O2 . Using `etl::move`/`std::move` still copies twice.
A test case `NoRedundentCopy` is added to verify that during `inplace_function::operator()`, if the contained callable signature is:

1. call-by-value: then at most 2 copyings are made in `inplace_function::operator()`.
2. call-by-reference: then no copying is made.

BartolomeyKant: > It is copied twice

What if `operator()` also would use `Arg&&...` as parameter?

BartolomeyKant: This example gives me one copy and one move https://godbolt.org/z/3YPMccbnz

jiangyilism: Yes. And if the move constructor is not explicitly defined then copy constructor is called instead. `NoRedundentCopy` only checks the number of copy is less than 3 . It roughly checks that inplace_function is not copying too many times than expected. It does not guarantee the exact number of copying.

With gcc:
1. copy ctor not explicitly defined : copy twice
2. copy ctor =default : copy once
3. copy ctor explicitly defined : copy once

test/test_inplace_function.cpp
```diff
@@ -0,0 +1,683 @@
...
+      return *this;
+    }
+  };
+}  // namespace
+
+namespace
+{
+  SUITE(test_inplace_function)
+  {
+    TEST(NoRedundentCopy)
```
rolandreichweinbmw: NoRedundantCopy

denravonska (changes requested):
It would be a quality of life change with `.call_if()`. Consider

```Cpp
for(auto cb : callbacks) {
   if(cb) {
      cb();
   }
}
```

vs
```Cpp
for(auto cb : callbacks) {
   cb.call_if();
}
```

That, `.call_or` and a `.clear` would make it a lot more compatible with `delegate`.

jiangyilism:
> call_if

std::function does not support call_if, call_or, clear .
std::function and etl::inplace_function clear its content with `operator=(nullptr_t) noexcept`, which is now tested in `HasValue` test case.

If we want all etl::delegate api then it might be better to make inplace_function and delegate into a same class.
We need @jwellbelove 's feedback to continue with one of the directions:
1. extend inplace_function with api compatible with std::function.     We can rename inplace_function though.
2. merge inplace_function into delegate
3. merge inplace_function (and delegate?) into etl::function
4. something else?

test/test_inplace_function.cpp
jiangyilism: Fixed.

jwellbelove:
@rolandreichweinbmw The current  `etl::function` (based on virtual functions) is indead deprecated.
I have a complete `std::function` clone in a private branch, although it currently has the same limitation as `etl::delegate` when it comes to lambdas, which is why I 'parked' it for the time being. Its only real advantage over `etl::delegate` was that it had the same API as `std::function` (apart from lambdas) and could accept runtime function pointers.
It's possible I _may_ be able to add runtime function pointers to `etl::delegate` without affecting `constexpr` compatibility.

I don't have any problem with adding functions that enhance the usability of an ETL class; making an `exact` clone of the STL was never my intention.

jwellbelove:
> std::function does not support call_if, call_or, clear .\nstd::function and etl::inplace_function clear its content with operator=(nullptr_t) noexcept

I've always treated the STL as _inspiration_ for classes in the ETL, not as a rigid specification for what can and can't be included.

jwellbelove:
I have limited options to review this PR at the moment, as I'm on holiday in Kos. I'm not quite sure when I will get back, as the Greek air traffic controllers are threatening a strike on the day we are supposed to fly home, and the postponed flights may be backed up for a day or two.

jwellbelove:
I still have some concerns with copyright issues.

_etl renames most of the class and variable names, except the name "inplace_function", which can be renamed too._

This seems to imply that this is not a 'cleanroom implementation', but actually contains a large proportion of copy/pasted code from another copyrighted project.

include/etl/private/inplace_function_cpp11.h
```diff
@@ -0,0 +1,247 @@
...
+      vtable_t(vtable_t&&) ETL_NOEXCEPT ETL_DELETE;
+
+      vtable_t& operator=(const vtable_t&) ETL_DELETE;
+      vtable_t& operator=(vtable_t&&) ETL_DELETE;
+    };
+
+    static const vtable_t default_vtable;
+
+    const vtable_t* vtable_ptr{&default_vtable};
+    mutable char    obj[Capacity];
```
jwellbelove: How can you ensure that this storage is aligned correctly for the type?

sorrowed:
> I still have some concerns with copyright issues.
> 
> _etl renames most of the class and variable names, except the name "inplace_function", which can be renamed too._
> 
> This seems to imply that this is not a 'cleanroom implementation', but actually contains a large proportion of copy/pasted code from another copyrighted project.

I'm assuming a lot of this would be "inspired" by the sg14 implementation.

This implementation however somehow indicates "Copyright(c) 2025 BMW AG". Is that in itself even compatible with the ETL?

jiangyilism:
> I still have some concerns with copyright issues.
> 
> _etl renames most of the class and variable names, except the name "inplace_function", which can be renamed too._
> 
> This seems to imply that this is not a 'cleanroom implementation', but actually contains a large proportion of copy/pasted code from another copyrighted project.

As discuss earlier, `inplace_function` can be renamed too.

include/etl/private/inplace_function_cpp11.h
jiangyilism: Can adopt `etl::aligned_storage` here. However, should we proceed with this PR or compare the implementation with the etl::function v2 ( https://etlcpp.slack.com/archives/C7U0HPFJA/p1742585235826159 ) first?

rolandreichweinbmw:
> This implementation however somehow indicates "Copyright(c) 2025 BMW AG". Is that in itself even compatible with the ETL?

Copyright can be by any contributor. So it is "compatible" be default, and license is MIT.

jiangyilism:
@jwellbelove `inplace_function` it is now renamed to `owning_callable` .
However, the idea is the same as SG14 inplace_function (boost license). So the implementation and the abi will still be similar.

About the aligned_storage adoption, does `etl::owning_callable` really need it?   `std::aligned_storage` is deprecated in c++23 with reasons: https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p1413r2.pdf 
The usage in SG14 ( https://github.com/WG21-SG14/SG14/blob/master/SG14/inplace_function.h#L370 ) does not seem necessary in my opinion. Since 
1. The type-erasure behavior might make the calculated alignment non-optimal in some platforms
2. The alignment does not seem to be a performance bottleneck. Perhaps if we store many `etl::owning_callable` objects in an array then the alignment might make some difference. However, even same capacity-alignment `owning_callable` objects can, in run-time, hold totally different callees with different alignments. So I doubt that compile-time specified alignment value might not improve performance in all cases. While memory overhead is likely to occur in some cases.

sorrowed:
> @jwellbelove `inplace_function` it is now renamed to `owning_callable` . However, the idea is the same as SG14 inplace_function (boost license). So the implementation and the abi will still be similar.

I'm not sure why changing the name (which to me sounds less like what it does than inplace_function, which resembles for examples std::inplace_vector) changes anything copyright related?

jiangyilism:
> > @jwellbelove `inplace_function` it is now renamed to `owning_callable` . However, the idea is the same as SG14 inplace_function (boost license). So the implementation and the abi will still be similar.
> 
> I'm not sure why changing the name (which to me sounds less like what it does than inplace_function, which resembles for examples std::inplace_vector) changes anything copyright related?

No matter how it deviate from sg14, there still share some similarity. All binaries look like a sequence of zeros and ones... If the current deviation is not enough, we need to decide what need to be met to avoid copyright issue.

sorrowed:
> > > @jwellbelove `inplace_function` it is now renamed to `owning_callable` . However, the idea is the same as SG14 inplace_function (boost license). So the implementation and the abi will still be similar.
> > 
> > 
> > I'm not sure why changing the name (which to me sounds less like what it does than inplace_function, which resembles for examples std::inplace_vector) changes anything copyright related?
> 
> No matter how it deviate from sg14, there still share some similarity. All binaries look like a sequence of zeros and ones... If the current deviation is not enough, we need to decide what need to be met to avoid copyright issue.

IANAL, but i don't think using SG14 as a source for your version is a problem as long as you honor the original license. The original Boost license of SG14 states that:

```
 * The copyright notices in the Software and this entire statement, including
 * the above license grant, this restriction and the following disclaimer,
 * must be included in all copies of the Software, in whole or in part, and
 * all derivative works of the Software
```
This means (IMHO, again IANAL) you can't change copyright to BMW AG (or any other entity) if the original copyright was claimed elsewhere and you can't change the license to MIT as the original license should be included.

The alternative is to create a [cleanroom implementation](https://en.wikipedia.org/wiki/Clean-room_design) but this can't be done by you as you already saw & used the SG14 version. Just changing some names in  the implementation is not a solution at this point.

jiangyilism:
If I understand the etl design correctly, we want to avoid introducing boost license.
@sorrowed Having seen sg14's implementation does not mean we can't implement something similar. But IANAL either.
@jwellbelove I think it is better to switch to your "function" v2 approach in another PR and avoid relating to this PR, which contains sg14 discussions, in any way.

jwellbelove:
@coderabbitai review

BartolomeyKant:
Any updates here?
And what is a "function" v2 approach mentioned by @jiangyilism earlier? Is it implemented already?

include/etl/private/owning_callable_cpp11.h
```diff
@@ -0,0 +1,248 @@
...
+
+      ~vtable_t() = default;
+      vtable_t() = default;
+
+      template <typename T>
+      explicit ETL_CONSTEXPR vtable_t(etl::type_identity<T>) ETL_NOEXCEPT
+        : invoke_func{[](char* pobj, Args&&... args) -> R
+                      { return (*reinterpret_cast<T*>(pobj))(
+                          etl::forward<Args>(args)...); }},
+          copy_func{[](char* dst_obj, char* src_obj) -> void
```
BartolomeyKant, line 205: This requires T to be copy_constructible.

Since for some cases copy is not required, I think there also should be implementation variant like std::move_only_function.

jwellbelove:
I've been working on my own variant of `inplace_function` that takes inspiration from both, `inplace_function` design techniques (in class storage, vtable), and `etl::delegate` (compile time initialisation). I'm a bit busy with other work at the moment, but it seems to be progressing quite well.

jwellbelove:
I am developing a version of `etl::inplace_function `that is not based on an existing codebase, that also merges in functionality from `etl:delegate`, such as compile time function pointers.

PR #1048, opened 2025-03-11
Support chrono.h

jiangyilism:
Note that the users of `etl::chrono::high_resolution_clock`
have to provide the definition of the C function `etl_get_system_time_nanosec()` .
If one does not use `etl::chrono::high_resolution_clock` then `etl_get_system_time_nanosec()` can be left undefined/ignored.

### DIFF
--- a/include/etl/chrono.h
+++ b/include/etl/chrono.h
@@ -1,0 +1,577 @@
+/******************************************************************************
+The MIT License(MIT)
+
+Embedded Template Library.
+https://github.com/ETLCPP/etl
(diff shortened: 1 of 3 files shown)


rolandreichweinbmw (approved):

jwellbelove:
I have implemented an almost full implementation of etl:chrono.

rolandreichweinbmw:
Sounds great! Can you already make a branch of it available?

I'm especially interested in how you solved the adapter for the high_resolution_clock to underlying platform's solutions.

PR #1049, opened 2025-03-12
Various cleanup changes

rolandreichweinbmw:
This is a collection of different cleanup changes:

* intrusive_forward_list.h: Remove remove() by pointer because erase() can be used for that
* algorithm.h: Fix signed distance handling in copy_s and move_s, with added check for order; added test
* Fix File IDs


include/etl/intrusive_forward_list.h
```diff
@@ -991,29 +991,6 @@ namespace etl
       }
     }
 
-    //*************************************************************************
```
jwellbelove, line 994: Why did you remove this member function?

rolandreichweinbmw: I only later noticed that you can use the previously existing erase() instead. Therefore, I don't need the remove() anymore because I can use erase() interchangeably.

jwellbelove: `remove` and `erase` do different things.
`remove` searches the list for a node that equals the value of the parameter.
`erase` unlinks the node passed in the parameter.

rolandreichweinbmw: In this PR, I'm removing only the remove() overload that provides the erase() use case.

The other remove() overload - directly above in the code - stays.

PR #1072, opened 2025-04-12
Add swap(circular_buffer_ext&&) (#1068)

### LINKED ISSUE #1068: Request: add swap(circular_buffer_ext&&) for circular_buffer_ext
Hello,

I noticed that `etl::circular_buffer_ext` lacks an overload in `swap()` that takes in an rvalue reference. This forces us to create an unnecessary temporary variable. Are there plans to add it?

helderTZ:
This change adds an overload for circular_buffer_ext's swap() for rvalue references and a corresponding test.
Closes #1068

### DIFF
--- a/include/etl/circular_buffer.h
+++ b/include/etl/circular_buffer.h
@@ -1385,6 +1385,25 @@ namespace etl
 #endif
     }
 
+#if ETL_USING_CPP11
+    //*************************************************************************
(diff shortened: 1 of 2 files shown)


jwellbelove:
I'm accepting the pull request, but this is part of a much larger issue regarding how `_ext` containers are moved, which will need to be addressed at some point.

Issue #1091, opened 2025-05-09
Clang20 Build Failure for C++17 and newer


wsciaroni:
I have seen a build failure due to a deprecation warning when using clang version 20. This is tested on the latest version of master 5896b3b90e401ec991ceda0097f8da3c8e9d8e9d

The build log is:

```log
--- Configuration Details ---
Compiler ID: clang20
C++ Standard (ETL_CXX_STANDARD): 17
NO_STL: OFF
ETL_FORCE_TEST_CPP03: OFF
CC: clang, CXX: clang++
--- Compiler Version ---
Debian clang version 20.1.4 (++20250426011826+182e8b7f8a71-1~exp1~20250426011849.108)
Target: x86_64-pc-linux-gnu
Thread model: posix
InstalledDir: /usr/lib/llvm-20/bin
--- CMake Version ---
cmake version 3.25.1

CMake suite maintained and supported by Kitware (kitware.com/cmake).
---------------------------
-- etl | Determined ETL version 20.40.0 from version.txt file
-- The CXX compiler identification is Clang 20.1.4
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/clang++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Compiling for C++17
-- Compiling for STL
-- Using Clang compiler
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Success
-- Found Threads: TRUE  
-- Configuring done
-- Generating done
CMake Warning:
  Manually-specified variables were not used by the project:

    ETL_FORCE_TEST_CPP03


-- Build files have been written to: /__w/etl/etl/build
[  0%] Building CXX object test/UnitTest++/CMakeFiles/UnitTestpp.dir/AssertException.cpp.o
[  0%] Building CXX object test/UnitTest++/CMakeFiles/UnitTestpp.dir/CompositeTestReporter.cpp.o
[  0%] Building CXX object test/UnitTest++/CMakeFiles/UnitTestpp.dir/Checks.cpp.o
[  1%] Building CXX object test/UnitTest++/CMakeFiles/UnitTestpp.dir/CurrentTest.cpp.o
[  1%] Building CXX object test/UnitTest++/CMakeFiles/UnitTestpp.dir/DeferredTestReporter.cpp.o
[...]

PR #1093, opened 2025-05-10
 Add Devcontainer configurations

wsciaroni:
This PR introduces `.devcontainer.json` files for a range of compiler versions (clang7-20 and gcc9-15).

This enhancement provides a consistent and isolated development environment for contributors. By using these configurations, developers can easily spin up pre-configured environments with the specified compilers, and their dependencies.

This will help to:

-   Ensure consistent build environments across different machines.
-   Simplify the setup process for new contributors when tracking compiler version specific issues.
-   Potentially reduce "it works on my machine" issues.

No code changes to the library itself are included in this PR.


.devcontainer/context/reinstall-cmake.sh
```diff
@@ -0,0 +1,295 @@
+#!/usr/bin/env bash
+#-------------------------------------------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See https://go.microsoft.com/fwlink/?linkid=2090316 for license information.
```
rolandreichweinbmw: Maybe just refer to the LICENSE file?

wsciaroni: Ah, this was left by mistake. This license no longer applies as I have rewritten the entire file. That license was automatically added by VSCode.

wsciaroni: Removed

.devcontainer/clang18/devcontainer.json
```diff
@@ -0,0 +1,12 @@
+// For format details, see https://aka.ms/devcontainer.json. For config options, see the
+// README at: https://github.com/devcontainers/templates/tree/main/src/cpp
+{
+	"name": "Clang 18",
+	"build": {
+		"dockerfile": "../Dockerfile",
+		"args": {
+			"BASE_IMAGE_NAME": "silkeh/clang:18-bullseye"
```
rolandreichweinbmw, line 8: As far as I can see, devcontainers uses Debian 11 "bullseye" by default? Why not using "bookworm" (Debian 12) instead, for the newer clang versions?

(With Debian 13 being expected to be released soon).

wsciaroni: I agree with your assessment, but when I use the default, the c++17 and newer builds fail (see #1091 for the error those images cause). When I was explicit with the version, I was able to build successfully.

This applies to the clang 18, 19, and 20 images.

PR #1103, opened 2025-05-19
Add IWYU pragmas to private headers which provide library symbols

mlandersonjlg:
This prevents warnings in clang compiler and IWYU tool 

Example for `#include "etl/delegate.h"`

```
Included header delegate.h is not used directly (fix available)clangd[unused-includes](https://clangd.llvm.org/guides/include-cleaner)
```

https://clangd.llvm.org/guides/include-cleaner#iwyu-pragmas 
https://github.com/include-what-you-use/include-what-you-use/blob/master/docs/IWYUPragmas.md#iwyu-pragma-export


drewr95 (approved):

drewr95:
This is nice when using clang-tidy as it always flags these headers and suggests to fix by including the direct private headers.  This also avoids having to set the .clang-tidy file to blanketly ignore these headers.

PR #1125, opened 2025-06-11
For cogapp, use python3 instead of python

rolandreichweinbmw:


jwellbelove:
I've just tried using 'cog' and that work fine for Windows 10 and Unbuntu 22.04.

rolandreichweinbmw:
Ok, adjusted accordingly.

jwellbelove:
Also, the python install on Windows defaults to python.exe, not python3.exe.

rolandreichweinbmw:
Fine - then the direct cog call should do. Or do you mean certain further adjustments?

jwellbelove:
A direct cog call should be fine, as far as I can tell.

PR #1182, opened 2025-09-08
Fix etl::typed_storage by supporting omitted destructors

rolandreichweinbmw:
In a recent change to alignment.h, the etl::typed_storage was changed in a way that in case of an already constructed object, the object is created via assignment.

However, this contradicts the original use case that led to etl::typed_storage in the first place:
https://github.com/ETLCPP/etl/pull/1023

The goal is to omit destructors (and at the same time support classes with deleted assignment operators), so they can be optimized out at link time.

This change reverts commit ac7b268 to restore the original functionality and changes the test to reflect the original use case.


rolandreichweinbmw:
Thanks to the coderabbitai review, I also fixed an unrelated bug (missing create()) and typos in alignment.h.

thomas-herrmann-1:
Sorry, I just have found this pull request, maybe a bit late. It addresses a problem also connected to the discussion [https://github.com/ETLCPP/etl/discussions/1170]() that I started some weeks ago. The inner structure of `typed_storage` is not giving any hint on the contained type which is not really easy to use for debugging and further analysis. Therefore I suggested to use the same representation as used for `optional`. And with the proposed patch I've also fixed the problem of the lately introduced assignment operator. While it wasn't clear to me who actually brings in this patch I haven't converted this into a pull request yet. Maybe the patch 
[0001-Change-to-typed_storage-without-requiring-operator.patch.txt](https://github.com/user-attachments/files/22248089/0001-Change-to-typed_storage-without-requiring-operator.patch.txt) could be used for reaching both goals?

rolandreichweinbmw:
Hi @thomas-herrmann-1, your patch was already completely applied to master and released in ETL 20.43.1 :-)

However, John added further changes to alignment.h (see the has_value() branches in the create() members). This unfortunately breaks the original use case for etl::type_storage where we want to optimize out destructors.

This PR https://github.com/ETLCPP/etl/pull/1182 fixes this.

Independently from your already included patch.

thomas-herrmann-1:
@rolandreichweinbmw thanks for the hint! I haven't seen the PR for my patch and am now a bit confused. When I pointed out to the problem with the newly introduced assignment operator John was responding with:
> I overlooked that that I had added new conditions for the type stored.

Therefore I didn't expect any later change like that.

PR #1185, opened 2025-09-13
Remove noexcept from delegate method stubs.

denravonska:
In addition to removing noexcept from call_if, this is also needed to prevent an abort when cancelling a pthread that is executing a delegate.


jwellbelove:
Sorry, I missed that.
Too much in a rush last night.

denravonska:
> Sorry, I missed that. Too much in a rush last night.

No worries, I did the exact same mistake when I tested it earlier this week.
I can confirm that what's in `development/` fixes all the threading and delegate issues for us.

PR #1186, opened 2025-09-17
Guards around usage of std::initializer_list in optional.h

rolandreichweinbmw:
In optional.h, ETL_HAS_INITIALIZER_LIST is being used to guard against cases where std::initializer_list is not available. But not consistently. This changes fixes it by adding it in the remaining places.


rolandreichweinbmw:
@coderabbit full review

PR #1188, opened 2025-09-22
etl array checks

mike919192:
This is reworking the previous PR #1175 based on the discussion in that PR.  In that discussion we decided that the additional checks should be enabled the same way that the existing `ETL_CHECK_PUSH_POP` is.

2 new check macros are added `ETL_CHECK_INDEX_OPERATOR` and `ETL_CHECK_EXTRA`.  `ETL_CHECK_INDEX_OPERATOR` is intended for checks using the [ ] operator.  `ETL_CHECK_EXTRA` is for any extra checks that don't fall into an existing check option.

The added checks are implemented only on `etl::array`.  Moving forward it can be implemented on more classes in the future.

### DIFF
--- a/include/etl/error_handler.h
+++ b/include/etl/error_handler.h
@@ -364,10 +364,6 @@ namespace etl
   #endif
 #endif
 
-#if defined(ETL_VERBOSE_ERRORS)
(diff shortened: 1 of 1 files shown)


mike919192:
@jwellbelove Just FYI, the branch that this was merged into still contains the ETL_DEBUG_ASSERT macros.  I was fine with removing those if we were not going to use that approach.

PR #1191, opened 2025-09-26
Introduce Cast to void for Condition of Assert

christophruethingbmw:
Currently, in case we use a parameter of a function _only_ inside of an `ETL_ASSERT` and the ETL configuration disables the ETL_ASSERT, we get a compiler warning about an unused parameter. Therefore, this change casts the condition of ETL_ASSERT to void.


rolandreichweinbmw (approved):

jwellbelove:
ChatGPT points out that using `(void)sizeof(b);` would get around the issue of `b` being evaluated.

christophruethingbmw:
Hi @jwellbelove, I am unsure what is the intended behavior, I agree that using `sizeof(b)` would never evaluate `b`. My two cents:

- In case somebody adds an expression with side-effects into an `ETL_ASSERT`, we might even want that it is not eliminated? I think this is not intended usage of this, i.e. the expression inside of `ETL_ASSERT` should neve have side effects, but you never now. Like it is done here, we have a t least a consistent behavior for all ETL profile settings.
- I would argue that any decent compiler will optimize away the evaluation of the expression in case it has no side-effects. I think that's the case for most of the usages.

I have no strong preference here, what do you think?

jwellbelove:
The standard `assert` will not evaluate arguments when not in a debug build, so I think that stating the same for `ETL_ASSERT` would be valid.

christophruethingbmw:
@jwellbelove , okay, I changed it to `static_cast<void>(sizeof(b))`. Can you please have a look?

PR #1192, opened 2025-09-26
Allow easy Creation of Generic Exceptions with just a Text

christophruethingbmw:
Today when using ETL_ASSERT you need to pass a concrete exception type. However, it can be quite unhandy to define a custom exception type for each assert, therefore we want to also allow to simply use the generic etl::exception type and directly provide the message.


jwellbelove:
I will merge this, but I may change the macro name.

PR #1195, opened 2025-10-02
Added conditional support of noexcept to fix IAR support

ockeydockey:
When exceptions are disabled in IAR 8.x, the keyword `noexcept` in any form will cause a build error.

Adding back in the conditional support for the `noexcept` keyword based on the etl_profile settings.

### DIFF
--- a/include/etl/platform.h
+++ b/include/etl/platform.h
@@ -331,8 +331,13 @@ SOFTWARE.
   #define ETL_ENUM_CLASS(name)            enum class name
   #define ETL_ENUM_CLASS_TYPE(name, type) enum class name : type
   #define ETL_LVALUE_REF_QUALIFIER        &
-  #define ETL_NOEXCEPT                    noexcept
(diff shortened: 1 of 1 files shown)


include/etl/platform.h
```diff
@@ -331,8 +331,13 @@ SOFTWARE.
   #define ETL_ENUM_CLASS(name)            enum class name
   #define ETL_ENUM_CLASS_TYPE(name, type) enum class name : type
   #define ETL_LVALUE_REF_QUALIFIER        &
-  #define ETL_NOEXCEPT                    noexcept
-  #define ETL_NOEXCEPT_EXPR(...)          noexcept(__VA_ARGS__)
+  #if ETL_USING_EXCEPTIONS
```
ockeydockey, line 334: @jwellbelove, if you don't like using this flag, I can also recommend using the feature test macro in the C++ standard:

```cpp
#if defined(__cpp_exceptions)
```

PR #1196, opened 2025-10-03
Added back in support for populating non 8-bit strings with 8-bit data

ockeydockey:
There was an issue where using a non-16-bit pointer to assign data to a u16string would attempt to use the `copy_characters()` which uses the `memove()` call.  This further specializes the template metaprogramming to use a for loop for any iterators that aren't `uint16_t*` or `uint16_t const*`.

This change was also applied to all other non 8-bit strings (u32 & wchar).

### DIFF
--- a/test/test_string_u16.cpp
+++ b/test/test_string_u16.cpp
@@ -321,7 +321,7 @@ namespace
     //*************************************************************************
     TEST_FIXTURE(SetupFixture, test_constructor_8bit_const_range)
     {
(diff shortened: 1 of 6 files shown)


jwellbelove:
Doesn't this change also apply to u32string?

jwellbelove:
And wstring?

ockeydockey:
Good point about the other string types.  I've added support and tests for those other string types.

PR #1199, opened 2025-10-06
Added support for transitions on state enter for HFSM

ockeydockey:
This adds the capability for the HFSM to transition while entering any state (just like with the FSM).  If a parent state signifies a transition before the child is entered, the transition will take place from the parent without entering the child state.

This also add the ability for a self-transition in the FSM to transition on enter.

### DIFF
--- a/test/test_hfsm_transition_on_enter.cpp
+++ b/test/test_hfsm_transition_on_enter.cpp
@@ -30,6 +30,7 @@ SOFTWARE.
 
 #include "etl/hfsm.h"
 #include "etl/string.h"
+#include <array>
 
 //                             entry && mode=RxEventDeviation
 //                 ┌─────────────────────────────────────────────────────┐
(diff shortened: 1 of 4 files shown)


test/test_hfsm_transition_on_enter.cpp
```diff
@@ -0,0 +1,362 @@
...
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
+******************************************************************************/
+
+#include "unit_test_framework.h"
+
+#include "etl/hfsm.h"
+#include "etl/string.h"
+
```
jwellbelove, line 34: Look like you're missing a `#include <array>`

ockeydockey: Ah, thanks.  That makes sense.

PR #1201, opened 2025-10-08
etl::span checks

mike919192:
This add the optional assert checks to etl::span.  Several of the span checks needed to throw from C++11 constexpr so the ternary operation was used where needed.

Adding macros `ETL_CHECKING_XXX` and `ETL_NOT_CHECKING_XXX` for the optional index operator checks and extra checks.  This makes the macro usage a little simpler compared to checking if the right macro is defined.

Added `span_out_of_range` exception.  Fixed the `at` functions were using the wrong exception (`array_out_of_range`).

Some of the static asserts in the static span class where unessesarily checking that `Extent != etl::dynamic_extent`.

### DIFF
--- a/include/etl/span.h
+++ b/include/etl/span.h
@@ -111,7 +111,7 @@ namespace etl
   public:
 
     span_out_of_range(string_type file_name_, numeric_type line_number_)
-      : span_exception("array:range", file_name_, line_number_)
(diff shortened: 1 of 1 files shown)


mike919192:
@coderabbitai review

PR #1202, opened 2025-10-08
Catch reentrant calls to FSM and HFSM start, receive, etc.

ockeydockey:
Adding the ability to catch when a reentrant call to a method which can change state has been made.  Since transitioning states while another state transition is active could cause state corruption, infinite loops, or other issues, this new mechanism will fail an assert.

The most common example would be a state transition (due to calling `receive()`) triggering another call to `receive()` before the state transitions have finished processing.  This second call will now throw an error.


ockeydockey:
@jwellbelove, sorry for spamming your inbox this month.  This is the last item I should need to push for a while.

PR #1204, opened 2025-10-10
Add etl::format

rolandreichweinbmw:
This adds etl::format, guided by std::format, avoiding dynamic memory allocation.

Also, synchronizes C++17 for "Windows - STL" and "Windows - No STL" due to different char* iteration behavior in VS 2022.


positron96:
Wow, this is a really great addition to the library! Thank you all!

PR #1210, opened 2025-10-20
Bugfix: Fixed span reinterpret_as should return fixed span

mike919192:
Previously, the fixed span reinterpret_as function forced returning a dynamic span.  It is possible instead to return a new fixed span using the compile time information.

Tests are updated to always work with fixed spans.


rolandreichweinbmw:
Looks good to me.

Even though the proposal by coderabbit is wrong since it might be useful to reinterpret an initial subspan instead of the complete one, as done when parsing initial bytes of input.

mike919192:
> Looks good to me.
> 
> Even though the proposal by coderabbit is wrong since it might be useful to reinterpret an initial subspan instead of the complete one, as done when parsing initial bytes of input.

Yeah the round down behavior seems fine to me.  I don't see any reason to not allow `etl::span<uint8_t, 9>` to be reinterpreted as `etl::span<uint32_t, 2>`

PR #1211, opened 2025-10-22
Change generators to reflect changes in fsm.h and type_traits.h

rolandreichweinbmw:
Previous changes were wrongfully made in fsm.h and type_traits.h instead of in their generator counterparts.

Add CI check to ensure generated files are in sync.


rolandreichweinbmw:
I just tested with a test commit that the CI returns an error if the generator headers and the generated headers are not in sync.

PR #1214, opened 2025-10-23
Spare more Resources for non-verbose Errors

christophruethingbmw:
At the moment we only remove the __FILE__ in case ETL_VERBOSE_ERRORS is not set. However, we also want to remove the error text which can consume quite some resources and without the file name the line number is also not very useful.


jwellbelove:
I can see the point about that line numbers without a filename are a bit pointless, but I'm not sure I like the idea of removing the `reason` field completely, as this could be a breaking change for those that use it in terse error mode.

I think having a third option would be better.
Enabled by `ETL_SIMPLE_ERRORS`?

christophruethingbmw:
@jwellbelove , thank you for the response.

Compilation would most likely not break, since the `what()` stays and just returns an empty string. However, I agree, that behavior-wise it could break in case somebody relies on the `reason` text even with disabled `VERBOSE_ERRORS`.

So you propose to introduce another flag for only disabling the `reason` instead of using the `VERBOSE_ERRORS`? That's definately an option but I am not sure if it complicates the thing more than needed. I personally have the feeling that there are already quite some flags for error handling and adding another one just makes it harder to understand.

In case we want to maintain backwards compatability as much as possible we can go this way.

jwellbelove:
I would propose, to avoid issues with run-time backward compatibility:-

`#define ETL_VERBOSE_ERRORS`
  Current functionality
  Verbose reason, file name, line number
  Example: `reason` for `etl::array_view`: `"array_view:bounds"`

`#define ETL_MINIMAL_ERRORS`
  No reason, no file name, no line number.

Default behaviour (no macro defined)
  Terse reason, no file name, no line numbers.
  Example: `reason` for `etl::array_view`: `"41A"`

christophruethingbmw:
@jwellbelove sounds okay for me, I will adapt this PR to this proposal, just one detail: what do you want to do in case both are defined? do we want to forbid this?

jwellbelove:
Yes, if both are defined then an error message should be output.

christophruethingbmw:
@jwellbelove , I updated the code and will push it soon. But one follow-up questions: I think it would be good to also update the documentation on the website. I think this is not part of this repository?

image

jwellbelove:
The documentation is a manually edited website, which was fine in the beginning. But now the ETL is over 10 years old, it's somewhat problematic.

christophruethingbmw:
@jwellbelove , can you have a look again? I think the two OSX jobs fail unrelated to these changes.

jwellbelove:
I noticed that this morning.
The first issue was that macos-13 was not longer supported.
The second issue, that I'm not sure I solved at the time, was the macos compuler couldn't find std::alignment_of.
I'll take a look when I get back home later.

Issue #1207, opened 2025-10-17
Monadic Operations for etl::expected


whitfijs-jw:
I have been playing around with adding monadic operations (and_then, or_else, transform, transform_error) to etl::expected. I couldn't find any mention of adding them to the project, so I apologize if this has already been discussed.  

A first draft of the implementation is [here](https://github.com/whitfijs-jw/etl/tree/topic/expected-monadic-operations) in expected.h, there are still some static_asserts and stl calls that would need to be swapped out for their etl equivalents.  I also added simple, non-exhaustive unit tests, but I haven't run them against the full suite of compilers and c++ standard. 

An example of transform and some helper traits:

```
  template <typename F, typename T>
  struct deduced_result_type {
    using type = std::decay_t<decltype(std::declval<F>()(std::declval<T&&>()))>;
  };

  template <typename F, typename T>
  using deduced_result_type_t = typename deduced_result_type<F, T>::type;

  template <typename F>
  auto transform(F&& f) const & {
    using U = typename deduced_result_type<F, TValue>::type;
    using new_expected = expected<U,TError>;

    if (has_value()) {
        return new_expected(f(etl::get<TValue>(storage)));
    } else {
        return new_expected(
          unexpected<TError>(etl::get<TError>(storage))
        );
    }
  }

  template <typename F>
  auto transform(F&& f) && {
    using U = typename deduced_result_type<F, TValue>::type;
    using new_expected = expected<U,TError>;

    if (has_value()) {
        return new_expected(f(etl::get<TValue>(etl::move(storage))));
    } else {
        return new_expected(
          unexpected<TError>(etl::get<TError>(etl::move(storage)))
        );
    }
  }
```
and an example usage in a simple unit test:
```
    TEST(test_transform_with_value) {
      struct Transform{
        Expected produceValue() {
          return Value("produceValue()");
        }
      };

      Transform transform;
      auto expected = transform.produceValue().transform([](Value v) {
        CHECK_EQUAL("produceValue()", v.v);
        auto s = v.v.append(" transformed");
        return s;
      });
      CHECK_TRUE(expected.has_value());
      CHECK_TRUE(bool(expected));
      CHECK_EQUAL("produceValue() transformed", expected.value());
    }
```

This is far from polished, but I didn't want to spend too much more time on it if there's no interest in adding it to the project.

jwellbelove:
20.45.0

PR #1215, opened 2025-10-24
Expected monadic operations

whitfijs-jw:
#1207 
C++11 compatible implementation of and_then, or_else, transform, and transform_error 

I'm not 100% satisfied with the way the unit tests turned out.  Suggestions as to how to clean them up are welcome.


whitfijs-jw:
I was a little premature on creating a PR, apologies

PR #1217, opened 2025-10-29
Add const specialization for etl::get_object_at

filipe-cuim:
# Description

A specialization for etl::get_object_at where it receives a pointer to const data, is provided, where it returns a const reference to the type we want to reinterpret the data


test/test_memory.cpp
```diff
@@ -1612,6 +1612,33 @@ namespace
...
+          flags.constructed = true;
+        }
+
+        ~Data() = default;
+
+        int a;
+        int b;
+      };
+
+      etl::array<uint8_t, 32U> buffer{};
```
rolandreichweinbmw: Add:

#include "etl/array.h"

to fix the test build error?

filipe-cuim: Used std::array instead

PR #1232, opened 2025-12-05
Exception std based option

ivanyib:
When a project already handles std::exceptions, it's much easier to adopt ETL if the exceptions thrown by the library are derived from std::exception. Etl::exception's interface is already very similar to std::exception's, so this change isn't too intrusive.
Clang-tidy also has a check that recommends throwing instances of std::exception.
https://clang.llvm.org/extra/clang-tidy/checks/hicpp/exception-baseclass.html

The reason for the constructors and the what function losing their constexprness in this case is that the constructor and the what function of std::exception is not constexpr and a constexpr constructor cannot call a non-constexpr constructor of the base, and the virtual what function has to match the declaration in the base.

### DIFF
--- a/include/etl/platform.h
+++ b/include/etl/platform.h
@@ -632,6 +632,7 @@ namespace etl
     static ETL_CONSTANT bool using_legacy_bitset              = (ETL_USING_LEGACY_BITSET == 1);
     static ETL_CONSTANT bool using_exceptions                 = (ETL_USING_EXCEPTIONS == 1);
     static ETL_CONSTANT bool using_libc_wchar_h               = (ETL_USING_LIBC_WCHAR_H == 1);
+    static ETL_CONSTANT bool using_std_exception              = (ETL_USING_STD_EXCEPTION == 1);
     
     // Has...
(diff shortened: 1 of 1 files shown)


include/etl/platform.h
```diff
@@ -293,6 +293,17 @@ SOFTWARE.
   #define ETL_HAS_VIRTUAL_MESSAGES 1
 #endif
 
+//*************************************
+// Indicate if etl::exception is to be derived from std::exception.
+#if defined(ETL_EXCEPTION_STD_BASE)
+#if ETL_NOT_USING_STL
+  #error "Requested std base for etl::exception, but STL is not used"
+#endif
```
jwellbelove, line 301: I would prefer the wording of the macros to be these, to keep in line with the common macro style in the ETL.:-

`#if defined(ETL_USE_STD_EXCEPTION)`
...
```
  #define ETL_USING_STD_EXCEPTION 1
#else
  #define ETL_USING_STD_EXCEPTION 0
#endif
```

Also, add `ETL_USING_STD_EXCEPTION` to the `etl::traits` namespace of `constexpr bools`.

PR #1248, opened 2025-12-26
Feature/#1233 add etl  inplace function

jwellbelove:
Added an owning function wrapper that contains its own storage.
Borrows some functionality from etl::delegate.
Adds invoke classes.


jwellbelove:
This PR pulled in historical changes from development for some reason.

PR #1251, opened 2025-12-31
Add etl::inplace_function

jwellbelove:
Added `etl::inplace_function`
Also added as part of the implementation `etl::invoke` and related classes

### DIFF
--- a/test/test_base64_RFC4648_URL_decoder_with_padding.cpp
+++ b/test/test_base64_RFC4648_URL_decoder_with_padding.cpp
@@ -653,7 +653,6 @@ namespace
         b64.flush();
         CHECK_TRUE(received_final_block);
 
-        std::vector<unsigned char> expected(input_data.begin(), std::next(input_data.begin(), i));
 #include "etl/private/diagnostic_null_dereference_push.h"
         std::vector<unsigned char> expected(input_data.begin(), input_data.begin() + i);
         std::vector<unsigned char> actual(decoded_output);


rolandreichweinbmw:
I noticed that in this PR, the >80 checks as done in other PRs are not run.

Is this intentional?

include/etl/inplace_function.h
```diff
@@ -0,0 +1,1531 @@
...
+      //*****************************************
+      // Member function pointer
+      //*****************************************
+      template <typename TObject>
+      static const inplace_function_vtable* for_member()
+      {
+        using target_t = member_target<TObject>;
+        constexpr bool destroy_src_on_move = !etl::is_trivially_destructible<TObject>::value;
+
+        destroy_type destroy_ptr = etl::is_trivially_destructible<TObject>::value ? nullptr
```
rolandreichweinbmw: Can this be made constexpr?

include/etl/inplace_function.h
```diff
@@ -0,0 +1,1531 @@
...
+
+    //*************************************************************************
+    /// Construct from const lambda/functor (runtime).
+    /// \tparam TLambda The lambda/functor type.
+    /// \param lambda  The lambda/functor.
+    //*************************************************************************
+    template <typename TLambda,
+              typename T = typename etl::decay<TLambda>::type,
+              typename = etl::enable_if_t<etl::is_class<T>::value && !is_inplace_function<T>::value, void>>
+      inplace_function(const TLambda& lambda)
```
rolandreichweinbmw: Why different indentation than in previous (non const) overload?

include/etl/inplace_function.h
```diff
@@ -0,0 +1,1531 @@
...
+  /// \return The constructed inplace_function.
+  //*************************************************************************
+  template <typename TObject, typename TReturn, typename... TArgs>
+  ETL_NODISCARD
+  etl::inplace_function<TReturn(TArgs...),
+                               sizeof(typename etl::private_inplace_function::inplace_function_vtable<TReturn, TArgs...>::template member_target<TObject>),
+                               alignof(typename etl::private_inplace_function::inplace_function_vtable<TReturn, TArgs...>::template member_target<TObject>)>
+    make_inplace_function(TReturn (TObject::*method)(TArgs...), TObject& obj)
+  {
+    using target_t = typename etl::private_inplace_function::inplace_function_vtable<TReturn, TArgs...>::template member_target<TObject>;
```
rolandreichweinbmw: Since this type is already stated above twice, maybe it can be defined commonly? E.g. in the template signature as in the other, below make_inplace_function overload (typename T = ...)

test/test_inplace_function.cpp
```diff
@@ -0,0 +1,1762 @@
...
+      // These should not trigger any static asserts.
+      decltype(d1) d1_1_copy(d1);
+      decltype(d1) d1_4_copy(d4);
+
+      decltype(d2) d2_1_copy(d1);
+      decltype(d2) d2_2_copy(d2);
+
+      decltype(d3) d3_3_copy(d3);
+      decltype(d3) d3_4_copy(d4);
+    }
```
rolandreichweinbmw, line 1433: To have some actual checks here, we could add a check involving the runtime size() maybe?

include/etl/inplace_function.h
drewr95: Agreed, this should be `constexpr`.  At the very least `const`

include/etl/inplace_function.h
```diff
@@ -0,0 +1,1531 @@
...
+      // Const member function pointer
+      //*****************************************
+      template <typename TObject>
+      static const inplace_function_vtable* for_const_member()
+      {
+        using target_t = const_member_target<TObject>;
+        constexpr bool destroy_src_on_move = !etl::is_trivially_destructible<TObject>::value;
+
+        destroy_type destroy_ptr = etl::is_trivially_destructible<TObject>::value ? nullptr
+                                                                                  : &inplace_function_vtable::template destroy_stub<target_t>;
```
drewr95: Same comment from @rolandreichweinbmw  (https://github.com/ETLCPP/etl/pull/1251/changes#r2659796174) here about possibly being `constexpr`

include/etl/inplace_function.h
```diff
@@ -0,0 +1,1531 @@
...
+      {
+        (*Object)(etl::forward<TArgs>(a)...);
+      }
+
+      //*****************************************
+      // vtable factories
+      //*****************************************
+      // Free function pointer
+      //*****************************************
+      static const inplace_function_vtable* for_function_ptr()
```
drewr95, line 323: If it's never possible to return a `nullptr` from this, I think it would be better to return the `vtable` reference.  Then store the address of it whoever wants to store it.

drewr95 (approved):
Looks really good.  Just so I'm clear though, this is an additional feature that is going to be beside `delegate` or is this meant to replace it?  The nice thing about delegates is that they can be placed in ROM whereas I don't think these `inplace_function` can be?

jwellbelove:
> Looks really good. Just so I'm clear though, this is an additional feature that is going to be beside `delegate` or is this meant to replace it?

`etl::inplace_function` does not replace `etl::delegate`.

`etl::inplace_function` has a storage size and alignment dependant on the template parameters; it will usually require more storage than `etl::delegate`. It can only be copied to another if the destination's storage size and alignment are compatible.
`etl::inplace_function` would only be able to be `constexpr` with C++20, and using the STL or a reverse-engineered `std::contruct_at` that could fool the compiler into thinking it was its own version.

`etl::delegate` has a fixed storage size and therefore any `delegate` instance can be copied to another.
`etl::delegate` is able to be `constexpr`.

jwellbelove:
`etl::inplace_function` also borrows from `etl::delegate` in that it supports compile time binding for C++11 and above.

jwellbelove:
> I noticed that in this PR, the >80 checks as done in other PRs are not run.

I created this PR from a feature branch and when I do changes on the feature branch they are reflected in the PR page.
It looks like I need to add `synchronize` to the type on the workflow.
Edit: I think I need to add development to `branches:` too.

include/etl/inplace_function.h
jwellbelove: It's a valid point, but as it is only usable within the internal functionality of `etl::inplace_function` I'm not sure it really makes much difference either way. If it were part of the public API then I would definitely say yes.

PR #1253, opened 2026-01-03
use correct parameter in ETS_ASSERT_FAIL (and friends) when ETL_NO_CHECKS is defined

hasanayan:
it seems when `ETL_NO_CHECKS` is defined, following macros refer to parameter `b` instead of `e`
`ETL_ASSERT_FAIL`
`ETL_ASSERT_FAIL_AND_RETURN`
`ETL_ASSERT_FAIL_AND_RETURN_VALUE`


jwellbelove:
Thanks for the PR, but a fix that solves this issue has already been made in development.

Issue #1250, opened 2025-12-31
vector / vector_ext swap function


mike919192:
I noticed that the vector classes are missing `swap()` functions and I also saw there was a similar issue #1068 .  I started doing a little experimentation but I want to check on some design questions before going too far.

`vector_ext` swap could be implemented by swapping the pointers, similar to the previous issue.  But I think that `vector` with internal storage must do per element swap and resizing.  Then also for swapping between `vector` and `vector_ext`, again per element swap and resizing would need to be done.

Let me know if there is any problem with this approach.  

The STL api for reference:

https://en.cppreference.com/w/cpp/container/vector/swap.html

jwellbelove:
`vector_ext` could be swapped by just swapping pointers, and you are right about a vector <=> vector_ext swap.

There is a larger question, that I have tried to look at in the past, about move for `vector_ext.` At the moment a `move` means to move the `elements` in the `vector_ext`, not 'stealing' the storage.

If a `swap` on a `vector_ext` means swapping the pointers, then it could be argued that a `move` should do something similar. 

Unfortunately this could be a breaking change, as currently a user can be sure that the buffer address stored in a `vector_ext` will never be changed. This expectation is also broken with a `swap` that swaps pointers.

If I were to implement low overhead moves/swaps for `_ext` containers then I think this should be done in a major release, with a big warning to users that want to upgrade.

mike919192:
Yeah I also had the same thought about the move constructors, but I didn't mess with them yet because of the breaking change.

Do you want to introduce the per element swap and resizing for `ivector` and the derived classes first?  Then later implement the swapping pointers for the `_ext` classes and breaking change to move constructors.

The only concern I had was it might be confusing to someone when the elements are swapped versus when just the pointers are swapped.  But in the end I don't think its too bad.

Then everything probably also applies the string and string_ext as well.

jwellbelove:
20.45.0

PR #1256, opened 2026-01-06
ivector swap

mike919192:
Here is proposed implementation of ivector swap discussed in #1250 

This does per element swap and resizing of two ivectors.  The capacities are permitted to be different and there is a check at the beginning to check that both ivectors have enough capacity to perform the swap.

cppreference links:
https://en.cppreference.com/w/cpp/container/vector/swap.html
https://en.cppreference.com/w/cpp/container/inplace_vector/swap.html


include/etl/private/ivectorpointer.h
```diff
@@ -1040,6 +1114,28 @@ namespace etl
...
+  //*********************************************************************
+  /// Overloaded swap for etl::ivector<const T*> 
+  ///\param lhs The first vector to swap with.
+  ///\param rhs The second vector to swap with.
+  //*********************************************************************
+  template<typename T>
+  void swap(ivector<const T*>& lhs, ivector<const T*>& rhs)
+  {
+    lhs.swap(rhs);
+  }
```
rolandreichweinbmw: coderabbitai's comment regarding redundancy of those free overloads might be correct.

Can you explain why they are really necessary?

mike919192: This is a funny AI interaction, because I ran coderabbit offline and it suggested adding this overload in the first place, haha.  I think the reasoning it gave was something about ADL which I'm not really an expert in so I can't say if it was valid or not.

But you are probably right, I will check if the CI is still ok with it removed.

mike919192: I removed the unnecessary overloads.  Thanks

include/etl/vector.h
```diff
@@ -947,6 +947,43 @@ namespace etl
...
+
+      ivector<T>& smaller = other.size() > this->size() ? *this : other;
+      ivector<T>& larger = other.size() > this->size() ? other : *this;
+
+      typename ivector<T>::iterator smaller_itr = smaller.begin();
+      typename ivector<T>::iterator larger_itr = larger.begin();
+
+      while(smaller_itr < smaller.end())
+      {
+        ETL_OR_STD::swap(*smaller_itr, *larger_itr);
```
jwellbelove: Could this be replaced with
`ETL_OR_STD::swap_ranges(smaller.begin(), smaller.end(), larger.begin());`

and similar for 978-982.

include/etl/vector.h
```diff
@@ -947,6 +947,43 @@ namespace etl
...
+        ++smaller_itr;
+        ++larger_itr;
+      }
+
+      typename ivector<T>::iterator erase_start = larger_itr;
+
+      while(larger_itr < larger.end())
+      {
+        smaller.push_back(etl::move(*larger_itr));
```
jwellbelove: Replace with?
`etl::move(larger_itr, larger.end(), etl::back_inserter(smaller));`

include/etl/vector.h
mike919192: Yes, that cleans things up.  Thanks

include/etl/vector.h
mike919192: This is also a good cleanup.  THanks

mike919192:
@coderabbitai review

PR #1257, opened 2026-01-06
Fix year_month arithmetic and correct chrono API behavior

bolry:
Additional tests are added to cover year_month arithmetic after defects were
identified. These tests reveal incorrect behavior in the existing chrono
implementation.

The public chrono API is corrected accordingly to better match the behavior of
the C++ standard chrono library as documented on cppreference.com.

Implicit signed/unsigned conversions for year, month, and day remain unchanged.
These conversions should ideally be made explicit, and comments have been added
to document this. The change itself is deferred because the current test suite
depends on the existing implicit conversions, and modifying them is considered
out of scope for this PR.


include/etl/private/chrono/day.h
```diff
@@ -48,48 +48,18 @@ namespace etl
       //***********************************************************************
       /// Default constructor
       //***********************************************************************
-      ETL_CONSTEXPR day() ETL_NOEXCEPT
-        : value(0)
+      day() ETL_NOEXCEPT
       {
       }
```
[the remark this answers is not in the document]
bolry: C++ standard says `Default constructor leaves the day value uninitialized.`

include/etl/private/chrono/month.h
```diff
@@ -59,8 +59,7 @@ namespace etl
       //***********************************************************************
       /// Default constructor
       //***********************************************************************
-      ETL_CONSTEXPR month() ETL_NOEXCEPT
-        : value(0)
+      month() ETL_NOEXCEPT
       {
       }
```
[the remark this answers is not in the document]
bolry: C++ standard says, `Default constructor leaves the month value uninitialized.`

include/etl/private/chrono/weekday.h
```diff
@@ -347,9 +347,7 @@ namespace etl
       //***********************************************************************
       /// Default constructor
       //***********************************************************************
-      ETL_CONSTEXPR weekday_indexed() ETL_NOEXCEPT
-        : wd()
-        , i()
+      weekday_indexed() ETL_NOEXCEPT
       {
       }
```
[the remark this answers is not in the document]
bolry: C++ standard say, `Default constructor leaves the weekday value uninitialized.`

include/etl/private/chrono/year_month_weekday.h
```diff
@@ -46,10 +46,7 @@ namespace etl
       //*************************************************************************
       /// Default constructor.
       //*************************************************************************
-      ETL_CONSTEXPR year_month_weekday()
-        : y()
-        , m()
-        , wdi()
+      year_month_weekday()
       {
       }
```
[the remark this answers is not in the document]
bolry: C++ standard says, `Default constructor leaves the fields uninitialized.`

include/etl/private/chrono/year.h
```diff
@@ -48,37 +48,18 @@ namespace etl
       //***********************************************************************
       /// Default constructor
       //***********************************************************************
-      ETL_CONSTEXPR year() ETL_NOEXCEPT
-        : value(0)
+      year() ETL_NOEXCEPT
       {
       }
```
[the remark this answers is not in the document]
bolry: C++ standard says, `Default constructor leaves the year value uninitialized.`

include/etl/private/chrono/month.h
```diff
@@ -59,8 +59,7 @@ namespace etl
       //***********************************************************************
       /// Default constructor
       //***********************************************************************
-      ETL_CONSTEXPR month() ETL_NOEXCEPT
-        : value(0)
+      month()
       {
       }
```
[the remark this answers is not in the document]
bolry: C++ standard says, `Default constructor leaves the month value uninitialized.`

include/etl/private/chrono/year_month.h
```diff
@@ -171,7 +174,9 @@ namespace etl
     inline ETL_CONSTEXPR14 etl::chrono::months operator -(const etl::chrono::year_month& ym1,
                                                           const etl::chrono::year_month& ym2) ETL_NOEXCEPT
     {
-      return etl::chrono::months(static_cast<int>(((int(ym1.year()) - int(ym2.year())) * 12) + (unsigned(ym1.month()) - unsigned(ym2.month()))));
+      return etl::chrono::months(
+        (ym1.year() - ym2.year()) + etl::chrono::months(
+          static_cast<unsigned>(ym1.month()) - static_cast<unsigned>(ym2.month())));
     }
```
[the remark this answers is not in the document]
bolry: True, but the creation of months is always singed so the unsigned is converted back to int, thus we get the intended value anyway.

If still desired, we can still cast that unsigned values again to int to make this more explicit to the reader.

include/etl/private/chrono/day.h
```diff
@@ -48,48 +48,18 @@ namespace etl
       //***********************************************************************
       /// Default constructor
       //***********************************************************************
-      ETL_CONSTEXPR day() ETL_NOEXCEPT
-        : value(0)
+      day()
```
bolry: Is `= default;` allowed instead or is C++98 comparability needed for chrono?

include/etl/private/chrono/day.h
```diff
@@ -48,48 +48,18 @@ namespace etl
...
       ETL_CONSTEXPR explicit day(unsigned value_) ETL_NOEXCEPT
-        : value(static_cast<unsigned char>(value_))
+        : value(static_cast<rep>(value_))
       {
       }
 
-      //***********************************************************************
-      /// Copy constructor
-      //***********************************************************************
-      ETL_CONSTEXPR14 day(const etl::chrono::day& other) ETL_NOEXCEPT
```
bolry, line 67: Removed due to compiler generated implicit copy-constructor is more reliable.

include/etl/private/chrono/day.h
```diff
@@ -48,48 +48,18 @@ namespace etl
...
-        : value(other.value)
-      {
-      }
-
-      //***********************************************************************
-      /// Assignment operator
-      //***********************************************************************
-      ETL_CONSTEXPR14 etl::chrono::day& operator =(const etl::chrono::day& rhs) ETL_NOEXCEPT
```
bolry: Compiler generated implicit assignment operator trumps home grown implementation.

include/etl/private/chrono/day.h
```diff
@@ -184,24 +154,6 @@ namespace etl
         return 0;
       }
 
-      //***********************************************************************
-      /// The minimum day value for which ok() will return <b>true</b>
-      //***********************************************************************
-      ETL_NODISCARD
-      static ETL_CONSTEXPR14 etl::chrono::day min() ETL_NOEXCEPT
```
bolry: No `min()` in the standard for `day`.

include/etl/private/chrono/day.h
```diff
@@ -184,24 +154,6 @@ namespace etl
...
-      {
-        return etl::chrono::day(1);
-      }
-
-      //***********************************************************************
-      /// The maximum day value for which ok() will return <b>true</b>
-      //***********************************************************************
-      ETL_NODISCARD
-      static ETL_CONSTEXPR14 etl::chrono::day max() ETL_NOEXCEPT
```
bolry: No `max()` in the standard for `day`.

include/etl/private/chrono/month.h
```diff
@@ -72,24 +71,6 @@ namespace etl
...
 
-      //***********************************************************************
-      /// Copy constructor
-      //***********************************************************************
-      ETL_CONSTEXPR14 month(const etl::chrono::month& other) ETL_NOEXCEPT
```
bolry: Relying on compiler generated copy constructor.

include/etl/private/chrono/month.h
```diff
@@ -331,10 +294,8 @@ namespace etl
         etl::chrono::months ms(difference);
 
         // Check for validity.
-        if (m1 == (m2 + ms))
-        {
-          return ms;
-        }
+        assert(m1 == (m2 + ms));
```
bolry: We trust our implementation to do the right thing. This is just a last check for this.

include/etl/private/chrono/year.h
```diff
@@ -48,37 +48,18 @@ namespace etl
...
-      ETL_CONSTEXPR explicit year(unsigned value_) ETL_NOEXCEPT
+      ETL_CONSTEXPR explicit year(int value_) ETL_NOEXCEPT
         : value(value_)
       {
       }
 
-      //***********************************************************************
-      /// Copy constructor
-      //***********************************************************************
-      ETL_CONSTEXPR14 year(const etl::chrono::year& other) ETL_NOEXCEPT
```
bolry: Use compiler version.

include/etl/private/chrono/year_month_day.h
```diff
@@ -76,6 +73,9 @@ namespace etl
       /// Construct from sys_days.
       //*************************************************************************
       ETL_CONSTEXPR14 year_month_day(const etl::chrono::sys_days& sd) ETL_NOEXCEPT
+        : y(0)
+        , m(0)
```
bolry: This and day should be `0U`. Any takers?

include/etl/private/chrono/day.h
rolandreichweinbmw: In `chrono.h`, ETL_USING_CPP11 is defined globally. So C++11's `= default` should be valid.

include/etl/private/chrono/month.h
rolandreichweinbmw, line 298: Are you sure you want to use assert() here instead of ETL_ASSERT()?

include/etl/private/chrono/year.h
```diff
@@ -94,7 +75,7 @@ namespace etl
       //***********************************************************************
       ETL_CONSTEXPR14 etl::chrono::year operator ++(int) ETL_NOEXCEPT
       {
-        const etl::chrono::year temp = *this;
+        etl::chrono::year temp = *this;
```
rolandreichweinbmw, line 79: What is the reasoning behind removing const here, if you don't intend to change it before the return? (The PR's descriptions just vaguely says "defects were identified".)

include/etl/private/chrono/month.h
bolry: Yes, we don't want to throw an exception or whatever the ETL_ASSERT is configured for. But I'm open for suggestions.

jwellbelove:
I was aware that the STL version of the chrono classes default constructors do not initialise the member values, but, as the ETL is attempting to become more acceptable to be used in safety certified projects, due to requests from users, I made the decision to make the default constructed values well defined.

For example:
```
std::chrono::month m;
bool isOk = m.ok(); // Maybe true, maybe false!
```

```
etl::chrono::month m;
bool isOk = m.ok(); // Always false!
```

Not using `etl::chrono::month m{};` is an easy mistake to make, which may or may not be picked up by the compiler.

The ETL was designed to diverge from the STL where necessary, rather than being a strict clone.

bolry:
Thank you @jwellbelove for your explanation. I'll try and restore the original intended construction functionality in my PR. The main point is that etl::chrono calculation are correct. The constructors was something I noticed in passing.

jwellbelove:
Yes, thanks for sorting out the calculations. I completely missed adding tests for rollover!

jwellbelove:
After I've accepted the PR I'll merge in the changes to the tests I've done here to enable the `explicit` for the conversion operators.

bolry:
@jwellbelove Seems I've messed up a rebase against branch master. I do believe the calendar's default constructors are restored. Hope you can make headway of the mess I've currently created.

jwellbelove:
Although `master` and `development` seem to be at the same commit on `origin`, I thought I'd try to rebase the PR on `development` to see if it would resolve itself, and it seems to have worked! I'm not entirely sure why, but that's a question for a Git Guru.

PR #1261, opened 2026-01-14
Fix incorrect namespace for iter_swap calls

jwellbelove:
itr_swap should not need to be called using ETL_OR_STD::
Changed ETL_OR_STD:: to etl::

### DIFF
--- a/include/etl/algorithm.h
+++ b/include/etl/algorithm.h
@@ -186,7 +186,7 @@ namespace etl
   {
     while (first1 != last1)
     {
-      iter_swap(first1, first2);
+      etl::iter_swap(first1, first2);
       ++first1;
       ++first2;
     }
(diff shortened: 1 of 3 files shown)


rolandreichweinbmw:
How does the bug to be fixed surface? It is not obvious from the proposed changes or PR/commit message.

jwellbelove:
I discovered the issue when compiling with VS2022 with a No-STL configuration. It works by chance for GCC and clang because they happen to define `iter_swap` in one of the headers included by the unit tests.

jwellbelove:
Also, as `iter_swap` is not overloaded like `swap` is, and is not looked up through ADL, there is no clash with `std::iter_swap` and `etl::iter_swap` existing at the same time.

rolandreichweinbmw (approved):

PR #1262, opened 2026-01-14
Document how to implement platform specifics

rolandreichweinbmw:
Some interfaces need to be implemented in every project or platform using the ETL:

* etl_get_high_resolution_clock
* etl_get_system_clock
* etl_get_steady_clock
* etl_putchar

### DIFF
--- a/include/etl/algorithm.h
+++ b/include/etl/algorithm.h
@@ -186,7 +186,7 @@ namespace etl
   {
     while (first1 != last1)
     {
-      iter_swap(first1, first2);
+      etl::iter_swap(first1, first2);
       ++first1;
       ++first2;
     }
@@ -338,7 +338,9 @@ namespace etl
   template <typename TIterator1, typename TIterator2>
   ETL_CONSTEXPR20 TIterator2 move_backward(TIterator1 sb, TIterator1 se, TIterator2 de)
   {
+#include "etl/private/diagnostic_array_bounds_push.h"
     return std::move_backward(sb, se, de);
+#include "etl/private/diagnostic_pop.h"
   }
 #elif ETL_USING_CPP11
   // For C++11
--- a/include/etl/memory.h
+++ b/include/etl/memory.h
@@ -382,9 +382,13 @@ namespace etl
   ///\ingroup memory
   //*****************************************************************************
   template <typename TInputIterator, typename TOutputIterator>
-  TOutputIterator  uninitialized_move(TInputIterator i_begin, TInputIterator i_end, TOutputIterator o_begin)
+  TOutputIterator uninitialized_move(TInputIterator i_begin, TInputIterator i_end, TOutputIterator o_begin)
   {
+#include "etl/private/diagnostic_array_bounds_push.h"
+#include "etl/private/diagnostic_stringop_overflow_push.h"
     return std::uninitialized_move(i_begin, i_end, o_begin);
+#include "etl/private/diagnostic_pop.h"
+#include "etl/private/diagnostic_pop.h"
   }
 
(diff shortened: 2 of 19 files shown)


jwellbelove:
You should also point out the chrono profile macro defaults.

If `ETL_CHRONO_SYSTEM_CLOCK_DURATION` is not defined it will default to `etl::chrono::nanoseconds`
If `ETL_CHRONO_SYSTEM_CLOCK_IS_STEADY` is not defined it will default to `true`
If `ETL_CHRONO_HIGH_RESOLUTION_CLOCK_DURATION` is not defined it will default to `etl::chrono::nanoseconds`
If `ETL_CHRONO_HIGH_RESOLUTION_CLOCK_IS_STEADY` is not defined it will default to `true`
If `ETL_CHRONO_STEADY_CLOCK_DURATION` is not defined it will default to `etl::chrono::nanoseconds`

rolandreichweinbmw:
> You should also point out the chrono profile macro defaults.

Thank you for the hint!

Adjusted accordingly.

mike919192:
Sorry I haven't kept up to date on the chrono development.  Can the system clock be an arbitrary time or does it need to conform to some interval (ex. 1 ns).  For example on a platform I work on, the raw system clock values are 1/(333MHz)

rolandreichweinbmw:
> Sorry I haven't kept up to date on the chrono development. Can the system clock be an arbitrary time or does it need to conform to some interval (ex. 1 ns). For example on a platform I work on, the raw system clock values are 1/(333MHz)

This highlights a typical misunderstanding.

"System clock" in std::chrono is defined as "real time wall clock", see https://en.cppreference.com/w/cpp/chrono/system_clock.html

This is different from a raw system clock in terms of CPU cycles counter as you mentioned.

I propose to stick to C++ convention here.

rolandreichweinbmw:
Regarding your question of conformance to SI units (e.g. "1ns"), chrono::duration is generally defined via ratio, which can be odd, e.g. "one_and_a_half_microsecond" which you can define individually if necessary.

mike919192:
> > Sorry I haven't kept up to date on the chrono development. Can the system clock be an arbitrary time or does it need to conform to some interval (ex. 1 ns). For example on a platform I work on, the raw system clock values are 1/(333MHz)
> 
> This highlights a typical misunderstanding.
> 
> "System clock" in std::chrono is defined as "real time wall clock", see https://en.cppreference.com/w/cpp/chrono/system_clock.html
> 
> This is different from a raw system clock in terms of CPU cycles counter as you mentioned.
> 
> I propose to stick to C++ convention here.

Thanks, I think that points me in the right direction.  Most of our time related function are dealing with the CPU counter like you mention (starts at 0 on startup) but we also have a real-time clock with some limited uses.  Its on my to do list to try out the new functions.

README.md
```diff
@@ -184,6 +184,84 @@ add_executable(foo main.cpp)
...
+### Example
+
+```
+#include <etl/chrono.h>
+#include <etl/print.h>
+
+extern "C"
+{
+
+ETL_CHRONO_HIGH_RESOLUTION_CLOCK_DURATION::rep etl_get_high_resolution_clock()
```
jwellbelove: You don't need the macros here.
`etl::high_resolution_clock::rep` should be fine, as the clock classes use the macros to define the `duration` type.

e.g.
```
using duration   = ETL_CHRONO_HIGH_RESOLUTION_CLOCK_DURATION;
using rep        = duration::rep;
```

rolandreichweinbmw: Right. Adjusted accordingly.

Issue #1263, opened 2026-01-14
Template parameter type error when compiling etl::tuple with TI's CL430 compiler


fleckerbr:
When attempting to use ETL's Finite State Machine framework, TI's CL430 compiler failed to compile my project due to the following errors:
- [`include/etl/tuple.h:562`](https://github.com/ETLCPP/etl/blob/20.45.0/include/etl/tuple.h#L562) `argument of type "<unknown-type>" is incompatible with template parameter of type "size_t"`
- [`include/etl/tuple.h:575`](https://github.com/ETLCPP/etl/blob/20.45.0/include/etl/tuple.h#L575) `argument of type "<unknown-type>" is incompatible with template parameter of type "size_t"`
- [`include/etl/tuple.h:588`](https://github.com/ETLCPP/etl/blob/20.45.0/include/etl/tuple.h#L588) `argument of type "<unknown-type>" is incompatible with template parameter of type "size_t"`
- [`include/etl/tuple.h:601`](https://github.com/ETLCPP/etl/blob/20.45.0/include/etl/tuple.h#L601) `argument of type "<unknown-type>" is incompatible with template parameter of type "size_t"`

From what I can tell, the problem is simply that the `NTypes` parameter is being assigned an invalid default value since the `number_of_types<THead, TTail...>` helper function is never called. Consequently, the fix seems to be as simple as changing the template signature on each of these lines to:
```cpp
template <typename U1, typename U2, size_t NTypes = number_of_types<THead, TTail...>(), etl::enable_if_t<NTypes == 2U, int> = 0>
```

My only reservation is that I don't understand why this issue hasn't been encountered sooner. I can reproduce this error by simply including `include/etl/tuple.h` in my project and pasting the following test case from `test/test_tuple.cpp`:
```cpp
etl::tuple<int> tp(1);
etl::tuple<int> otherTuple(tp);

assert(etl::get<0>(tp) == etl::get<0>(otherTuple));
```

If there is something I am misunderstanding about the nature of this problem, please do let me know. Otherwise, I would be more than happy to open a pull request with my proposed fix.

jwellbelove:
You have discovered incorrect function definitions that were not covered by the unit tests.
If you could add the corrections and missing unit tests in a PR, that would be great.
Thanks for discovering that.

fleckerbr:
Ah, I may have worded my concerns about the test case incorrectly. Based on my experience, the existing test cases should have caught this bug as they are currently written.

[...]

PR #1265, opened 2026-01-16
Fix etl::tuple template signature error in pair assignment operator

fleckerbr:
This fix is intended to resolve issue https://github.com/ETLCPP/etl/issues/1263 but is lacking additional test cases as I could not figure out how to reproduce the compilation error using ETL's build system or test suite.

When using TI's CL430 compiler, I can trivially reproduce this bug using test cases from the existing test suite. Therefore, I was wondering if it would be possible to merge this fix without the addition of further tests?

### DIFF
--- a/include/etl/string_utilities.h
+++ b/include/etl/string_utilities.h
@@ -782,9 +782,9 @@ namespace etl
     etl::optional<string_view_t> token;
 
     size_t count = 0;
-    while ((count != output.max_size()) &&
-           (count != max_n_tokens) &&
-           (token = etl::get_token(input, delimiters, token, ignore_empty_tokens)))
+    while ((token = etl::get_token(input, delimiters, token, ignore_empty_tokens)) &&
+           (count != output.max_size()) &&
+           (count != max_n_tokens))
     {
       output.push_back(token.value());
       ++count;
--- a/test/test_string_utilities.cpp
+++ b/test/test_string_utilities.cpp
@@ -1116,6 +1116,48 @@ namespace
     }
 
     //*************************************************************************
+    //before the issue fix, 
+    //the return value was false when output vector max size equaled the number of tokens
+    //it should be true
+    TEST(test_issue_1270_output_vector_max_size_equals_num_tokens)
+    {
+      String text(STR(",,,The,cat,sat,,on,the,mat"));
+      etl::vector<StringView, 6> views;
+
+      bool all_views_found = etl::get_token_list(text, views, STR(","), true);
+
+      CHECK_TRUE(all_views_found);
+      CHECK_EQUAL(6, views.size());
+      CHECK_EQUAL(std::string("The"), std::string(views[0].begin(), views[0].end()));
+      CHECK_EQUAL(std::string("cat"), std::string(views[1].begin(), views[1].end()));
+      CHECK_EQUAL(std::string("sat"), std::string(views[2].begin(), views[2].end()));
+      CHECK_EQUAL(std::string("on"),  std::string(views[3].begin(), views[3].end()));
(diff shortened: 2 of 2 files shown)


jwellbelove:
I've discovered that on the unit tests on my machine, an assignment from a `pair`, which should call `operator=()`, is in fact calling one of the constructors instead. I've done some experiments to see why, but I haven't found a solution yet.

jwellbelove:
Try this patch on your PR.
[Uploading 1265-etl-tuple.patch…]()

fleckerbr:
@jwellbelove The patch you referenced seems to just be a link to this pull-request. Did your file upload fail?

jwellbelove:
[1265-etl-tuple.patch](https://github.com/user-attachments/files/24722608/1265-etl-tuple.patch)

Note: This patch includes additional unit tests.

fleckerbr:
The patch looks good to me. Tests are failing when the template signature defines `NTypes` as `size_t NTypes = number_of_types<THead, TTail...>` and are passing when updated to `size_t NTypes = number_of_types<THead, TTail...>()`.

If you don't mind, would you be willing to explain how you discovered that an assignment from `pair` was calling a constructor instead of `operator=()` and how you fixed it?

I've never needed to troubleshoot a problem like this before and would love to understand how to identify and approach it.

jwellbelove:
I ran the unit tests with the debugger with a new 'assignment' test. When I looked at the assignment operator I realised that the parameter was a plain 'pair' and not qualified with a namespace, and the compiler was ignoring it. Adding ETL_OR_STD fixed the problem.

fleckerbr:
Ah, that makes sense though I'm somewhat surprised that the compiler would ignore an unqualified name instead of issuing a warning or an error of some sort.

At this point is there anything else you need from me before this patch gets merged?

jwellbelove:
No, I think that's ready to merge.

include/etl/tuple.h
```diff
@@ -598,12 +598,12 @@ namespace etl
     //*********************************
     /// Assign from const rvalue pair tuple type.
     //*********************************
-    template <typename U1, typename U2, size_t NTypes = number_of_types<THead, TTail...>, etl::enable_if_t<NTypes == 2U, int> = 0>
+    template <typename U1, typename U2, size_t NTypes = number_of_types<THead, TTail...>(), etl::enable_if_t<NTypes == 2U, int> = 0>
     ETL_CONSTEXPR14
-      tuple& operator =(const pair<U1, U2>&& p)
+    tuple& operator =(const ETL_OR_STD::pair<U1, U2>&& p)
```
serges147, line 603: @jwellbelove  I have questions:

1) Why `p` is const rvalue (aka `&&`)?
I believe it should be just `const pair<U1, U2>&` lvalue (as [operator#9](https://en.cppreference.com/w/cpp/utility/tuple/operator=.html) at cppreferece).

2) IMHO this overload should participate in overload resolution only if `first` and `second` are `std::is_assignable` (see the same operataror#9). Is it (lack of various overload enablers) by ETL design, or more like oversight. I see similar theme at `etl::optional` as well.

serges147: Ok, now I see that there is indeed `tuple& operator =(const pair<U1, U2>& p)` at line 577. But my questions stand in general, maybe the 1st one now is more like:
1) What is the use case for `const T&&` argument (how can I trigger this overload)?... I don't see such thing at c++ standard, so I'm wondering why ETL needs it.

Also, putting breakpoints at these `operator=` methods and running "etl_tests" under debugger never hits them - feels like this code is not covered by unit tests. I wish it does (so I saw answer to my question # 1).

PR #1268, opened 2026-01-18
Remove some UB in test_vector_non_trivial.cpp

bolry:
Some of the tests' UB are detectable by Gcc15 and thus give a compile error due to warnings-as-error flag.

The problem that many of the tests compare iterators from different containers, mostly in the asserts, which is UB according to the standard.

### DIFF
--- a/README.md
+++ b/README.md
@@ -184,6 +184,84 @@ add_executable(foo main.cpp)
 target_link_libraries(foo PRIVATE etl::etl)
 ```
 
+## Profile definition
+
+When using ETL in a project, there is typically an `etl_profile.h` defined to
+adjust ETL to the project needs. ETL will automatically find `etl_profile.h`
+if it is available in the include path(s). If it's not available, ETL will
+work with default values.
+
+### Example
+
+```
+#ifndef __ETL_PROFILE_H__
+#define __ETL_PROFILE_H__
+
+#define ETL_TARGET_DEVICE_GENERIC
+#define ETL_TARGET_OS_NONE
+
+#define ETL_NO_STL
+
+#endif
+```
+
+## Platform specific implementation
+
+Although ETL is generally a self-contained header-only library, some interfaces need to be
+implemented in every project or platform, at least if those interfaces are actually being
+used, due to project specifics:
+
+| ETL header | Platform specific API to be implemented | Needed when using                   |
+|------------|-----------------------------------------|-------------------------------------|
+| `chrono.h` | `etl_get_high_resolution_clock()`       | `etl::high_resolution_clock::now()` |
+|            | `etl_get_system_clock()`                | `etl::system_clock::now()`          |
+|            | `etl_get_steady_clock()`                | `etl::steady_clock::now()`          |
+| `print.h`  | `etl_putchar()`                         | `etl::print()`                      |
+|            |                                         | `etl::println()`                    |
+
+### Example
+
+```
+#include <etl/chrono.h>
+#include <etl/print.h>
+
+extern "C"
+{
+
+etl::chrono::high_resolution_clock::rep etl_get_high_resolution_clock()
+{
+  etl::chrono::high_resolution_clock::rep(static_cast<int64_t>(getSystemTimeNs()));
+}
+
+etl::chrono::system_clock::rep etl_get_system_clock()
+{
+  return etl::chrono::system_clock::rep(static_cast<int64_t>(getSystemTimeNs()));
+}
+
+etl::chrono::system_clock::rep etl_get_steady_clock()
+{
+  return etl::chrono::system_clock::rep(static_cast<int64_t>(getSystemTimeNs()));
+}
+
+void etl_putchar(int c)
+{
+  putByteToStdout(static_cast<uint8_t>(c));
+}
+
+}
+```
+
+The following default values apply if the respective macros are not defined
+(e.g. in `etl_profile.h`):
+
+| Macro                                         | Default                    |
+|-----------------------------------------------|----------------------------|
+| `ETL_CHRONO_SYSTEM_CLOCK_DURATION`            | `etl::chrono::nanoseconds` |
+| `ETL_CHRONO_SYSTEM_CLOCK_IS_STEADY`           | `true`                     |
+| `ETL_CHRONO_HIGH_RESOLUTION_CLOCK_DURATION`   | `etl::chrono::nanoseconds` |
+| `ETL_CHRONO_HIGH_RESOLUTION_CLOCK_IS_STEADY`  | `true`                     |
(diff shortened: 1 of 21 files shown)


mike919192:
Hi @bolry .  I think that the tests were originally written as you are now proposing, but I changed them after feedback from coderabbit because modifying the iterator before begin or after end is also UB in the general case.  For the contiguous container I think both approaches end up being fine because the iterator comparison is essentially a pointer comparison, but thats not true in general.

So if both approaches are UB, maybe some sort of diagnostics push / pop is also needed?  @coderabbitai do you any further input on this?

bolry:
Hi @mike919192 . Yes, it's all true, but didn't know coderabbitai hinted at the UB to leave my current proposal close to the "original". I still believe that writing UB code not really seen by the compiler is to be preferred over UB code that made it harder to understand the what the test was originally designed to test. If at some point this can't be done then, `#pragma` blocking approach should be used. In either case we are in UB-land and the compiler may happily ignore the checks if needed anyway.

So, unless there's a compelling reason some of the older compiler currently used will detect the current UB, I say we don't change my proposal.

mike919192:
> So, unless there's a compelling reason some of the older compiler currently used will detect the current UB, I say we don't change my proposal.

I am agreeing your changes are clearer and a better test, because it tests right at the boundary of the comparison operation.  I am just stating that there could be a compiler warning in the future even if there is not one now.

The mechanism that I have seen in other tests to deal with this is diagnostics push / pop like this: https://github.com/ETLCPP/etl/blob/e9c2577d8e98ca9dd3a9c211cd2bb9395890978e/test/test_string_char.cpp#L1559

I don't know off the top of my head if any existing diagnostics fit this.  Just making a suggestion.

jwellbelove:
Out of interest, what were the exact warnings from GCC15?

bolry:
The warning (turned error) is `-Werror=array-bounds=`. See my "dump" below.

```
/opt/rh/gcc-toolset-15/root/usr/bin/c++ -DETL_DEBUG -I/home/kevin/CLionProjects/etl/test/../include -isystem /home/kevin/CLionProjects/etl/test/UnitTest++/.. -O3 -DNDEBUG -std=c++26 -fdiagnostics-color=always -fno-omit-frame-pointer -fno-common -pedantic-errors -Wall -Wextra -Werror -Wfloat-equal -Wshadow -Wnull-dereference -Wextra-semi -g -MD -MT test/CMakeFiles/etl_tests.dir/test_vector_non_trivial.cpp.o -MF test/CMakeFiles/etl_tests.dir/test_vector_non_trivial.cpp.o.d -o test/CMakeFiles/etl_tests.dir/test_vector_non_trivial.cpp.o -c /home/kevin/CLionProjects/etl/test/test_vector_non_trivial.cpp
In file included from /opt/rh/gcc-toolset-15/root/usr/include/c++/15/string:56,
                 from /opt/rh/gcc-toolset-15/root/usr/include/c++/15/bits/locale_classes.h:42,
                 from /opt/rh/gcc-toolset-15/root/usr/include/c++/15/bits/ios_base.h:43,
                 from /opt/rh/gcc-toolset-15/root/usr/include/c++/15/ios:46,
                 from /opt/rh/gcc-toolset-15/root/usr/include/c++/15/istream:42,
                 from /opt/rh/gcc-toolset-15/root/usr/include/c++/15/sstream:42,
                 from /home/kevin/CLionProjects/etl/test/UnitTest++/MemoryOutStream.h:9,
                 from /home/kevin/CLionProjects/etl/test/UnitTest++/ExecuteTest.h:8,
                 from /home/kevin/CLionProjects/etl/test/UnitTest++/TestMacros.h:7,
                 from /home/kevin/CLionProjects/etl/test/UnitTest++/UnitTestPP.h:5,
                 from /home/kevin/CLionProjects/etl/test/UnitTest++/UnitTest++.h:1,
                 from /home/kevin/CLionProjects/etl/test/unit_test_framework.h:33,
                 from /home/kevin/CLionProjects/etl/test/test_vector_non_trivial.cpp:29:
In member function ‘constexpr std::__cxx11::basic_string<_CharT, _Traits, _Alloc>::pointer std::__cxx11::basic_string<_CharT, _Traits, _Alloc>::_M_data() const [with _CharT = char; _Traits = std::char_traits<char>; _Alloc = std::allocator<char>]’,
    inlined from ‘constexpr std::__cxx11::basic_string<_CharT, _Traits, _Alloc>::basic_string(const std::__cxx11::basic_string<_CharT, _Traits, _Alloc>&) [with _CharT = char; _Traits = std::char_traits<char>; _Alloc = std::allocator<char>]’ at /opt/rh/gcc-toolset-15/root/usr/include/c++/15/bits/basic_string.h:617:34,
    inlined from ‘TestDataNDC<T>::TestDataNDC(const TestDataNDC<T>&) [with T = std::__cxx11::basic_string<char>]’ at /home/kevin/CLionProjects/etl/test/data.h:132:3,
    inlined from ‘void etl::create_copy_at(T*, const T&) [with T = TestDataNDC<std::__cxx11::basic_string<char> >]’ at /home/kevin/CLionProjects/etl/test/../include/etl/memory.h:1828:5,
    inlined from ‘void etl::ivector<T>::create_back(const_reference) [with T = TestDataNDC<std::__cxx11::basic_string<char> >]’ at /home/kevin/CLionProjects/etl/test/../include/etl/vector.h:1115:26,
    inlined from ‘T* etl::ivector<T>::emplace(const_iterator, Args&& ...) [with Args = {const std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >&}; T = TestDataNDC<std::__cxx11::basic_string<char> >]’ at /home/kevin/CLionProjects/etl/test/../include/etl/vector.h:648:20,
    inlined from ‘virtual void {anonymous}::Suitetest_vector_non_trivial::Testtest_emplace_out_of_range::RunImpl() const’ at /home/kevin/CLionProjects/etl/test/test_vector_non_trivial.cpp:1017:7:
/opt/rh/gcc-toolset-15/root/usr/include/c++/15/bits/basic_string.h:239:28: error: array subscript 0 is outside array bounds of ‘{anonymous}::Suitetest_vector_non_trivial::DataNDC [1]’ {aka ‘etl::vector<TestDataNDC<std::__cxx11::basic_string<char> >, 10> [1]’} [-Werror=array-bounds=]
  239 |       { return _M_dataplus._M_p; }
      |                            ^~~~
/home/kevin/CLionProjects/etl/test/test_vector_non_trivial.cpp: In member function ‘virtual void {anonymous}::Suitetest_vector_non_trivial::Testtest_emplace_out_of_range::RunImpl() const’:
/home/kevin/CLionProjects/etl/test/test_vector_non_trivial.cpp:1013:15: note: at offset -8 into object ‘data’ of size 512
 1013 |       DataNDC data;
      |               ^~~~
```

drewr95 (approved):

PR #1269, opened 2026-01-19
Deduce underlying storage size when constructing string_ext from char[].

denravonska:
This removes the need for passing sizeof(storage) to the constructor.


jwellbelove:
You will need to replicate this for:-
u8string
u16string
u32string
wstring
(duplicated this as a review change)

denravonska:
> You will need to replicate this for:- u8string u16string u32string wstring

Wait, did I miss anything?

Edit: NVM, saw the duplication

jwellbelove:
No, sorry, I meant the `_ext` versions, which you've done.

test/test_string_char_external_buffer.cpp
```diff
@@ -120,6 +120,20 @@ namespace
       CHECK_FALSE(text.is_truncated());
     }
 
+    //*************************************************************************
+    TEST_FIXTURE(SetupFixture, test_constructor_from_array)
+    {
+      value_t buffer[SIZE + 1];
+      Text text(buffer);
+
+      CHECK_EQUAL(text.size(), size_t(0));
```
jwellbelove: These values are reversed.
Should be (Expected, Actual)

denravonska:
Should there also be constructors that mimic the current behavior of char pointers where you pass in the same pointers and it retains the strings that are currently stored?

jwellbelove:
> Should there also be constructors that mimic the current behavior of char pointers where you pass in the same pointers and it retains the strings that are currently stored?

My original review comment was incorrect; Any text in the buffer will normally be lost in the simple buffer allocation constructor.

But you are correct, there should be additional constructors that mimic the behaviour of those that initialise the strings from a parameter. (In fact, it looks to be eight constructors in total).

denravonska:
> > Should there also be constructors that mimic the current behavior of char pointers where you pass in the same pointers and it retains the strings that are currently stored?
> 
> My original review comment was incorrect; Any text in the buffer will normally be lost in the simple buffer allocation constructor.
> 
> But you are correct, there should be additional constructors that mimic the behaviour of those that initialise the strings from a parameter. (In fact, it looks to be eight constructors in total).

Looking at the constructors now. Why are there constructors for both `istring` and `string_ext`?

```Cpp
    /// Copy constructor.
    ///\param other The other string_ext.
    string_ext(const etl::string_ext& other, value_type* buffer, size_type buffer_size)
      : istring(buffer, buffer_size - 1U)
    {
      if (this->is_within_buffer(other.data()))
      {
        this->current_size = other.size();
      }
      else
      {
        this->initialise();
        this->assign(other);
      }
    }

    /// From other istring.
    ///\param other The other istring.
    string_ext(const etl::istring& other, value_type* buffer, size_type buffer_size)
      : istring(buffer, buffer_size - 1U)
    {
      if (this->is_within_buffer(other.data()))
      {
        this->current_size = other.size();
      }
      else
      {
        this->initialise();
        this->assign(other);
      }
    }
```

jwellbelove:
For constructing an `etl::string_ext` from any `etl::string<N>`.

denravonska:
Yeah but both `string_ext` and `string` are `istrings`. Shouldn't the `istring` constructor cover both cases?

jwellbelove:
I'm not sure if the compiler sees the `istring` version as a valid copy constructor. It's a long time since I wrote `string.h`, so I'm not sure.

test/test_string_u16_external_buffer.cpp
```diff
@@ -134,6 +134,20 @@ namespace
       CHECK_FALSE(text.is_truncated());
     }
 
+    //*************************************************************************
+    TEST_FIXTURE(SetupFixture, test_constructor_from_array)
+    {
+      value_t buffer[SIZE + 1];
+      Text text(buffer);
+
+      CHECK_EQUAL(text.size(), size_t(0));
```
jwellbelove: These values are reversed.
Should be (Expected, Actual)

test/test_string_u32_external_buffer.cpp
```diff
@@ -134,6 +134,20 @@ namespace
       CHECK_FALSE(text.is_truncated());
     }
 
+    //*************************************************************************
+    TEST_FIXTURE(SetupFixture, test_constructor_from_array)
+    {
+      value_t buffer[SIZE + 1];
+      Text text(buffer);
+
+      CHECK_EQUAL(text.size(), size_t(0));
```
jwellbelove: These values are reversed.
Should be (Expected, Actual)

test/test_string_u8_external_buffer.cpp
```diff
@@ -137,6 +137,20 @@ namespace
       CHECK_FALSE(text.is_truncated());
     }
 
+    //*************************************************************************
+    TEST_FIXTURE(SetupFixture, test_constructor_from_array)
+    {
+      value_t buffer[SIZE + 1];
+      Text text(buffer);
+
+      CHECK_EQUAL(text.size(), size_t(0));
```
jwellbelove: These values are reversed.
Should be (Expected, Actual)

test/test_string_wchar_t_external_buffer.cpp
```diff
@@ -137,6 +137,20 @@ namespace
       CHECK_FALSE(text.is_truncated());
     }
 
+    //*************************************************************************
+    TEST_FIXTURE(SetupFixture, test_constructor_from_array)
+    {
+      value_t buffer[SIZE + 1];
+      Text text(buffer);
+
+      CHECK_EQUAL(text.size(), size_t(0));
```
jwellbelove: These values are reversed.
Should be (Expected, Actual)

jwellbelove (changes requested):
You need to swap the Expected/Actual parameters in 
`CHECK_EQUAL(text.size(), size_t(0));`
to
`CHECK_EQUAL(size_t(0), text.size());`
in all five of the unit tests.

rolandreichweinbmw (approved):

include/etl/string.h
```diff
@@ -383,6 +438,27 @@ namespace etl
...
+    ///\param buffer The array buffer.
+    //*************************************************************************
+    template <typename TPointer, size_t BufferSize>
+    string_ext(TPointer text, value_type (&buffer)[BufferSize],
+               typename etl::enable_if<etl::is_same<const value_type*, TPointer>::value, int>::type* = ETL_NULLPTR)
+      : istring(buffer, BufferSize - 1U)
+    {
+      if (this->is_within_buffer(text))
+      {
+        this->current_size = etl::strlen(buffer);
```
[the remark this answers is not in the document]
denravonska: @copilot Does that mean that the existing code is also incorrect?
```Cpp
    template <typename TPointer>
    string_ext(TPointer text, value_type* buffer, size_type buffer_size,
               typename etl::enable_if<etl::is_same<const value_type*, TPointer>::value, int>::type* = ETL_NULLPTR)
      : istring(buffer, buffer_size - 1U)
    {
      if (this->is_within_buffer(text))
      {
        this->current_size = etl::strlen(buffer);
      }
```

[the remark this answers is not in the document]
denravonska: Ping, @jwellbelove ^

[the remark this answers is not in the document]
jwellbelove: If the text pointer is within the buffer, then the buffer contains the text, so `strlen(buffer)` should be fine.
There can be strange effects if text pointer is within the buffer, but not at the start of the buffer.
Maybe this requires an extra assert check?

Strange effect example
```
char text[12];
strcpy(text, "Hello World");
etl::string_ext s(text + 4, text, 12); // Size = 7, buffer = "Hello World", size = 7
```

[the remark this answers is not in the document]
jwellbelove: I think the 'strange effect' should be dealt within another issue.
Maybe rejecting the text pointer if it is within the buffer, but not at the start.

Also adding a new constructor with an `etl::in_place_t` parameter when the buffer contains the initial string?

`string_ext(etl::in_place_t, value_type* buffer, size_type buffer_size)`

Issue #1270, opened 2026-01-19
get_token_list returns false when output vector is exactly sized


mike919192:
I recently needed to try out `get_token_list` and overall its a nice API and pretty easy to use.  I noticed one strange behavior though where the return value `all_tokens_found` will be false if the output vector is sized exactly to the number of tokens.  The output vector needs to be oversized by at least 1 in order to return true.

For example:
```
String text(STR("one, two, three, four, five, six"));
etl::vector<StringView, 6> views;

bool all_views_found = etl::get_token_list(text, views, STR(", "), true);
```
Will return false even though all tokens are read and the vector is full.

Looking at the implementation, it does not look too difficult to modify `get_token_list` so that the example would return true instead.  Would this be considered a breaking change to change the return value?  Or would it be OK to make a PR to modify this.

jwellbelove:
That looks like a bug. If you have a solution then definitely raise a PR.
Don't forget to add a unit test that triggers this issue, before adding the fix.

mike919192:
PR #1271 merged into development.  Thanks

PR #1271, opened 2026-01-20
Fix return value of get_token_list

mike919192:
This PR proposes fixes to 2 related issues.

- When the output vector has capacity equal to the number of tokens that are read, the return value is false even though all the tokens are read.
- When the max number of tokens is equal to the actual number of tokens, the return value is also false even though all the tokens are read.

The reason for this is that in the while loop, `get_token` was the last expression to evaluate.  So if it is short circuited by the other expression, its never checked if the parsing has reached the end of the string.  Moving `get_token` to the first expression, means that the next token will always parse before breaking out of the while loop, so that the return value correctly indicates if the end of the string is reached.

Issue: #1270

### DIFF
--- a/include/etl/generators/type_traits_generator.h
+++ b/include/etl/generators/type_traits_generator.h
@@ -1227,11 +1227,11 @@ typedef integral_constant<bool, true>  true_type;
   /// is_pod
   ///\ingroup type_traits
   template <typename T>
-  struct is_pod : std::integral_constant<bool, std::is_standard_layout<T>::value && std::is_trivial<T>::value> {};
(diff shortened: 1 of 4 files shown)


jwellbelove (changes requested):
You will need to update the other unit tests:-

test_string_utilities_std.cpp
test_string_utilities_std_u8.cpp
test_string_utilities_std_u16.cpp
test_string_utilities_std_u32.cpp
test_string_utilities_std_wchar.cpp
test_string_utilities_u8.cpp
test_string_utilities_u16.cpp
test_string_utilities_u32.cpp

mike919192:
> You will need to update the other unit tests:-

The only file that has tests for `get_token_list` is `test_string_utilities.cpp`, the function `get_token` is tested in all the files you mentioned.  I assumed that this was because `get_token_list` does not directly read from the string.

PR #1274, opened 2026-01-21
Feature/#1234 add etl  type list api to more classes

jwellbelove:


jwellbelove:
Closed because the code was not based on master and pulled in too many other PR changes.

Issue #1276, opened 2026-01-21
Data corruption in the `etl::bip_buffer_spsc_atomic`.


serges147:
Internal `etl::bip_buffer_spsc_atomic_base::apply_write_reserve` has mistake in the "wrapping around" logic, which leads to not updated `.last` index according to the last `write_commit`. As a result, size of the next `read_reserve` could be bigger than actual data in the tail of the ring buffer, and so "garbage" data could be read from the reservation.

This issue was found in production during troubleshooting/debugging of occasional garbage characters coming out of the BIP buffer, and later was confirmed by 2 new unit tests (see corresponding PR) , as well as by running `etl::bip_buffer_spsc_atomic` API under American Fuzzy Lop (AFL) fuzzer.

serges147:
Closing as completed.

PR #1277, opened 2026-01-21
Fix for issue 1276 "Data corruption in the etl::bip_buffer_spsc_atomic"

serges147:
See issue #1276 .

- Added two new unit tests for the BIP buffer:
  - minimalistic reproduction of the case
  - random (seeded) running of write/read operations on a BIP

Both tests are expectedly failing (without the fix below).

- Added fix to the BIP buffer: `.last` member is updated even in the case of write wrapping around.

### DIFF
--- a/include/etl/bip_buffer_spsc_atomic.h
+++ b/include/etl/bip_buffer_spsc_atomic.h
@@ -281,6 +281,9 @@ namespace etl
         else
         {
           ETL_ASSERT_OR_RETURN((windex == 0) && ((wsize + 1) <= read_index), ETL_ERROR(bip_buffer_reserve_invalid));
+
+          // Correct wrapping point
+          last.store(write_index, etl::memory_order_release);
         }
         
         // Always update write index


test/test_bip_buffer_spsc_atomic.cpp
```diff
@@ -332,6 +334,123 @@ namespace
...
+        CHECK_EQUAL(2U, reader.size());
+        CHECK_EQUAL('9', reader[0]);
+        CHECK_EQUAL('A', reader[1]);
+      }
+    }
+
+    //*************************************************************************
+    TEST(test_write_random_back_pressure)
+    {
+      // Deliberately seed with fixed number, so that test always fails without the fix.
```
rolandreichweinbmw: You don't need to refer to "the fix" in the source code. That makes sense only in a commit message or PR description.

test/test_bip_buffer_spsc_atomic.cpp
```diff
@@ -332,6 +334,123 @@ namespace
...
+      std::mt19937 mte(123);
+
+      constexpr size_t N = 256;
+      etl::bip_buffer_spsc_atomic<int, N> stream;
+      etl::ibip_buffer_spsc_atomic<int>&   istream = stream;
```
rolandreichweinbmw: spacing?

test/test_bip_buffer_spsc_atomic.cpp
serges147: >Also, please squash...

Roland, do you mean commits 0d6e7af5 & d7408b26 ?
I'm not sure how to squash already pushed commits in the middle of history... can we squash all of them (6 currently) altogether during merging to the master?

rolandreichweinbmw: Yes, and also ec1b41b49694be5824db39a063b5c4a2876dedd9 . You can selectively and interactively squash commits with "git rebase -i" and then force push to a not yet merged PR.

But you are right: Squashing into one single commit at merge would be an option also.

rolandreichweinbmw (approved):

serges147:
> Also, please squash the commits inside the PR where reasonable, so they are not correcting each other anymore.

@rolandreichweinbmw I rebased commits to just 2 : reproduce & fix.

jwellbelove (approved):

Issue #1258, opened 2026-01-07
New feature: Manchester coding


tzijnge:
Hi @jwellbelove,

I've been working on an efficient implementation of Manchester coding and decoding that might fit the ETL. I developed this code as part of my work, but with permission to make it open source, I spent some time reworking it into the ETL format. Please find my work [here](https://github.com/imec-int/etl/pull/1) and let me know what you think. It's not finished yet, but it's getting close.

Here's a quick explanation of how it works

``` C++
/*
 * 1. Manchester coded value of 0b11001100 is equal to the XOR of that value with 0xAAAA:
 *    0b1010101010101010 and 0b1111000011110000, where the latter value has
 *    duplicated every bit of the input
 * 2. Bit duplication is achieved as follows (example for )
 *
 *    +---+-----------------+-----------------+     ||
 *    | 0 | _ _ _ _ _ _ _ _ | A B C D E F G H |     || input value (i)
 *    +---+-----------------+-----------------+     ||
 *    | 1 | _ _ _ _ A B C D | _ _ _ _ E F G H |     || (i | (i << 4)) & 0x0F0F
 *    +---+-----------------+-----------------+     ||
 *    | 2 | _ _ A B _ _ C D | _ _ E F _ _ G H |     || (i | (i << 2)) & 0x5555
 *    +---+-----------------+-----------------+     ||
 *    | 3 | _ A _ B _ C _ D | _ A _ B _ C _ D |     || (i | (i << 1)) & 0x0F0F
 *    +---+-----------------+-----------------+     ||
 *    | 4 | A A B B C C D D | A A B B C C D D |     || (i | (i << 1))
 *    +---+-----------------+-----------------+     ||
 *
 * 3. Manchester decoding is done in a similar, but reversed way
 * 4. Error detection during Manchester decoding is comparing 2 neighboring bits. If they are
 *    equal, then an there is an error in the encoded input data. Comparing all 8 bit-pairs in a 16-bit
 *    word is done as follows (example for 8 bits):
 *    11011000      || First bit pair (lsb, 00) is invalid. Last bit pair is also invalid. Other bit pairs are valid
 *    01101100      || shift right by one bit
 *    10110100      || XOR
 *    01010101      || Mask with 0x55
 *    00010100      || If result is not equal to 0x55, there was an error in the input
 */
```

[...]

PR #1285, opened 2026-01-28
#1258 Manchester coding

tzijnge:
As discussed in #1258, here is the code that adds Manchester coding and decoding to the ETL.

@jwellbelove I think this PR needs some documentation. I'd be happy to write it, but I'm not sure where and how. I added basic Doxygen documentation to the source code, but a more detailed explanation of the algorithm would be useful. Do I put this in the source code itself? The documentation on https://www.etlcpp.com/ should also be updated, but I don't understand how that is done. Please let me know your thoughts on this.

### DIFF
--- a/include/etl/manchester.h
+++ b/include/etl/manchester.h
@@ -174,7 +174,7 @@ namespace etl
     //*************************************************************************
     /// Alias for memcpy. etl::mem_copy is not suitable for the Manchester
     /// algorithm because all memory copies are between different types. This
-    /// alias is a way to respect ETL_USING_BUILTIN_MEMCPY while using the 
+    /// alias is a way to respect ETL_USING_BUILTIN_MEMCPY while using the
     /// memcpy function signature
     //*************************************************************************
     inline void* memcpy(void* dest, const void* src, std::size_t count) ETL_NOEXCEPT
@@ -524,7 +524,7 @@ namespace etl
         uint16_t chunk{};
         if (etl::endianness::value() == etl::endian::little)
         {
-          chunk = static_cast<uint16_t>((encoded[i + 1] << CHAR_BIT | encoded[i]);
+          chunk = static_cast<uint16_t>((encoded[i + 1] << CHAR_BIT) | encoded[i]);
         }
         else
         {


include/etl/manchester.h
```diff
@@ -0,0 +1,521 @@
...
+    template <>
+    struct decoded<uint32_t>
+    {
+      typedef uint16_t type;
+    };
+    template <>
+    struct decoded<uint64_t>
+    {
+      typedef uint32_t type;
+    };
```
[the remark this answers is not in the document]
tzijnge: Done

tzijnge:
@jwellbelove Apart from documentation (for which I need your guidelines) I don't have anything to add to this branch. Please review, so the PR can be proceeded

jwellbelove:
@tzijnge The documentation is manually created at the moment in a website editor. As this has become unwieldy, given the size and complexity of the ETL, I am look at moving to automatically generated documentation using Doxygen/Breathe/Sphinx at some point, but that may be a while coming.

For the moment, can you just create a text file that lists the public API with a short description about each function, and possibly a short demonstration of code encoding and decoding some data?

rolandreichweinbmw:
> For the moment, can you just create a text file that lists the public API with a short description about each function, and possibly a short demonstration of code encoding and decoding some data?

Following common convention, I propose a MarkDown text file similar to the root README.md under doc/ .

include/etl/manchester.h
```diff
@@ -0,0 +1,537 @@
...
+
+      ETL_ASSERT(encoded.size() >= decoded.size() * 2, ETL_ERROR(manchester_invalid_size));
+      ETL_ASSERT(decoded.size() % sizeof(TDecoded) == 0, ETL_ERROR(manchester_invalid_size));
+
+      size_t dest_index = 0;
+      size_t source_index = 0;
+      for (size_t i = 0; i < decoded.size() / sizeof(TDecoded); ++i)
+      {
+        TDecoded decoded_value = 0;
+        memcpy(&decoded_value, &decoded[source_index], sizeof(TDecoded));
```
rolandreichweinbmw: You can use etl::mem_copy which automatically uses builtin or libc memcpy, depending on etl profile.

include/etl/manchester.h
```diff
@@ -0,0 +1,550 @@
...
+#if ETL_USING_64BIT_TYPES
+      static const uint64_t inversion_mask = 0xFFFFFFFFFFFFFFFFULL;
+#else
+      static const uint32_t inversion_mask = 0xFFFFFFFFUL;
+#endif
+    };
+
+    //*************************************************************************
+    /// Alias for memcpy. etl::mem_copy is not suitable for the Manchester
+    /// algorithm. This is an alternative way to respect ETL_USING_BUILTIN_MEMCPY
```
rolandreichweinbmw: What about etl::mem_copy is not suitable? I know about the different argument order, but I expect this to be solved easily?

include/etl/manchester.h
tzijnge: While I normally like using `etl::mem_copy` or `std::copy` for their type safety, I don't think it makes sense to use `etl::mem_copy` in the Manchester algorithm as the source and destination types are always different. I'd end up casting everything to the right type only to cast it to `void*` inside `etl::mem_copy`. Your remark did bring `ETL_USING_BUILTIN_MEMCPY` to my attention and I think that setting should be respected. I ended up writing an alias for `memcpy` in the `private_manchester` namespace

include/etl/manchester.h
tzijnge: It's about the source and destination types being different. Please see my response to your other comment. I'll update the in-code documentation to make this more clear

jwellbelove: It could be worth looking again at the definitions of `etl::mem_copy` to allow for the destination pointer to be a different type from the source.

```
template <typename TSrc, typename TDst>
TDst* mem_copy(const TSrc* sb, size_t n, TDst* db) ETL_NOEXCEPT
```

tzijnge:
> @tzijnge The documentation is manually created at the moment in a website editor. As this has become unwieldy, given the size and complexity of the ETL, I am look at moving to automatically generated documentation using Doxygen/Breathe/Sphinx at some point, but that may be a while coming.
> 
> For the moment, can you just create a text file that lists the public API with a short description about each function, and possibly a short demonstration of code encoding and decoding some data?

@jwellbelove Following the suggestion of @rolandreichweinbmw, I added documentation in markdown format in the docs folder in the root of the repo. I also added a basic _index.md_ and a jekyll configuration. This way it should be very easy for you to add the 'jekyll-gh-pages' action that is provided by GitHub. I made this work on a [personal repo](https://github.com/tzijnge/test_pages). Please have a look at the resulting pages [here](https://tzijnge.github.io/test_pages/). This could be a first step towards a new documentation format. At least it gives you the possibility to render the page and link to it.

I see that you already merged this branch (thanks ;-)), so I'll leave it up you to decide what to do with the documentation

rolandreichweinbmw:
Thank you for adding the documentation! Since John already merged the branch before you added further commits, please create a follow-up PR to not miss the latest additions.

PR #1286, opened 2026-01-31
Add missing includes

rolandreichweinbmw:
Before this change, the includes needed to be done explicitly by files using basic_string_stream.h, and be included first. (E.g. in include/etl/u8string_stream.h)

This was error prone, especially if includes are reordered (e.g. via the currently defined clang-format rules).

### DIFF
--- a/README.md
+++ b/README.md
@@ -182,6 +182,84 @@ add_executable(foo main.cpp)
 target_link_libraries(foo PRIVATE etl::etl)
 ```
 
+## Profile definition
+
+When using ETL in a project, there is typically an `etl_profile.h` defined to
+adjust ETL to the project needs. ETL will automatically find `etl_profile.h`
+if it is available in the include path(s). If it's not available, ETL will
+work with default values.
+
+### Example
+
+```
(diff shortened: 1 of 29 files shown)


jwellbelove (approved):

jwellbelove:
I realised after pulling this PR that this change causes headers to be included, whether they are required or not.

Before, a `#include "etl/string_stream.h"` would only include `to_string.h`.
Now that all of the includes are in `basic_string_stream.h`, `#include "etl/string_stream.h"` will include `to_string.h`, `to_u8string.h`, `to_wstring.h`, `to_u16string.h`, and `to_u32string.h`.

Also, the `to_xxstring` headers are still being included in the `xxstring_stream headers`.

rolandreichweinbmw:
Right. OTOH, that's a reasonable compromise regarding the original issue raised.

I encountered it before and would like to have it solved.

Otherwise, it would need to be documented that a certain include order is necessary, and implemented in clang-format which would be difficult to overlook, debug and maintain.

PR #1288, opened 2026-02-01
Add basic_format_arg constructor for ibasic_string

mike919192:
Add `basic_format_arg` constructor that takes `etl::ibasic_string` so that etl strings can be used directly as format args without the user needing to construct a string_view.  The constructor creates the string_view so no other changes to format internals are needed.

Added tests mostly duplicated from the string_view tests.  Also added test to make sure temporary lifetime works properly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -182,6 +182,84 @@ add_executable(foo main.cpp)
 target_link_libraries(foo PRIVATE etl::etl)
 ```
 
+## Profile definition
+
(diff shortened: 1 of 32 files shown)


rolandreichweinbmw (approved):

rolandreichweinbmw:
Here is a sibling PR: https://github.com/ETLCPP/etl/pull/1290

PR #1290, opened 2026-02-02
Add support for size_t and unsigned long to etl::format

rolandreichweinbmw:


include/etl/format.h
```diff
@@ -254,6 +254,13 @@ namespace etl
     {
     }
 
+    // Additional type to list of basic types as defined for std::basic_format_arg:
+    // Mapping unsigned long to unsigned long long int
+    basic_format_arg(const unsigned long v)
```
jwellbelove, line 280: Should there be a `basic_format_arg(const unsigned long v)` too?

jwellbelove:
When I was looking at the code following my review comment, I found I had a question about this line.

```
  // supported types to format
  using supported_format_types = etl::variant<
    etl::monostate,
    bool,
    char,
    int,
    unsigned int,
    long long int,
    unsigned long long int,
    float,
    double,
    long double,
    const char*,
    etl::string_view,
    const void*
    // basic_format_arg::handle,
  >;
```
Is there any particular reason why the fundamental types
```
signed char
unsigned char
short
unsigned short
long int,
unsigned long int
```
are missing from this list?

include/etl/format.h
rolandreichweinbmw: That's actually the one above. I guess you mean a different type?

rolandreichweinbmw:
The source for the concrete list of types in https://en.cppreference.com/w/cpp/utility/format/basic_format_arg.html is the definition of std::basic_format_arg which defines a limited set of implemented basic types:

https://en.cppreference.com/w/cpp/utility/format/basic_format_arg.html

Further types can be added as converting constructors, e.g.

* https://github.com/ETLCPP/etl/pull/1288
* https://github.com/ETLCPP/etl/pull/1290

I will document this better.

rolandreichweinbmw:
I also added the type conversions for the remaining types from the list you mentioned.

drewr95 (approved):

test/test_format.cpp
```diff
@@ -131,8 +187,10 @@ namespace
     {
       etl::string<100> s;
 
-      CHECK_EQUAL("34", test_format(s, "{}", static_cast<uint8_t>(34)));
-      CHECK_EQUAL("-14", test_format(s, "{}", static_cast<int8_t>(-14)));
+      // mapped to unsigned char
```
jwellbelove, line 190: Commented out code

rolandreichweinbmw: Right - we now re-mapped uint_8 from integer number formatting to char formatting (as char). Tested in the char test instead now.

Would you rather like to have those commented lines removed, or documented here ?

PR #1291, opened 2026-02-02
Make typed_storage constructor constexpr

rolandreichweinbmw:


jwellbelove:
What is the benefit of making the constructor `constexpr`?
The only way to creating a value in the storage requires the use of placement-new, which cannot be `constexpr`.

rolandreichweinbmw:
The use case is that the placement new via the _other_ constructor is being done later at runtime, at a time chosen explicitly. (And no constexpr obviously.)

But the default constructor is still being called, over which time we don't have control. (It's done implicitly.)

By making the default / empty constructor constexpr, the default value becomes a kind of literal that will be placed into the .data section as bytes already. No constructor code being emitted, or called at runtime.

This makes it smaller, faster, and we don't need to worry anymore about the order of the constructor calls.

rolandreichweinbmw:
See the simplified version at https://godbolt.org/z/5aeEY7Kx6

The constructor is generated and called only if not constexpr.

Issue #1297, opened 2026-02-16
Intrusive `etl::tree_link` is not updating grandparent's link after `etl::link_rotate`.


serges147:
"etl::link_rotate_left/right" functions - they update most of the links according to the desired rotation but one - a child link at parent of `leaf` is not updated to point to the `leaf`. As a result, tree becomes broken after rotation, and requires manual/extra intervention to fix it.

See discussion #1296 for details.

jwellbelove:
Can you try this code?

[etl-#1296-tree-link-fix.patch](https://github.com/user-attachments/files/25339301/etl-.1296-tree-link-fix.patch)

serges147:
> Can you try this code?
> 
> [etl-#1296-tree-link-fix.patch](https://github.com/user-attachments/files/25339301/etl-.1296-tree-link-fix.patch)

Yes, it works!

Two minor issues with the patch though:
- `// \` fails with "error: multi-line comment [-Werror=comment]"
- I believe diagram at lines 1735...1741 is wrong.

serges147:
This could be closed as fixed, right?

jwellbelove:
In 20.46.0

PR #1299, opened 2026-02-16
Fix link rotations for etl::tree_node

jwellbelove:
Fixes the issue in #1297


rolandreichweinbmw (approved):
LGTM.

Ragarding the Copilot comments regarding whitespace, we can use clang-format more consistently (and adjust .clang-format where necessary). I made good experience with "treefmt" which automates this.

PR #1304, opened 2026-02-18
Fix etl::variant get() for ambiguous types

rolandreichweinbmw:
When a type occured multiple times in a variant, the comparison operator asserted before.

Consider the following use case:

```
#define ETL_THROW_EXCEPTIONS
#define ETL_VERBOSE_ERRORS
#define ETL_DEBUG_THROW_EXCEPTIONS

#include <etl/variant.h>

int main()
{
    etl::variant<int, int> t1;
    etl::variant<int, int> t2;

    t1.template emplace<1>(1);
    t2.template emplace<1>(1);

    return t1 == t2;
}
```

This crashes with e.g. `terminate called after throwing an instance of 'etl::variant_incorrect_type_exception'`. But it works with the std version. See also https://en.cppreference.com/w/cpp/utility/variant.html for reference. The same type is intentionally allowed several times, and comparison needs to compare successfully if the same index is activated.

In ETL, only the first found type was considered before.

### DIFF
--- a/include/etl/basic_string.h
+++ b/include/etl/basic_string.h
@@ -2757,10 +2757,11 @@ namespace etl
     }
 
     //*********************************************************************
-    /// Common implementation for 'assign' and 'append' for iterators.
+    /// Common implementation for 'assign' and 'append' for non-pointer iterators.
     //*********************************************************************
     template <typename TIterator>
-    void append_impl(iterator position, TIterator first, TIterator last, bool truncated, bool secure)
+    typename etl::enable_if<!etl::is_pointer<typename etl::remove_reference<TIterator>::type>::value>::type
+    append_impl(iterator position, TIterator first, TIterator last, bool truncated, bool secure)
     {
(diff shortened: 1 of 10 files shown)


christophruethingbmw (approved):

include/etl/private/variant_variadic.h
```diff
@@ -1217,15 +1223,78 @@ namespace etl
...
   };
 
+  namespace private_variant
+  {
+    //***************************************************************************
+    /// is_same_type_in.
+    /// Checks if specified type T is at specified index in given type list
+    //***************************************************************************
+    template<typename T, typename T0, typename T1, typename... Ts>
+    typename etl::enable_if<etl::is_same<T, T0>::value, bool>::type
```
jwellbelove: As this variant class is C++11 you can use `etl::enable_if_t<etl::is_same<T, T0>::value, bool>`

rolandreichweinbmw: Adjusted accordingly.

PR #1308, opened 2026-02-20
Fix comparison and swap operators not in etl namespace

### LINKED ISSUE #1282: Comparison operators for expected are not in the etl namespace
Google test isn't able to find the comparison operator for etl::expected.

In `expected.h` the operators aren't put in the `etl` namespace, so maybe that is the reason?

HueCodes:
Fixes #1282.

The `operator==`, `operator!=`, and `swap` overloads for `etl::expected` and `etl::unexpected` were defined outside `namespace etl`, breaking ADL. Test frameworks like Google Test use ADL to locate these operators, so comparisons using `EXPECT_EQ` would fail to compile.

Moved the operators inside the `namespace etl` block and removed the now-redundant `etl::` qualification from type names within that block.

### DIFF
--- a/include/etl/expected.h
+++ b/include/etl/expected.h
@@ -1296,14 +1296,13 @@ namespace etl
     }
 #endif
   };
-}
 
 //*******************************************
 /// Equivalence operators.
 //*******************************************
 template <typename TValue, typename TError, typename TValue2, typename TError2>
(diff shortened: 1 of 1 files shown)


jwellbelove:
This was addressed in PR #1287.
It in in development awaiting the next release (very soon).

PR #1309, opened 2026-02-20
Reduce string flash usage

denravonska:
Measured on ARM Cortex-M33 (EFR32FG28) at `-Os` with `-ffunction-sections`/`--gc-sections`. Measuring the impact of individual changes was a bit tricky since the compiler could do its magic in trivial tests. The changes combined was 3,012 B in our 550 kB project.

Posting a PR so I can let the CI build under more configurations and with more compilers.

# Remove redundant initialise() in constructors

Every `string<N>` and `string_ext` constructor called `initialise()` before immediately calling `assign()`, which overwrites the same variables `initialise()` just cleared. Removing the redundant call eliminates the inlined `initialise()` body at each call site.

One reason why this change might be unwanted is if more initialization is added to `initialise()` in the future and if `assign()` depends on the additions being pre-setup. Hopefully future tests are able to catch this.

# Avoid templated assign(iter, iter) with C strings

`string_ext` constructors ~~and `make_string`~~ called the templated `assign(text, text + strlen(text))` or constructed with `string<N>(text, strlen(text, N-1))` instead of the non-template `assign(text)` overload. The templated path pulls in the full iterator-based `append_impl` and `copy_characters template` machinery. The non-template path uses `mem_move` directly.

# Dispatch pointer iterators to non-template `append_impl` core

When `append_impl` is called with pointer iterators (e.g. `begin()`/`end()`), SFINAE now routes them to a non-template core that takes a pointer and length. This avoids duplicating the function body (truncation checks, secure clear, cleanupetc ) for each pointer type instantiation.

### DIFF
--- a/include/etl/type_list.h
+++ b/include/etl/type_list.h
@@ -48,11 +48,14 @@ namespace etl
   static ETL_CONSTANT size_t type_list_npos = etl::integral_limits<size_t>::max;
 
   //***************************************************************************
-  /// Type list forward declaration.
+  // Type list forward declaration.
   //***************************************************************************
   template <typename... TTypes>
   struct type_list;
 
+  //***************************************************************************
+  /// Check if a type is an etl::type_list.
+  //***************************************************************************
   template <typename T>
   struct is_type_list : etl::false_type {};
 
@@ -76,6 +79,7 @@ namespace etl
 
   private:
 
+    // A type_list cannot be instantiated, so delete the constructor and assignment operators.
     type_list() ETL_DELETE;
     type_list(const type_list&) ETL_DELETE;
     type_list& operator =(const type_list&) ETL_DELETE;
@@ -106,6 +110,7 @@ namespace etl
 
   private:
 
+    // A type_list cannot be instantiated, so delete the constructor and assignment operators.
     type_list() ETL_DELETE;
     type_list(const type_list&) ETL_DELETE;
     type_list& operator =(const type_list&) ETL_DELETE;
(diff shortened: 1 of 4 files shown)


include/etl/basic_string.h
```diff
@@ -2809,20 +2810,25 @@ namespace etl
...
-      etl::mem_move(src, count, dst);
+      etl::mem_move(src, count, position);
 
-      truncated |= (src[count] != 0);
       current_size = size_t(start) + count;
       p_buffer[current_size] = 0;
 
 #if ETL_HAS_STRING_TRUNCATION_CHECKS
-      set_truncated(truncated);
+      set_truncated((length > size_t(free_space)) || truncated);
```
rolandreichweinbmw: Use static_cast<>() here, or define free_space as size_t in the definition since it is used as size_t everywhere.

rolandreichweinbmw (commented):
Is the Copilot-raised question of removed safety check already compensated?

denravonska:
> Is the Copilot-raised question of removed safety check already compensated?

I just fixed those. I'll stop force pushing now that someone is looking at it :)
Speaking of, there were some inconsistencies in the different implementations. Some used `- 1` while some used `- 1U`.

include/etl/basic_string.h
denravonska: Should I also fix the iterator version or leave it to keep the PR down?
I'd argue that all 3 of them should be `size_t`.

jwellbelove:
Speaking of, there were some inconsistencies in the different implementations. Some used `- 1` while some used `- 1U`.

They should be 1U as the sizes are of type `size_t`.

include/etl/basic_string.h
rolandreichweinbmw: Agree. It's not a big resulting diff. While we are touching the lines already, can you please do?

denravonska:
I'm torn about the `initialise()` removals. Maybe it can be solved in a better way.

denravonska:
> Speaking of, there were some inconsistencies in the different implementations. Some used `- 1` while some used `- 1U`.
> 
> They should be 1U as the sizes are of type `size_t`.

Alrighty, I can post a separate PR is nobody has fixed it already.

PR #1311, opened 2026-02-21
Enforce worse-case O(log n) dispatch for messages when using message router for c++11 and up

jwellbelove:
Changes the implementations of `receive(const etl::imessage&)` and `accepts(const etl::imessage&)`.

The 'fold expressions' for C++17, and the `switch/case` for C++11/14 are replaced by an internal `constexpr` dispatch jump table.

The dispatch complexity is O(1) if the message IDs handled by the router are contiguous, O(logN) if not. For O(1) the index of the message ID in the type list is the dispatch index (a simple ID - FirstId calculation), for O(logN), a sorted table of message IDs is binary searched for a match, and that index is used.

The order of the messages in the template parameter list is not important as they are sorted by message ID internally.

The public API has not been modified, except for the addition of a new public type `sorted_message_types`.
The C++03 compatible `switch/case` version is still available.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,7 @@ Its design goals include:
 
 -	Offering APIs that closely resemble those of the STL, enabling familiar and consistent usage.
 
--	Maintaining compatibility with C++98 while implementing many features introduced in later standards 
+-	Maintaining compatibility with C++98 while implementing many features introduced in later standards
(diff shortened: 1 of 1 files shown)


jwellbelove:
I'm closing this for the moment as the branch needs to be rebased on development first.

PR #1312, opened 2026-02-22
Enforce worse-case O(log n) dispatch for messages when using message router for c++11 and up

jwellbelove:
Changes the implementations of `receive(const etl::imessage&)` and `accepts(const etl::imessage&)`.

The 'fold expressions' for C++17, and the `switch/case` for C++11/14 are replaced by an internal `constexpr` dispatch jump table.

The dispatch complexity is O(1) if the message IDs handled by the router are contiguous, O(logN) if not. For O(1) the index of the message ID in the type list is the dispatch index (a simple ID - FirstId calculation), for O(logN), a sorted table of message IDs is binary searched for a match, and that index is used.

The order of the messages in the template parameter list is not important as they are sorted by message ID internally.

The public API has not been modified, except for the addition of a new public type `sorted_message_types`.
The C++03 compatible `switch/case` version is still available.

### DIFF
--- a/include/etl/basic_string.h
+++ b/include/etl/basic_string.h
@@ -2757,10 +2757,11 @@ namespace etl
     }
 
     //*********************************************************************
-    /// Common implementation for 'assign' and 'append' for iterators.
+    /// Common implementation for 'assign' and 'append' for non-pointer iterators.
     //*********************************************************************
     template <typename TIterator>
-    void append_impl(iterator position, TIterator first, TIterator last, bool truncated, bool secure)
+    typename etl::enable_if<!etl::is_pointer<typename etl::remove_reference<TIterator>::type>::value>::type
+    append_impl(iterator position, TIterator first, TIterator last, bool truncated, bool secure)
     {
       difference_type start      = etl::distance(p_buffer, position);
       difference_type count      = etl::distance(first, last);
@@ -2809,20 +2810,29 @@ namespace etl
         return;
       }
 
-      difference_type start      = etl::distance(p_buffer, position);
-      difference_type free_space = etl::distance(position, p_buffer + CAPACITY);
+      append_impl(position, src, get_string_length(src), truncated, secure);
+    }
 
-      pointer dst    = position;     
-      size_t  length = get_string_length(src);
-      size_t  count  = (length < size_t(free_space)) ? length : size_t(free_space);
-      etl::mem_move(src, count, dst);
+    //*********************************************************************
+    /// Core non-template implementation for 'assign' and 'append'.
+    //*********************************************************************
+    void append_impl(iterator position, const_pointer src, size_t length, bool truncated, bool secure)
+    {
+      size_t start      = static_cast<size_t>(etl::distance(p_buffer, position));
+      size_t free_space = static_cast<size_t>(etl::distance(position, p_buffer + CAPACITY));
+      size_t count      = etl::min(length, free_space);
 
-      truncated |= (src[count] != 0);
-      current_size = size_t(start) + count;
+#if ETL_IS_DEBUG_BUILD
+      ETL_ASSERT(start <= CAPACITY, ETL_ERROR(string_iterator));
+#endif
+
+      etl::mem_move(src, count, position);
+
+      current_size = start + count;
       p_buffer[current_size] = 0;
 
 #if ETL_HAS_STRING_TRUNCATION_CHECKS
-      set_truncated(truncated);
+      set_truncated((length > free_space) || truncated);
 #if ETL_HAS_ERROR_ON_STRING_TRUNCATION
       ETL_ASSERT(is_truncated() == false, ETL_ERROR(string_truncation));
 #endif
@@ -2840,6 +2850,16 @@ namespace etl
 #endif
 
       cleanup();
+    }
+
+    //*********************************************************************
+    /// Common implementation for 'assign' and 'append' for pointer iterators.
+    //*********************************************************************
+    template <typename TIterator>
+    typename etl::enable_if<etl::is_pointer<typename etl::remove_reference<TIterator>::type>::value>::type
+    append_impl(iterator position, TIterator first, TIterator last, bool truncated, bool secure)
+    {
+      append_impl(position, first, size_t(etl::distance(first, last)), truncated || is_truncated(), secure);
     }
 
     //*************************************************************************
--- a/include/etl/private/variant_variadic.h
+++ b/include/etl/private/variant_variadic.h
@@ -410,6 +410,12 @@ namespace etl
     template <typename T, typename... VTypes>
     friend ETL_CONSTEXPR14 const T&& get(const etl::variant<VTypes...>&& v);
 
+    template< class T, typename... VTypes >
+    friend ETL_CONSTEXPR14 etl::add_pointer_t<T> get_if(etl::variant<VTypes...>* pv) ETL_NOEXCEPT;
+
+    template< class T, typename... VTypes >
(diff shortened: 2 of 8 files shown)


jwellbelove:
@coderabbitai full review

include/etl/generators/message_router_generator.h
```diff
@@ -203,6 +205,12 @@ namespace etl
   {
   public:
 
+#if ETL_USING_CPP11
+    using message_packet       = etl::message_packet<>;                                               ///< The message packet type.
+    using message_types        = etl::type_list<>;                                                    ///< The message types.
+    using sorted_message_types = etl::type_list_sort_t<message_types, etl::compare_message_id_less>;  ///< The message types sorted by message id.
```
drewr95: I'm not sure these comments are adding any benefit.  These aliases you made are clear and self-documenting in my opinion.

include/etl/generators/message_router_generator.h
```diff
@@ -281,6 +289,12 @@ namespace etl
...
```
drewr95: Same here

include/etl/generators/message_router_generator.h
```diff
@@ -410,55 +424,94 @@ namespace etl
...
+  private:
+
+    using message_id_types = etl::type_list<etl::integral_constant<etl::message_id_t, TMessageTypes::ID>...>;
+
   public:
 
-    typedef etl::message_packet<TMessageTypes...> message_packet;
+    using message_packet       = etl::message_packet<TMessageTypes...>;                               ///< The message packet type.
+    using message_types        = etl::type_list<TMessageTypes...>;                                    ///< The message types.
+    using sorted_message_types = etl::type_list_sort_t<message_types, etl::compare_message_id_less>;  ///< The message types sorted by message id.
```
drewr95: same here

include/etl/generators/message_router_generator.h
```diff
@@ -410,55 +424,94 @@ namespace etl
...
       ETL_ASSERT(id_ <= etl::imessage_router::MAX_MESSAGE_ROUTER, ETL_ERROR(etl::message_router_illegal_id));
     }
 
+    //**********************************************
+    /// Allow visibility of base class receive() methods.
     //**********************************************
     using etl::imessage_router::receive;
 
+    //**********************************************
+    /// This will be called for all messages passed as an etl::imessage.
```
drewr95, line 521: I use doxygen for my code's documentation but never know what it smartly picks up on and what it doesn't.  I usually put a @brief or in this case \brief for description of the function.

include/etl/generators/message_router_generator.h
```diff
@@ -472,34 +525,75 @@ namespace etl
...
+      {
+        if (has_successor())
+        {
+          return get_successor().accepts(id);
+        }
+        else
+        {
+          return false;
+        }
+      }
```
drewr95, line 626: This would trigger the clang-tidy warning [readability-else-after-return](https://clang.llvm.org/extra/clang-tidy/checks/readability/else-after-return.html)

include/etl/message_router.h
```diff
@@ -191,6 +193,12 @@ namespace etl
   {
   public:
 
+#if ETL_USING_CPP11
+    using message_packet       = etl::message_packet<>;                                               ///< The message packet type.
+    using message_types        = etl::type_list<>;                                                    ///< The message types.
+    using sorted_message_types = etl::type_list_sort_t<message_types, etl::compare_message_id_less>;  ///< The message types sorted by message id.
```
drewr95: I think these comments can be removed.

include/etl/message_router.h
```diff
@@ -510,75 +604,195 @@ namespace etl
...
+                            contiguous_impl<Index + 1U>::value>
+    {
+    };
+
+    // The message ids are contiguous if there are 0 or 1 message types, or if each message id is one greater than the previous message id.
+    static constexpr bool Message_Ids_Are_Contiguous = (Number_Of_Messages <= 1U) ? true : contiguous_impl<0U>::value;
+
+    using handler_ptr              = void (*)(TDerived&, const etl::imessage&);         ///< Pointer to a handler function that takes a reference to the derived class and a reference to the message.
+    using message_dispatch_table_t = etl::array<handler_ptr, Number_Of_Messages>;       ///< The dispatch table type. An array of handler pointers, one for each message type.
+    using message_id_table_t       = etl::array<etl::message_id_t, Number_Of_Messages>; ///< The message id table type. An array of message ids, one for each message type.
```
drewr95, line 664: I remember reviewing this in an earlier file. May want to see if you can reduce duplication but having a shared type trait struct?

include/etl/message_router.h
```diff
@@ -510,75 +604,195 @@ namespace etl
...
-        }
-        else
+        // The IDs are not contiguous, so we need to do a binary search.
+        size_t left  = 0;
+        size_t right = Number_Of_Messages;
+
+        while (left < right)
         {
-          return false;
+          size_t mid = (left + right) / 2;
```
drewr95, line 736: Couldn't mid be cosntexpr?  left is being assigned an integral constant 0, and Number_Of_Messages is also an integral constant.  You could pre-compute mid.  Maybe even just store it an static inline constexpr variable.

include/etl/message_router.h
```diff
@@ -510,75 +604,195 @@ namespace etl
...
   template <typename TDerived>
   class message_router<TDerived> : public imessage_router
   {
   public:
 
-    using message_packet = etl::message_packet<>;
-    using message_types  = etl::type_list<>;
+    using message_packet = etl::message_packet<>;                                                     ///< The message packet type.
+    using message_types = etl::type_list<>;                                                           ///< The message types.
+    using sorted_message_types = etl::type_list_sort_t<message_types, etl::compare_message_id_less>;  ///< The message types sorted by message id.
```
drewr95: Maybe these types could also be shared in a type trait?  I've noticed this duplicated a couple of times.

include/etl/generators/message_router_generator.h
jwellbelove: I see clang-tidy's point in its example, but I'm not so sure it's true in a simple 'Do this or do that instead' type of conditional.

The 'fix' would be

```
if (has_successor())
{
  return get_successor().accepts(id);
}

return false;
```
which I think makes the one-or-the-other choice more obvious.

include/etl/generators/message_router_generator.h
jwellbelove: You're right, they are sort of stating the obvious.

include/etl/message_router.h
jwellbelove: `mid` is being recalculated at run-time on each pass of the loop as `left` or `right` change.

drewr95: Yeah that slipped my mind :)

include/etl/generators/message_router_generator.h
jwellbelove: 'll have to see how my experience with using Doxygen works out.

include/etl/generators/message_router_generator.h
jwellbelove: I just re-read what I typed.
I meant that this 
```
if (has_successor())
{
  return get_successor().accepts(id);
}
else
{
  return false;
}
```
makes the one-or-the-other choice more obvious.

include/etl/message_router.h
jwellbelove: I've implemented these as traits.

include/etl/generators/message_router_generator.h
drewr95: Yeah sounds fine to me :)

drewr95 (approved):

PR #1329, opened 2026-03-04
Add ETL_FORMAT_NO_FLOATING_POINT control macro for etl::format

MRNIU:
This pull request adds conditional compilation support to exclude floating-point formatting functionality from the `etl::format` module when the `ETL_FORMAT_NO_FLOATING_POINT` macro is defined. This allows the codebase to be built without floating-point dependencies, which is useful for embedded or resource-constrained environments. The changes are applied consistently across both the implementation (`include/etl/format.h`) and tests (`test/test_format.cpp`).

**Conditional floating-point support:**

* Wrapped all floating-point type support and formatting logic in `etl::format` (e.g., support for `float`, `double`, `long double`, and their formatters) with `#if !defined(ETL_FORMAT_NO_FLOATING_POINT)` guards to allow exclusion when the macro is defined.

**Test adjustments:**

* Wrapped all floating-point formatting tests in `test/test_format.cpp` with the same macro guard to ensure tests are only run when floating-point support is enabled.

See https://github.com/ETLCPP/etl/discussions/1326

### DIFF
--- a/include/etl/algorithm.h
+++ b/include/etl/algorithm.h
@@ -34,7 +34,7 @@ SOFTWARE.
 #define ETL_ALGORITHM_INCLUDED
 
 ///\defgroup algorithm algorithm
-/// Including reverse engineered algorithms from C++ 0x11, 0x14, 0x17
+/// Including reverse engineered algorithms from C++11, 14, 17
 /// Additional new variants of certain algorithms.
 ///\ingroup utilities
 
@@ -1193,7 +1193,12 @@ namespace etl
     typename etl::enable_if<etl::is_random_access_iterator<TIterator>::value, TIterator>::type
       rotate_general(TIterator first, TIterator middle, TIterator last) 
     {
-      if (first == middle || middle == last)
+      if (first == middle)
+      {
+        return last;
+      }
+
+      if (middle == last)
       {
         return first;
       }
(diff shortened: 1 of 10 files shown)


include/etl/format.h
```diff
@@ -138,9 +140,11 @@ namespace etl
     unsigned int,
     long long int,
     unsigned long long int,
+#if !defined(ETL_FORMAT_NO_FLOATING_POINT)
```
jwellbelove: What I've done in the past to unify the way that macro flags are handled is to transform the `#if defined` preprocessor commands into `#if` by adding helper macros to `platform.h`.

For example, in this case:

```
#if defined(ETL_FORMAT_NO_FLOATING_POINT)
  #define ETL_USING_FORMAT_FLOATING_POINT     0
  #define ETL_NOT_USING_FORMAT_FLOATING_POINT 1
#else
  #define ETL_USING_FORMAT_FLOATING_POINT     1
  #define ETL_NOT_USING_FORMAT_FLOATING_POINT 0
#endif
```

Then this line becomes:
`#if ETL_USING_FORMAT_FLOATING_POINT`

jwellbelove: Don't forget to update the `etl::traits` namespace definitions

PR #1335, opened 2026-03-10
Add pre C++20 Support for ETL_CONSTINIT using Compiler Extensions

christophruethingbmw:
The `constinit` keyword can be very useful to have more control on the initialization at startup. One use-case is to mitigate issues with the static initialiation order fiasko.

Althgouh the actual `constinit` keyword is only supported from C++20 upwards, compilers already have custom extensions to support it earlier. For `clang` we have the `__attribute__((require_constant_initialization))` since version 4 (see https://github.com/llvm/llvm-project/commit/341e825eae615f8019c667f15ff62ebc4ade005a) and for `gcc` we have `__constinit` since version 10.

Note, that there is also the C++11 attribute syntax `[[clang::require_constant_initialization]]` which we do not use, since it would require C++11 and thus introduce another level of `#if`.


christophruethingbmw:
@jwellbelove , can you have a look at these changes? Do you think we can integrate this?

PR #1336, opened 2026-03-10
Add a inplace_function constructor from a nullptr

Brambovich:
When using the etl::inplace_function constructor with a nullptr argument, it sees it as a function_ptr which will create a valid v-table with a invalid function pointer. When using is_valid() it will return true. The = operator has a override for a nullptr, which I now also added for the construction.

Simple compilation error code, will give ambiquous error:

```c++
etl::inplace_function<void()> callback(nullptr);
if (callback.is_valid())
{
    callback();
}
```

Simple faulty code:
```c++
void test_function(etl::inplace_function<void()> callback)
{
    if (callback.is_valid())
    {
        callback();
    }
}

void main()
{
    test_function(nullptr);
}
```


rolandreichweinbmw:
coderabbitai reasonably proposed noexcept. You can add it as NO_EXCEPT.

Brambovich:
Sure, will add a unit test soon! Thanks :)

jwellbelove (approved):

PR #1353, opened 2026-03-18
Manchester big endian support

tzijnge:
* Support Manchester for both little-endian and big-endian platforms
* Added c390x devcontainer to allow development in a GitHub Codespace on a big-endian platform (using QEMU)
* This pull request introduces a s390x Big-Endian development container environment with cross-compilation setup including Dockerfile, build script, CMake toolchain, and devcontainer configuration. The Manchester implementation is refactored to use explicit little-endian read/write helpers. Encode and decode functions for span have been simplified to a single constexpr capable implementation

### DIFF
--- a/include/etl/manchester.h
+++ b/include/etl/manchester.h
@@ -32,10 +32,8 @@ SOFTWARE.
 #define ETL_MANCHESTER_INCLUDED
 
 #include "platform.h"
-#include "endianness.h"
 #include "span.h"
 #include "static_assert.h"
-#include <cstring>
 
 ///\defgroup manchester manchester
 /// Manchester encoding and decoding
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -524,8 +524,9 @@ target_include_directories(etl_tests
 add_subdirectory(UnitTest++)
 target_link_libraries(etl_tests PRIVATE UnitTestpp)
 
+enable_testing()
 # Enable the 'make test' CMake target using the executable defined above
-add_test(etl_unit_tests etl_tests)
+add_test(NAME etl_unit_tests COMMAND etl_tests)
 
 # Since ctest will only show you the results of the single executable
 # define a target that will output all of the failing or passing tests


tzijnge:
@rolandreichweinbmw @jwellbelove This PR also includes a devcontainer and QEMU setup that I used in combination with GitHub Codespaces to test and develop the Manchester modifications. I think this setup is useful for anyone having to test on a big-endian platform and it may even be converted into a GitHub Actions workflow in the future.

Process:
1. Create Codespace as shown in picture. Select the right branch and devcontainer
2. Open Codespace in VS Code
3. In VS Code: press F1 and select 'CMake: Configure'. Select the s390x toolchain
4. In VS Code: press F1 and select 'CMake: Build'.
5. In VS Code: press F1 and select 'CMake: Run tests'.

image

tzijnge:
@coderabbitai full review

rolandreichweinbmw:
I support this approach, since it both:
* Fixes a test that currently fails on big endian
* Provides a new facility to easily test big endian on most developer machines. (And even prepares a relatively easy and cheap later integration into current ETL CI checks.)

serges147:
@tzijnge Timon, I wonder would it be possible to use already existing LE/BE features of `<etl/unaligned_type.h>` ?

include/etl/manchester.h
```diff
@@ -297,44 +316,13 @@ namespace etl
...
-    /// Encode a span of data with the minimum chunk size. This version is
-    /// constexpr so that it can be used to encode data at compile time.
+    /// Encode a span of data with the specified chunk size.
     ///\param source      The source data to encode.
     ///\param destination The destination buffer for encoded data.
     ///\tparam TChunk     The chunk size for encoding (default: uint_least8_t).
     //*************************************************************************
     template <typename TChunk = uint_least8_t>
-    static ETL_CONSTEXPR14 typename etl::enable_if<etl::is_same<TChunk, uint_least8_t>::value, void>::type encode(etl::span<const uint_least8_t> decoded, etl::span<uint_least8_t> encoded)
+    static ETL_CONSTEXPR14 void encode(etl::span<const uint_least8_t> decoded, etl::span<uint_least8_t> encoded)
```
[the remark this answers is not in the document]
tzijnge: The code doesn't work on platforms where `uint_least8_t` is not exactly equal to either `uint8_t`, `uint16_t` or `uint32_t`. That's accepted

tzijnge:
> @tzijnge Timon, I wonder would it be possible to use already existing LE/BE features of `<etl/unaligned_type.h>` ?

If you are referring to the `copy_value_to_store` functions, I believe they are indeed doing something similar, but I think the Manchester implementation is the most efficient. Manchester coding is an operation on a range of bits and endianness does not play a conceptual role, but the algorithm happens to map very well to a little-endian platform. When doing Manchester operations in chunks larger than 1 byte, the platform endianness starts playing a role because of how `memcpy` works. Therefore Manchester does not use `memcpy` to copy bytes, but a special helper function to make sure the copy is always done correctly regardless of the platform endianness.

In other words, it's probably possible to use `memcpy` and then reverse bytes on a big endian platform, but I think it only introduces additional code complexity and asymmetry between big and little endian platforms. Maybe I could use `etl::copy` or `etl::reverse_copy` depending on the endianness

test/CMakeLists.txt
```diff
@@ -524,8 +524,9 @@ target_include_directories(etl_tests
 add_subdirectory(UnitTest++)
 target_link_libraries(etl_tests PRIVATE UnitTestpp)
 
+enable_testing()
```
tzijnge, line 527: I made these changes to _tests/CMakelists.txt_ to be able to run the tests with CTest in devcontainer/codespace. I don't believe it has any effect on the current CI

`enable_testing()`
* Enable testing with ctest and with cross-compiler

`add_test(NAME etl_unit_tests COMMAND etl_tests)`
* CMake recognizes etl_tests as a CMake target (an executable built by the project), not just a string.
* It automatically prepends CMAKE_CROSSCOMPILING_EMULATOR when set
* It can resolve the actual binary path (handles Debug/Release configs on multi-config generators)
* It establishes proper dependencies
Generator expressions work correctly

rolandreichweinbmw: Agree. This was missing from CMakeLists.txt anyway.

PR #1354, opened 2026-03-18
Add Dockerfile for local development

rolandreichweinbmw:
The current .devcontainer/Dockerfile doesn't work because of non-availability of the referenced base container, and also because of errors later on.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,0 +1,30 @@
+ARG BASE_IMAGE_NAME="debian:trixie-20260316"
+
+FROM ${BASE_IMAGE_NAME}
+
+RUN apt-get update && export DEBIAN_FRONTEND=noninteractive \
+    && apt-get -y install --no-install-recommends \
+        clang \
+        cmake \
+        g++ \
+        git \
+        make \
+        python3-cogapp \
+        wget \
+    && rm -rf /var/lib/apt/lists/*
+
+ARG USERNAME=developer
+ARG USER_UID=1000
+ARG USER_GID=${USER_UID}
+RUN EXISTING_USER=$(getent passwd ${USER_UID} | cut -d: -f1) \
+    && if [ -n "${EXISTING_USER}" ] && [ "${EXISTING_USER}" != "${USERNAME}" ]; then \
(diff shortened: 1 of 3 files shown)


rolandreichweinbmw:
> Can you address the AI bot's concerned regarding VS Code direct usage?

Thanks for the note! It was not clear to me that the .devcontainers/Dockerfile is intended for GitHub Codespaces and kind of conflicts with requirements for running docker locally (for development, testing or CI purposes).

Therefore, I separated them now. Keeping .devcontainers/* untouched for Github Codespaces (even though I think the image clang:latest is not working - but please @wsciaroni have a separate look), and adding docker/* for a locally useable Dockerfile.

Documentation in docs/docker.md

rolandreichweinbmw:
Regarding debian:trixie-20260316 vs. debian:trixie, I intentionally focused on 100% reproducability instead of receiving security updates (the latter is not currently a priority for the build environments of ETL whose target is a broad range of environments old and new).

wsciaroni:
I got codespaces working by dropping the cogap package. Works on my fork. What's driving the desire to separate them?

This is working just fine for me: https://github.com/wsciaroni/etl/blob/master/.devcontainer%2FDockerfile

wsciaroni:
I opened an MR which fixes the GitHub codepace functionality for me. https://github.com/ETLCPP/etl/pull/1360

wsciaroni:
> What's driving the desire to separate them?

Forgive my quick reply.

After digging into it a bit more, there was an error in the Dockerfile specifying a strange version of the image that was not recognizable by a docker runtime.

I have updated the devcontainer's `Dockerfile` to use the same image as the `devcontainer.json`. I set both of these to the [Microsoft cpp devcontainer](https://hub.docker.com/r/microsoft/devcontainers-cpp) version 2. The AI reviewer doesn't like that... We can discuss there.

My recommendation is to NOT have two dockerfiles that have the same intent / toolchain.

We can definitely make the main one (`.devcontainer/Dockerfile`) meet all of our needs.

rolandreichweinbmw:
Closing in favour of https://github.com/ETLCPP/etl/pull/1360 (as the correct fix)

PR #1358, opened 2026-03-19
Bug fixes in variant_variadic.h

rolandreichweinbmw:


include/etl/private/variant_variadic.h
```diff
@@ -569,16 +569,13 @@ namespace etl
       : operation(other.operation)
       , type_id(other.type_id)
     {
-      if (this != &other)
```
jwellbelove, line 572: Should we really be removing the 'self' check that eliminates unnecessary copies?

rolandreichweinbmw: In a copy constructor, we can safely assume that `this` is always different from `&other`.

Even considering the case of placement new when `this == &other`, because this is undefined behavior anyway.

This relates to this copy constructor here. In copy assignment, we keep the check.

jwellbelove: Yes, I should have realised that.

PR #1359, opened 2026-03-19
Restrict etl::atomic for general types

rolandreichweinbmw:
std::atomic<T> restricts T for general types to several properties:

* std::is_trivially_copyable<T>::value
* std::is_copy_constructible<T>::value
* std::is_move_constructible<T>::value
* std::is_copy_assignable<T>::value
* std::is_move_assignable<T>::value
* std::is_same<T, typename std::remove_cv<T>::type>::value

Doing this here also.
 
Needs adding is_copy_assignable and is_move_assignable, and adjustments to is_trivially_copyable and is_assignable.

### DIFF
--- a/.devcontainer/s390x/Dockerfile
+++ b/.devcontainer/s390x/Dockerfile
@@ -1,0 +1,40 @@
+# s390x Big-Endian Test Environment for ETL
+# Uses QEMU user-mode emulation to run s390x binaries on x64 host
+FROM debian:trixie
+
+# Avoid prompts from apt
+ENV DEBIAN_FRONTEND=noninteractive
+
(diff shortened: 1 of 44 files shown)


rolandreichweinbmw:
Resolved mutable T value vs. volatile qualified methods, as suggested by coderabbitai.

rolandreichweinbmw:
Added a separate commit that removes the volatile qualified methods, because they are deprecated in C++20 and it turned out that they don't work in the way that users expect.

If we don't want to include this last commit of the PR, we would need to conditionally disable the new test test_atomic_non_scalar_trivially_copyable_struct on MSVC (because the std version there doesn't support the volatile overloads).

PR #1360, opened 2026-03-19
Update Dockerfile and Devcontainer.

wsciaroni:
Fixed GitHub Workspaces and local devcontainer by moving to the cpp devcontainer maintained by microsoft.


.devcontainer/Dockerfile
```diff
@@ -1,20 +1,38 @@
-ARG BASE_IMAGE_NAME="clang:latest"
+ARG BASE_IMAGE_NAME="mcr.microsoft.com/devcontainers/cpp:2@sha256:a5eb5a1e9109af88bf82ebb0f71903608a68144851ed1e4b852e31b251ac59c6"
 
 FROM ${BASE_IMAGE_NAME}
 
 ARG REINSTALL_CMAKE_VERSION_FROM_SOURCE="3.31.7"
+ARG DEBIAN_SNAPSHOT="20260223T000000Z"
```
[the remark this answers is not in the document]
wsciaroni: We're leaving this snapshot in, because it is what gets used by GitHub codespace.

.devcontainer/devcontainer.json
```diff
@@ -5,7 +5,9 @@
 	"build": {
 		"dockerfile": "./Dockerfile",
 		"args": {
-			"BASE_IMAGE_NAME": "mcr.microsoft.com/devcontainers/cpp:debian-12"
+			"BASE_IMAGE_NAME": "mcr.microsoft.com/devcontainers/cpp:2@sha256:a5eb5a1e9109af88bf82ebb0f71903608a68144851ed1e4b852e31b251ac59c6",
+			"DEBIAN_SNAPSHOT": "20260223T000000Z",
```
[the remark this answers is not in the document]
wsciaroni: MR description updated to match behavior.

.devcontainer/Dockerfile
```diff
@@ -1,20 +1,38 @@
...
+COPY ./reinstall-cmake.sh /usr/local/share/reinstall-cmake.sh
 
-RUN apt-get update && export DEBIAN_FRONTEND=noninteractive \
+RUN set -eux \
+    && export DEBIAN_FRONTEND=noninteractive \
+    && if [[ "${DEBIAN_SNAPSHOT}" != "none" ]]; then \
+        test -f /etc/apt/sources.list.d/debian.sources; \
+        sed -ri "s|^URIs: http://deb.debian.org/debian$|URIs: http://snapshot.debian.org/archive/debian/${DEBIAN_SNAPSHOT}|g" /etc/apt/sources.list.d/debian.sources; \
+        sed -ri "s|^URIs: http://deb.debian.org/debian-security$|URIs: http://snapshot.debian.org/archive/debian-security/${DEBIAN_SNAPSHOT}|g" /etc/apt/sources.list.d/debian.sources; \
+        printf 'Acquire::Check-Valid-Until "false";\n' > /etc/apt/apt.conf.d/99snapshot; \
```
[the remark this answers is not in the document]
wsciaroni: Updated.

wsciaroni:
This seems to have fixed the default container, but the rest of the devcontainers are not functional. I'm not sure if they're functional right now on master though.

rolandreichweinbmw:
@wsciaroni : Thank you for having a look into this!

I agree that separate Dockerfiles don't make sense if we can have a consolidated solution that fits the Github Codespaces and also local development requirements. The original motivation was not to interfere with the Codespaces solution. But if that can be solved jointly, even better!

We probably don't need https://github.com/ETLCPP/etl/pull/1354 anymore, then.

Taking over documentation drom there might be good, anyway.

rolandreichweinbmw:
When building this with `docker build -t wsciaroni .devcontainer/`, I get:

```
[+] Building 1.4s (6/8)                                                                                                                                                                           docker:default
 => [internal] load build definition from Dockerfile                                                                                                                                                        0.0s
 => => transferring dockerfile: 2.58kB                                                                                                                                                                      0.0s
 => [internal] load metadata for mcr.microsoft.com/devcontainers/cpp:2@sha256:a5eb5a1e9109af88bf82ebb0f71903608a68144851ed1e4b852e31b251ac59c6                                                              1.2s
 => [internal] load .dockerignore                                                                                                                                                                           0.0s
 => => transferring context: 2B                                                                                                                                                                             0.0s
 => [internal] load build context                                                                                                                                                                           0.0s
 => => transferring context: 2B                                                                                                                                                                             0.0s
 => CANCELED [1/4] FROM mcr.microsoft.com/devcontainers/cpp:2@sha256:a5eb5a1e9109af88bf82ebb0f71903608a68144851ed1e4b852e31b251ac59c6                                                                       0.0s
 => => resolve mcr.microsoft.com/devcontainers/cpp:2@sha256:a5eb5a1e9109af88bf82ebb0f71903608a68144851ed1e4b852e31b251ac59c6                                                                                0.0s
 => => sha256:a5eb5a1e9109af88bf82ebb0f71903608a68144851ed1e4b852e31b251ac59c6 1.61kB / 1.61kB                                                                                                              0.0s
 => => sha256:9ed24906bf15886dedaf817fa2a7a6db4f864c8e0ad601526ab9d5cd75457a1b 3.52kB / 3.52kB                                                                                                              0.0s
 => => sha256:049276894d7c9ba7e8753f679b37a73029045e400319b8dff43bba2adc68b67e 19.00kB / 19.00kB                                                                                                            0.0s
 => ERROR [2/4] COPY ./reinstall-cmake.sh /usr/local/share/reinstall-cmake.sh                                                                                                                               0.0s
------
 > [2/4] COPY ./reinstall-cmake.sh /usr/local/share/reinstall-cmake.sh:
------
Dockerfile:15
--------------------
  13 |
  14 |     # Optionally install the cmake for vcpkg
  15 | >>> COPY ./reinstall-cmake.sh /usr/local/share/reinstall-cmake.sh
  16 |
  17 |     RUN set -eux \
--------------------
ERROR: failed to solve: failed to compute cache key: failed to calculate checksum of ref 1a22efe4-6c7a-497e-ab31-ec38b0410b2a::tku48zhs32rddmgq1ee1t716v: "/reinstall-cmake.sh": not found
```

Wondering why we need this reinstall-cmake in the first place. Why not using cmake from the distribution directly, when it's recent?

rolandreichweinbmw:
And when removing the reinstall-cmake parts, I get:

```
 > [2/2] RUN set -eux     && export DEBIAN_FRONTEND=noninteractive     && if [[ "20260223T000000Z" != "none" ]]; then         snapshot_main="https://snapshot.debian.org/archive/debian/20260223T000000Z";
  snapshot_security="https://snapshot.debian.org/archive/debian-security/20260223T000000Z";         if [[ -f /etc/apt/sources.list.d/debian.sources ]]; then             sed -ri "s|^URIs: https?://deb.debian.or
g/debian$|URIs: ${snapshot_main}|g" /etc/apt/sources.list.d/debian.sources;             sed -ri "s|^URIs: https?://deb.debian.org/debian-security$|URIs: ${snapshot_security}|g" /etc/apt/sources.list.d/debian.s
ources;         elif [[ -f /etc/apt/sources.list ]] || compgen -G "/etc/apt/sources.list.d/*.list" > /dev/null; then             for list_file in /etc/apt/sources.list /etc/apt/sources.list.d/*.list; do
          [[ -f "${list_file}" ]] || continue;                 sed -ri "s|https?://deb.debian.org/debian|${snapshot_main}|g" "${list_file}";                 sed -ri "s|https?://security.debian.org/debian-secur
ity|${snapshot_security}|g" "${list_file}";                 sed -ri "s|https?://deb.debian.org/debian-security|${snapshot_security}|g" "${list_file}";             done;         else             echo "ERROR: DE
BIAN_SNAPSHOT='20260223T000000Z' is set, but no supported apt source files were found.";             echo "Expected /etc/apt/sources.list.d/debian.sources or one or more .list files under /etc/apt/.";
    exit 1;         fi;         printf 'Acquire::Check-Valid-Until "false";\n' > /etc/apt/apt.conf.d/99snapshot;     fi     && apt-get update     && apt-get -y install --no-install-recommends         git
   wget         python3-cogapp     && rm -rf /var/lib/apt/lists/*:
0.239 + export DEBIAN_FRONTEND=noninteractive
0.239 + DEBIAN_FRONTEND=noninteractive
0.239 + [[ 20260223T000000Z != \n\o\n\e ]]
0.239 + snapshot_main=https://snapshot.debian.org/archive/debian/20260223T000000Z
0.239 + snapshot_security=https://snapshot.debian.org/archive/debian-security/20260223T000000Z
0.239 + [[ -f /etc/apt/sources.list.d/debian.sources ]]
0.239 + sed -ri 's|^URIs: https?://deb.debian.org/debian$|URIs: https://snapshot.debian.org/archive/debian/20260223T000000Z|g' /etc/apt/sources.list.d/debian.sources
0.243 + sed -ri 's|^URIs: https?://deb.debian.org/debian-security$|URIs: https://snapshot.debian.org/archive/debian-security/20260223T000000Z|g' /etc/apt/sources.list.d/debian.sources
0.243 + printf 'Acquire::Check-Valid-Until "false";\n'
0.243 + apt-get update
0.267 Ign:1 https://snapshot.debian.org/archive/debian/20260223T000000Z trixie InRelease
0.270 Ign:2 https://snapshot.debian.org/archive/debian/20260223T000000Z trixie-updates InRelease
0.273 Ign:3 https://snapshot.debian.org/archive/debian-security/20260223T000000Z trixie-security InRelease
1.279 Ign:3 https://snapshot.debian.org/archive/debian-security/20260223T000000Z trixie-security InRelease
1.283 Ign:2 https://snapshot.debian.org/archive/debian/20260223T000000Z trixie-updates InRelease
1.285 Ign:1 https://snapshot.debian.org/archive/debian/20260223T000000Z trixie InRelease
3.289 Ign:1 https://snapshot.debian.org/archive/debian/20260223T000000Z trixie InRelease
3.292 Ign:2 https://snapshot.debian.org/archive/debian/20260223T000000Z trixie-updates InRelease
3.296 Ign:3 https://snapshot.debian.org/archive/debian-security/20260223T000000Z trixie-security InRelease
7.301 Err:3 https://snapshot.debian.org/archive/debian-security/20260223T000000Z trixie-security InRelease
7.301   SSL connection failed: error:0A0000C6:SSL routines::packet length too long / Success [IP: 169.254.254.1 3128]
7.305 Err:2 https://snapshot.debian.org/archive/debian/20260223T000000Z trixie-updates InRelease
7.305   SSL connection failed: error:0A0000C6:SSL routines::packet length too long / Success [IP: 169.254.254.1 3128]
7.308 Err:1 https://snapshot.debian.org/archive/debian/20260223T000000Z trixie InRelease
7.308   SSL connection failed: error:0A0000C6:SSL routines::packet length too long / Success [IP: 169.254.254.1 3128]
7.311 Reading package lists...
```

I personally would not try to use snapshot.debian.org which only opens another risky dependency on a further external server somewhere in the internet, including their TLS configuration.

wsciaroni:
> Wondering why we need this reinstall-cmake in the first place. Why not using cmake from the distribution directly, when it's recent?

It used to be that the cmake version included with Debian was quite old. So, you pretty much always had to install it from their package here.

Do you know what version was installed from apt?

wsciaroni:
> I personally would not try to use snapshot.debian.org which only opens another risky dependency on a further external server somewhere in the internet, including their TLS configuration.

Two things:
1. I think this error is my fault. I added the 's' for the uri that is failing after I took this little trick from elsewhere.
2. That is a fair concern regarding the external dependency and TLS risk. However, standard deb.debian.org is also an external infrastructure dependency (a dynamically routed CDN). By using the standard mirrors, our builds are non-deterministic and subject to upstream package removals (like cogap) or silent updates. Relying on snapshot.debian.org trades a slight availability risk for guaranteed, reproducible builds. If the snapshot server goes down, we fail loudly; if a standard mirror updates a package, we might fail silently.

rolandreichweinbmw:
I'm sure the cmake version issue comes from old pre-trixie (Debian 13) times. In Debian 13, we can assume cmake 3.31.6. I wonder if cmake 3.31.7 as used via reinstall-cmake is really necessary?

wsciaroni:
> I'm sure that comes from old pre-trixie (Debian 13) times. In Debian 13, we can assume cmake 3.31.6. I wonder if cmake 3.31.7 as used via reinstall-cmake is really necessary?

3.31.6 is plenty new. No need for 3.31.7.

I wonder if you were experiencing the error because of the context directory path. I'm going to remove the script if I can get the devcontainer variants to work without it.

rolandreichweinbmw:
I agree that referenced fix versions is best.

However, I'd prefer to limit that to a single server location. Microsoft/Github instead of Debian additionally is fine for me.

Using snapshot.debian.org for all of that looks somehow brittle.

.devcontainer/devcontainer.json
```diff
@@ -5,7 +5,9 @@
...
+			"REINSTALL_CMAKE_VERSION_FROM_SOURCE": "3.31.7"
```
rolandreichweinbmw: I think you can remove this also.

wsciaroni: Removed.

wsciaroni:
After some more tweaks I have gotten the Dockerfile to build devcontainers for each of the items.

Additionally, I added support for the ~clang22~ clang21 image. I think we're good to go now.

.devcontainer/Dockerfile
```diff
@@ -1,20 +1,62 @@
-ARG BASE_IMAGE_NAME="clang:latest"
+ARG BASE_IMAGE_NAME="mcr.microsoft.com/devcontainers/cpp:2@sha256:a5eb5a1e9109af88bf82ebb0f71903608a68144851ed1e4b852e31b251ac59c6"
 
 FROM ${BASE_IMAGE_NAME}
 
 ARG REINSTALL_CMAKE_VERSION_FROM_SOURCE="3.31.7"
```
rolandreichweinbmw: I think this is still a leftover, and `REINSTALL_CMAKE_VERSION_FROM_SOURCE` is not necessary anymore?

.devcontainer/Dockerfile
```diff
@@ -1,20 +1,62 @@
...
+        snapshot_security="http://snapshot.debian.org/archive/debian-security/${DEBIAN_SNAPSHOT}"; \
+        if [[ -f /etc/apt/sources.list.d/debian.sources ]]; then \
+            sed -ri "s|^URIs: https?://deb.debian.org/debian$|URIs: ${snapshot_main}|g" /etc/apt/sources.list.d/debian.sources; \
+            sed -ri "s|^URIs: https?://deb.debian.org/debian-security$|URIs: ${snapshot_security}|g" /etc/apt/sources.list.d/debian.sources; \
+        elif [[ -f /etc/apt/sources.list ]] || compgen -G "/etc/apt/sources.list.d/*.list" > /dev/null; then \
+            for list_file in /etc/apt/sources.list /etc/apt/sources.list.d/*.list; do \
+                [[ -f "${list_file}" ]] || continue; \
+                sed -ri "s|https?://deb.debian.org/debian|${snapshot_main}|g" "${list_file}"; \
+                sed -ri "s|https?://security.debian.org/debian-security|${snapshot_security}|g" "${list_file}"; \
+                sed -ri "s|https?://deb.debian.org/debian-security|${snapshot_security}|g" "${list_file}"; \
```
rolandreichweinbmw, line 26: This is still related to my previous comment: My use case is to run docker locally via Dockerfile directly. We now depend on:
* mcr.microsoft.com
* snapshot.debian.org
* deb.debian.org
* security.debian.org
* Fastly for pip

of which only one is an alias of one other.

In case of unavailability, the risk multiplies up. But if we don't find a better solution that fits everyone, we'll probably need to live with that.

Works for me now, thanks!

rolandreichweinbmw:
> After some more tweaks I have gotten the Dockerfile to build devcontainers for each of the items.
> 
> Additionally, I added support for the clang22 image. I think we're good to go now.

Did you add `.devcontainers/clang22/*`? At least currently, it's not contained in this branch. Maybe you meant a different location or still need to explicitly commit and push it?

.devcontainer/clang21/devcontainer.json
```diff
@@ -0,0 +1,13 @@
+// For format details, see https://aka.ms/devcontainer.json. For config options, see the
+// README at: https://github.com/devcontainers/templates/tree/main/src/cpp
+{
+	"name": "Clang 21",
+	"build": {
+		"dockerfile": "../Dockerfile",
+		"args": {
+			"BASE_IMAGE_NAME": "silkeh/clang:21-bullseye",
```
rolandreichweinbmw, line 8: Why use this very outdated Debian version (from 2021) when there is also `silkeh/clang:21-trixie` (from 2025) available?

wsciaroni:
> > After some more tweaks I have gotten the Dockerfile to build devcontainers for each of the items.
> > 
> > Additionally, I added support for the clang22 image. I think we're good to go now.
> 
> Did you add `.devcontainers/clang22/*`? At least currently, it's not contained in this branch. Maybe you meant a different location or still need to explicitly commit and push it?

Sorry, typo. I added a clang21 image. The clang22 docker image wasn't available yet.

wsciaroni:
@rolandreichweinbmw 

So, devcontainers that actually install stuff are always... A bit fragile.

Not only are you relying on the URLs you mentioned, you're also relying on the servers hosting the actual packages that you're installing. 

I think what I would do is create another repository within the ETLCPP GitHub Org to house all of the containers. That repository would have a pipeline that builds the containers once and publishes them to the GitHub container registry. 

This allows for ultimate stability as we can point these Dev containers to the GitHub container registry and at "runtime/build time" they will only depend on GitHub container registry being up. 

What do you think of that?

rolandreichweinbmw:
Sounds doable also. You can decide either way. It both has pros and cons. Current status has the multiple download location increased risk, and with the additional container location another piece of complexity would need to be maintained in the long run, by the project. Questionable also.

If you ask me, just keep the current implementation for now.

.devcontainer/clang21/devcontainer.json
wsciaroni: For consistency with the other clang containers

.devcontainer/Dockerfile
wsciaroni: Agreed. Need to remove.

wsciaroni:
```suggestion
```

.devcontainer/Dockerfile
wsciaroni: I don't see a way to get rid of the dependency on:

- some container registry (mcr.microsoft.com)
- deb.debian.org
- security.debian.org
- Fastly for pip

without building and pushing our containers to a container registry.

If you'd like to remove the dependency on snapshot, you can add the `--build-arg DEBIAN_SNAPSHOT=none` to your docker command. That will make it fall back to the non-snapshot stuff.

wsciaroni:
> If you ask me, just keep the current implementation for now.

So, should we go through with its current implementation on this PR? I believe it is good to go.

I can experiment with a proof of concept on the container registry thing after we merge this. The most compelling reason for me to want to build our own and push to a container registry is to actually install the toolchains ourselves on top of debian. I'm not keeping a close eye on what's in the silkeh clang images.

rolandreichweinbmw:
Agree!

PR #1376, opened 2026-04-05
Run generator_test.py in CI checks

rolandreichweinbmw:
We currently have 2 facilities for testing generators: Running generator.bat directly and checking the result, and generator_test.py. Only the first one was run in the CI checks, silently missing if the second one failed. Now adding the second one also to the workflow.


jwellbelove:
Actually, I'm not sure that `generate.bat` should actually be run at all at CI.
I think that `generator_test` is only one that should be run, as what we are really trying to do is ensure the generators would exactly re-generate the existing files.

rolandreichweinbmw:
Agreed. Removed generator.bat, kept generator_test.py in CI check.

drewr95 (approved):

jwellbelove (approved):

PR #1379, opened 2026-04-05
Optimize formatting in format.h for float values

rolandreichweinbmw:
The old implementation was an order of magnitude slower due to conversion of float values back and forth where actually only power of 2 was calculated.

### DIFF
--- a/include/etl/format.h
+++ b/include/etl/format.h
@@ -1039,22 +1039,27 @@ namespace etl
     {
       const size_t fractional_decimals = 6; // default
 
-      T                      integral;
-      T                      fractional = modf(value, &integral);
-      bool                   sign;
-      unsigned long long int fractional_int;
-      unsigned long long int integral_int;
-      if (integral < 0.0)
+      // Detect sign using signbit to correctly handle -0.0
+      bool sign = signbit(value);
+
+      T integral;
+      T fractional = modf(value, &integral);
+
+      // Take absolute values to avoid casting negative values to unsigned
+      if (sign)
       {
-        sign           = true;
-        fractional_int = static_cast<unsigned long long int>(-fractional * pow(10., fractional_decimals));
-        integral_int   = static_cast<unsigned long long int>(-integral);
+        fractional = -fractional;
+        integral   = -integral;
       }
-      else
+
+      unsigned long long int scale          = int_pow<unsigned long long int>(10, fractional_decimals);
+      unsigned long long int fractional_int = static_cast<unsigned long long int>(round(fractional * scale));
+      unsigned long long int integral_int   = static_cast<unsigned long long int>(integral);
+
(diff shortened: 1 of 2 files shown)


serges147:
@rolandreichweinbmw Take a look here please: https://github.com/serges147/etl_avl/pull/1
This also solves red unit tests on 32bits platform (built with -m32 flag).

Maybe you can apply my rounding (and unit tests changes) to your PR?

rolandreichweinbmw:
> Maybe you can apply my rounding (and unit tests changes) to your PR?

Thank you for the hint! Looks correct. Let me test - I hope my optimizations are not too much regressed by this.

rolandreichweinbmw:
Update: I added the round() approach, thanks to @serges147 !

Tested with the cross-compile-and-test-via-qemu from https://github.com/ETLCPP/etl/pull/1380 for:
* armhf
* i386
* riscv64
* s390x

Those typically showed this kind of issue previously, and it looks like it is solved consistently now.

serges147:
> Those typically showed this kind of issue previously, and it looks like it is solved consistently now.

Almost, see https://github.com/ETLCPP/etl/pull/1394

rolandreichweinbmw:
> Almost, see #1394

Interesting! I meant the format issue is solved, and I consider #1394 a separate issue, solved by your commit.

I added the foreign cross compiled architectures in https://github.com/ETLCPP/etl/pull/1380 which can be automatically covered via qemu in the CI, but big endian 32 bit is missing. Is there a concrete docker base image you would propose here to be added? armeb? Or maybe powerpc is easier since a standard (yet unofficial) port exists?

rolandreichweinbmw:
Update: I just added powerpc as another foreign architecture to cross-compile and qemu-test in https://github.com/ETLCPP/etl/pull/1380
Now, big and little endian for both 32 and 64 bit should be covered.

serges147:
>  Is there a concrete docker base image you would propose here to be added? armeb? Or maybe powerpc is easier since a standard (yet unofficial) port exists?

At first I tried to find something ready to use, like docker image, or at least toolchain. But no luck. Linaro dropped armeb releases long long time ago (like gcc 7.5 or so). So I've built gcc 15.2 armeb 32-bits toolchain myself, using crosstool-ng. And qemu-armeb (almost) successfully ran etl_tests on my Ubuntu 24.04 amd64 machine. I even managed to debug/step the tests in CLion (using embedded remote gdb target). I mentioned "almost" b/c:
- couple unit tests failed (hence the pr)
- initially `etl_tests` didn't ran at all (segfault 11 due some issued with eabi clib and its DSO stuff), but when I rebuilt toolchain with uClib then everything worked fine.

drewr95 (approved):

jwellbelove (approved):

PR #1380, opened 2026-04-05
Extensions for testing

rolandreichweinbmw:
This PR extends several aspects of testing:
* Support for cross-compiling armhf, i386, powerpc, riscv64 and (extending) s390x, to run in CI checks via qemu
* Devcontainers for the aforementioned
* Fixes to source code to support all those architectures
* Extending run-tests.sh to run tests for all supported C++ versions
* Fixing exit code of run-tests.sh (always returned 0)
* Documentation of testing and generators


docs/generators.md
```diff
@@ -0,0 +1,190 @@
...
+
+This script:
+
+1. Runs Cog on every `*_generator.h` file.
+2. Compares each output against the corresponding file in
+   `include/etl/private/`.
+3. Reports success or failure.
+
+The `generator.yml` GitHub Actions workflow runs this automatically on
+every push and pull request.
```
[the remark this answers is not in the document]
rolandreichweinbmw: This is fixed separately in https://github.com/ETLCPP/etl/pull/1376

.devcontainer/armhf/devcontainer.json
```diff
@@ -0,0 +1,29 @@
...
+          "-DBUILD_TESTS=ON",
+          "-DNO_STL=OFF",
+          "-DETL_CXX_STANDARD=23"
+        ],
+        "cmake.buildDirectory": "${workspaceFolder}/build-armhf",
+        "cmake.generator": "Ninja"
+      }
+    }
+  },
+  "remoteUser": "root"
```
christophruethingbmw, line 28: As far as I can see below, you have created a non-root user in the image, is there a reason we still run it with root here (assuming the `remoteUser` option here specifies the user to use when connecting to the running container)?

.github/workflows/gcc-c++23.yml
```diff
@@ -23,7 +23,7 @@ jobs:
         export ASAN_OPTIONS=alloc_dealloc_mismatch=0,detect_leaks=0
         export CC=gcc
         export CXX=g++
-        cmake -DBUILD_TESTS=ON -DNO_STL=OFF -DETL_USE_TYPE_TRAITS_BUILTINS=OFF -DETL_USER_DEFINED_TYPE_TRAITS=OFF -DETL_FORCE_TEST_CPP03=OFF -DETL_CXX_STANDARD=23 ./
```
christophruethingbmw, line 26: I think this was simply wrong before? At least I do not see that this PR changes the define in the code?

.github/workflows/gcc-c++23-armhf.yml
```diff
@@ -0,0 +1,49 @@
...
+  build-gcc-cpp23-linux-no-stl-armhf:
+    name: GCC C++23 Linux - No STL - armhf
+    runs-on: ubuntu-24.04
+    container:
+      image: debian:trixie
+
+    steps:
+    - uses: actions/checkout@v4
+
+    - name: Install dependencies
```
christophruethingbmw: Would it make sense to re-use the `Dockefile`s from above also in the CI? That would make the CI job also more easily reproducible locally and simplifies maintainability since there is just one place to adapt?

christophruethingbmw (commented):
Looks good to me overall and really improves testing on non-x64 plaforms.

.github/workflows/gcc-c++23.yml
rolandreichweinbmw: Exactly, it's a bug fix.

.devcontainer/armhf/devcontainer.json
rolandreichweinbmw: I took over the originally contributed `devcontainer.json` by @tzijnge . Not sure about the consequences of changing `remoteUser` now in the respective environments. E.g. in github workspaces: Do they need root permissions, e.g. to install via `apt`?

.github/workflows/gcc-c++23-armhf.yml
rolandreichweinbmw: Absolutely. Adjusted accordingly.

PR #1381, opened 2026-04-05
Add missing syntax checks

rolandreichweinbmw:
Now, instead of throwing #error on CPP03, conditionally compile all headers. This enables us to just include the headers and check them for syntax issues.

ETL_IN_UNIT_TEST is not defined in run-syntax-checks.sh anymore. This leads to a more realistic scenario, still checking all headers. This already uncovered bugs (also fixed in the included commits).


drewr95 (approved):

test/test_rounded_integral_division.cpp
```diff
@@ -34,6 +34,8 @@ SOFTWARE.
 #include <limits>
 #include <stdint.h>
 
+#if ETL_USING_CPP11
```
jwellbelove: The unit tests are never run for standards less than C++11

test/test_manchester.cpp
```diff
@@ -6,6 +6,8 @@
 #include <etl/span.h>
 #include <utility>
 
+#if ETL_USING_CPP11
```
jwellbelove: The unit tests are never run for standards less than C++11

test/syntax_check/crc.h.t.cpp
```diff
@@ -0,0 +1,29 @@
...
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
+******************************************************************************/
+
+#include <etl/crc.h>
```
jwellbelove: `crc`.h just includes all of the crc headers, which means the syntax check is run twice for each crc type.

rolandreichweinbmw: Thanks for the note! Removing that one.

test/test_rounded_integral_division.cpp
rolandreichweinbmw: OK, removing.

test/test_manchester.cpp
rolandreichweinbmw: OK, removing.

Issue #1134, opened 2025-06-21
Request: make vector methods Noexcept


khauersp:
I'm getting warnings from clang-tidy about vector objects that are being created statically possibly throwing an exception. I think ETL_NOEXCEPT can be added to the methods inside the vector class

rolandreichweinbmw:
Done in 20.48.0

PR #1383, opened 2026-04-05
Make etl::variant and etl::vector methods noexcept

### LINKED ISSUE #1133: Request: make variant methods Noexcept
I'm getting warnings from clang-tidy about variant objects that are being created statically possibly throwing an exception. I think ETL_NOEXCEPT can be added to the methods inside the variant class

rolandreichweinbmw:
Fixes https://github.com/ETLCPP/etl/issues/1133 and https://github.com/ETLCPP/etl/issues/1134


drewr95 (approved):
Thank you very much for doing this!

serges147:
I believe that `etl::optional` has similar deficiencies, namely lack of "nothrow"-s , including conditional one (see constructor # 4 at https://en.cppreference.com/w/cpp/utility/optional/optional.html). Similar for move assignment.

rolandreichweinbmw:
> I believe that `etl::optional` has similar deficiencies, namely lack of "nothrow"-s , including conditional one (see constructor # 4 at https://en.cppreference.com/w/cpp/utility/optional/optional.html). Similar for move assignment.

I just added another commit that handles etl::optional also.

include/etl/optional.h
```diff
@@ -146,7 +146,7 @@ namespace etl
       /// Copy constructor.
       //***************************************************************************
       ETL_CONSTEXPR20_STL
-      optional_impl(const optional_impl<T>& other)
+      optional_impl(const optional_impl<T>& other) ETL_NOEXCEPT_IF(etl::is_nothrow_copy_constructible<T>::value)
```
drewr95, line 149: I've seen this repeated elsewhere, could we possibly add a:

`ETL_NOEXCEPT_IF_NO_THROW_COPYABLE(T)` and
`ETL_NOEXCEPT_IF_NO_THROW_MOVABLE(T)` ?

drewr95 (approved):

include/etl/optional.h
rolandreichweinbmw: While this sounds reasonable and could shorten those lines, I try to limit the introduction of new macros if I can use the existing, more general ones, and do the rest via C++ means.

In this case reading `ETL_NOEXCEPT_IF_NO_THROW_COPYABLE` would be an oversimplification of `ETL_NOEXCEPT_IF(etl::is_nothrow_copy_constructible<T>::value)`

include/etl/optional.h
```diff
@@ -230,7 +231,7 @@ namespace etl
       /// Assignment operator from optional_impl.
       //***************************************************************************
       ETL_CONSTEXPR20_STL
-      optional_impl& operator=(const optional_impl<T>& other)
+      optional_impl& operator=(const optional_impl<T>& other) ETL_NOEXCEPT_IF(etl::is_nothrow_copy_constructible<T>::value)
```
serges147, line 234: Should it be something like `etl::is_nothrow_copy_assignable<T>::value` ?
Similar for move assignment.

include/etl/optional.h
```diff
@@ -277,7 +278,7 @@ namespace etl
       template <typename U,
                 typename etl::enable_if< etl::is_constructible<T, U&&>::value && !etl::is_same<typename etl::decay<U>::type, optional_impl>::value,
                                          int>::type = 0>
-      ETL_CONSTEXPR20_STL optional_impl& operator=(U&& value_)
+      ETL_CONSTEXPR20_STL optional_impl& operator=(U&& value_) ETL_NOEXCEPT_IF((etl::is_nothrow_constructible<T, U&&>::value))
```
serges147, line 281: double `((` b/c of some reason?

rolandreichweinbmw: Yes, preprocessor macros need this because with the contained comma, it misunderstands as 2 arguments where only one is meant here.

include/etl/optional.h
```diff
@@ -1580,7 +1583,7 @@ namespace etl
                                           && !etl::is_same<typename etl::decay<U>::type, etl::nullopt_t>::value
                                           && !etl::is_pod<typename etl::remove_cv<T>::type>::value,
                                         int>::type = 0>
-    ETL_CONSTEXPR20_STL optional& operator=(U&& value_)
+    ETL_CONSTEXPR20_STL optional& operator=(U&& value_) ETL_NOEXCEPT_IF((etl::is_nothrow_constructible<T, U&&>::value))
```
serges147, line 1586: I would also try to test these changes. And probably try to `static_assert` in the tests that `etl::is_nothrow_copy/move_constructible<etl::optional<T>>` depending on the same property of `T`.

include/etl/optional.h
rolandreichweinbmw: The implemented operator is assignment, that's correct! However, the implementation actually does copy construction (l.240), hence `etl::is_nothrow_copy_constructible`.

include/etl/optional.h
rolandreichweinbmw: Good point. Added respective tests.

drewr95 (approved):

test/test_type_traits.cpp
```diff
@@ -1490,6 +1490,134 @@ namespace
...
 
+    //*************************************************************************
+    TEST(test_is_nothrow_constructible)
+    {
+#if defined(ETL_USE_TYPE_TRAITS_BUILTINS) || (ETL_USING_STL && !defined(ETL_USER_DEFINED_TYPE_TRAITS))
+  #if ETL_USING_CPP17
+      CHECK((etl::is_nothrow_constructible_v<int>) == (std::is_nothrow_constructible_v<int>));
+      CHECK((etl::is_nothrow_constructible_v<Copyable>) == (std::is_nothrow_constructible_v<Copyable>));
+      CHECK((etl::is_nothrow_constructible_v<Moveable>) == (std::is_nothrow_constructible_v<Moveable>));
+      CHECK((etl::is_nothrow_constructible_v<MoveableCopyable>) == (std::is_nothrow_constructible_v<MoveableCopyable>));
```
serges147, line 1509: @rolandreichweinbmw Roland, what concerns me here is that we test whether it's "nothrow" or potentially throwable using already existing `Copyable`, `Movable` etc. , which are not really about whether they are fallible or infallible, but rather about presence of copy|move operations. So, verification of ETL against STL is definitely good one, but IMHO there is no clear picture that we're covering both cases (`false == false` and `true == true`).

What if make all these `Copyable`|`Movable`|`MoveableCopyable` completely "fallible" (without `noexcept` at all), but also introduce `InfallibleCopyable`|`InfallibleMovable`|`InfallibleMoveableCopyable` and check ETL vs STD using these "infallible" types as well. Like this e.g.:
```
    ...
      CHECK((etl::is_nothrow_constructible<int>::value) == (std::is_nothrow_constructible<int>::value));
      CHECK((etl::is_nothrow_constructible<Copyable>::value) == (std::is_nothrow_constructible<Copyable>::value));
      CHECK((etl::is_nothrow_constructible<Moveable>::value) == (std::is_nothrow_constructible<Moveable>::value));
      CHECK((etl::is_nothrow_constructible<MoveableCopyable>::value) == (std::is_nothrow_constructible<MoveableCopyable>::value));
      CHECK((etl::is_nothrow_constructible<InfallibleCopyable>::value) == (std::is_nothrow_constructible<InfallibleCopyable>::value));
      CHECK((etl::is_nothrow_constructible<InfallibleMoveable>::value) == (std::is_nothrow_constructible<InfallibleMoveable>::value));
      CHECK((etl::is_nothrow_constructible<InfallibleMoveableCopyable>::value) == (std::is_nothrow_constructible<InfallibleMoveableCopyable>::value));
```

serges147: > ..., but also introduce `InfallibleCopyable`|`InfallibleMovable`|`InfallibleMoveableCopyable` and check ETL vs STD using these "infallible" types as well. Like this e.g.:

or make 
```
template <bool Infallible = false>
class MoveableCopyable {
   ...
   MoveableCopyable(const MoveableCopyable&) nothrow(Infallible) {}
   MoveableCopyable(MoveableCopyable&&) nothrow(Infallible) {}
   ...
};
```

rolandreichweinbmw: @serges147 thank you for reviewing! I agree that testing explicit true/false cases was missing. Therefore, I added explicit comparisons against those values. (I prefer this low-profile approach in this PR.) Besides, made some test lines to be run unconditionally where possible.

Regarding the InfallibleXYZ approach, I propose doing this on a more general level, since the pattern is already established far before line 1494. Therefore, I propose a separate PR for this kind of generalization.

serges147: I closely inspected all these new explicit comparisons, almost all look good, but not at lines 1582 & 1626. I believe it's b/c of mistake at line 106 (missing second `&`).

serges147 (changes requested):

serges147:
Searching for `(\w+)&\s+operator\s*=\s*\(\s*\1&[^&]` revealed the same bug at `NotDefaultConstructible`.

test/test_type_traits.cpp
rolandreichweinbmw: Good catch! Thanks for reporting.

This also uncovered a previously present bug in test code: is_assignable<> and is_move_assignable<> both need to be false_type, not true_type because of the deleted move assignments.

rolandreichweinbmw:
> Searching for `(\w+)&\s+operator\s*=\s*\(\s*\1&[^&]` revealed the same bug at `NotDefaultConstructible`.

Right! Fixing this one along, as well.

serges147:
@rolandreichweinbmw Roland, could you please explain me what is the use case for the below code:
```
...
// Definitions for when the STL and compiler built-ins are not available.
#if ETL_NOT_USING_STL && !defined(ETL_USE_TYPE_TRAITS_BUILTINS)

using etl::is_assignable;
using etl::is_constructible;
using etl::is_copy_assignable;
using etl::is_copy_constructible;
using etl::is_move_assignable;
using etl::is_move_constructible;

//*************************
template <>
struct etl::is_assignable<Copyable, Copyable> : public etl::true_type
{
};
...
```
I see `#if` guard, but I'm not sure in which environment this could happen? E.g. do we have it in CI running?
Also, I wonder should it be actually with lvalue ref (like `struct etl::is_assignable<Copyable&, Copyable>`) b/c this is how we use it at `test_is_nothrow_assignable`.

serges147 (approved):
Just one final "wonder" question ;)

rolandreichweinbmw:
> @rolandreichweinbmw Roland, could you please explain me what is the use case for the below code:
> I see `#if` guard, but I'm not sure in which environment this could happen? E.g. do we have it in CI running?
> ...

Yes, this is a standard test case: Both in test/run-tests.sh and in the CI scripts (e.g. .github/workflows/gcc-c++17.yml), one of the tested cases is always NO_STL. At the same time, having type traits builtins from the compiler is considered a special case. This luxury is available with clang and gcc only. When using other compilers, be prepared that you can't even try to use them. In this case, the user (as demonstrated in the test) needs to manually specialize some traits for their types (depending on use).

And of course, the tests would not work otherwise.

> Also, I wonder should it be actually with lvalue ref (like `struct etl::is_assignable<Copyable&, Copyable>`) b/c this is how we use it at `test_is_nothrow_assignable`.

While is_assignable<Copyable&, Copyable> is the more typical use case covered by test_is_nothrow_assignable, the case in test_is_assignable with is_assignable_v<Copyable, Copyable> is actually different. Maybe even less useful: assigning rvalue to rvalue? But this is clearly outside the scope of this PR - maybe ask the author or propose a refactoring.

Anyway, the two locations are in mutually exclusive cases: NO_STL+no type traits vs. test_is_nothrow_assignable guards for the opposite.

PR #1389, opened 2026-04-08
Suppress false positive compiler warnings when compiling with -O3

rolandreichweinbmw:
The CI checks currently only check everything with -O0. Wenn activating higher optimization levels, more warnings kick in. Leading to errors, depending on the configuration.


rolandreichweinbmw:
How about adding -O3 also in the CI checks? Since this consumes more CPU resources, at least for one of the builds, e.g. C++23, since coverage is highest in the higher C++ versions.

jwellbelove:
> How about adding -O3 also in the CI checks? Since this consumes more CPU resources, at least for one of the builds, e.g. C++23, since coverage is highest in the higher C++ versions.

Yes, I occasionally run with -O2 and it does take quite some time to complete, even on my i7 32GB machine.

rolandreichweinbmw:
I just added an additional commit in this branch to use -O3 in:

    * .github/workflows/clang-c++23.yml
    * .github/workflows/gcc-c++23.yml

rolandreichweinbmw:
> I just added an additional commit in this branch to use -O3 in:
> 
> ```
> * .github/workflows/clang-c++23.yml
> * .github/workflows/gcc-c++23.yml
> ```

Removing for now. There are several open PRs right now that together fix related warnings.

After being merged, will add this separately.

jwellbelove (approved):

PR #1392, opened 2026-04-08
Kick-off use of etl algorithms to replace the raw loops in byte_stream

jputcu:
I'm trying to replace my own equivalent class but missing some features, so I would like to get my feet wet starting small.

### DIFF
--- a/include/etl/byte_stream.h
+++ b/include/etl/byte_stream.h
@@ -177,13 +177,7 @@ namespace etl
     template <typename T>
     typename etl::enable_if<etl::is_integral<T>::value || etl::is_floating_point<T>::value, void>::type write_unchecked(const etl::span<T>& range)
     {
-      typename etl::span<T>::iterator itr = range.begin();
-
-      while (itr != range.end())
-      {
-        to_bytes(*itr);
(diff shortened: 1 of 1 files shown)


include/etl/byte_stream.h
```diff
@@ -177,13 +177,7 @@ namespace etl
...
     typename etl::enable_if<etl::is_integral<T>::value || etl::is_floating_point<T>::value, void>::type write_unchecked(const etl::span<T>& range)
     {
-      typename etl::span<T>::iterator itr = range.begin();
-
-      while (itr != range.end())
-      {
-        to_bytes(*itr);
-        ++itr;
-      }
+      etl::for_each(range.cbegin(), range.cend(), [this](T const& t) { to_bytes(t); });
```
rolandreichweinbmw, line 180: While this is actually the right thing to do in application code, ETL itself is not supposed to change for compatibility reasons. The CI checks also reported this.

jputcu: I'm so used to modern C++ I was indeed confronted with this reality and seeing if it indeed is worth it.

PR #1394, opened 2026-04-09
Fix red unit tests on 32 bits big-endian platform.

serges147:
Discovered on qemu-armeb run.

### DIFF
--- a/test/test_hash.cpp
+++ b/test/test_hash.cpp
@@ -149,7 +149,14 @@ namespace
 
       if (ETL_PLATFORM_32BIT)
       {
-        CHECK_EQUAL(0xEC6A8D69UL, hash);
+        if (etl::endianness::value() == etl::endian::little)
+        {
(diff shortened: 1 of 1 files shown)


rolandreichweinbmw (approved):

rolandreichweinbmw:
For your info: I added this commit to https://github.com/ETLCPP/etl/pull/1380 to fix the resulting compile error now that we are actually testing 32 bit big endian also.

serges147:
> For your info: I added this commit to #1380 to fix the resulting compile error now that we are actually testing 32 bit big endian also.

Should I close this PR then?

And thanks Roland for all the work in the #1380 !

rolandreichweinbmw:
> > For your info: I added this commit to #1380 to fix the resulting compile error now that we are actually testing 32 bit big endian also.
> 
> Should I close this PR then?

Yes, please. Let's track it via #1380.

PR #1396, opened 2026-04-13
Fix chrono.h year_month_weekday_last and year_month_weekday sysdays()

rolandreichweinbmw:
Bug 1: year_month_weekday_last::operator sys_days() — wrong weekday construction

The code was constructing a weekday from a raw day count using weekday(unsigned), which treats the value as a weekday encoding (0–6). The fix uses weekday(sys_days), which correctly accounts for the epoch being a Thursday (+4 offset).

Bug 2: year_month_weekday::operator sys_days() — same wrong weekday construction + off-by-one in day_of_month

Same weekday(unsigned) vs weekday(sys_days) issue. Additionally, the day_of_month calculation was missing the 1 + base — it computed a 0-based offset from day 1, but forgot to add the 1 back when converting to an actual day number.


denravonska:
These are the same fixes I had when testing locally.

drewr95 (approved):

jwellbelove (approved):

PR #1409, opened 2026-04-24
Add test/run-clang-tidy.sh

rolandreichweinbmw:
This is a further effort to add static code analysis which is necessary in safety context.

Also, add .clang-tidy configuration file.

In test/syntax_check/CMakeLists.txt, make ETL headers non-system headers.

Added .github/workflow/clang-tidy.yaml. Does not break the build for now on clang-tidy findings/warnings.


.github/workflows/clang-tidy.yaml
```diff
@@ -0,0 +1,27 @@
+name: clang-tidy
+
+on:
+  push:
+    branches: [ master, development, pull-request/* ]
+  pull_request:
+    branches: [ master, development, pull-request/* ]
+    types: [opened, synchronize, reopened]
```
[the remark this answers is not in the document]
rolandreichweinbmw: It's intentional.

.clang-tidy
```diff
@@ -0,0 +1,182 @@
...
+    clang-analyzer-*,
+    bugprone-*,
+    -bugprone-easily-swappable-parameters,
+    cppcoreguidelines-pro-type-vararg,
+    cppcoreguidelines-pro-type-reinterpret-cast,
+    llvm-*,-llvm-header-guard,-llvm-include-order,
+    google-readability-casting
+ExtraArgs: ['-Wno-unknown-warning-option']
+HeaderFileExtensions: ['h', 'hpp']
+HeaderFilterRegex: '^(?!.*(UnitTest|test|mock)).*'
```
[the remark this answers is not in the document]
rolandreichweinbmw: Supporting old clang-tidy now.

test/syntax_check/CMakeLists.txt
```diff
@@ -30,7 +30,7 @@ endif()
 add_library(tests OBJECT)
 target_compile_definitions(tests PRIVATE __STDC_LIMIT_MACROS __STDC_CONSTANT_MACROS __STDC_FORMAT_MACROS)
 target_include_directories(tests PRIVATE "")
-target_include_directories(tests SYSTEM PRIVATE ../../include)
+target_include_directories(tests PRIVATE ../../include)
```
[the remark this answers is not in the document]
rolandreichweinbmw: It's intentional

drewr95 (approved):
Looks really good. There is also a lightweight static analysis option for MSVC as well `/analyze`

include/etl/message_timer.h
```diff
@@ -60,8 +60,8 @@ SOFTWARE.
   #endif
 
   #if defined(ETL_MESSAGE_TIMER_USE_ATOMIC_LOCK)
-    #define ETL_DISABLE_TIMER_UPDATES (++process_semaphore)
-    #define ETL_ENABLE_TIMER_UPDATES  (--process_semaphore)
+    #define ETL_DISABLE_TIMER_UPDATES (++process_semaphore);
```
jwellbelove: I did not add trailing `;` to the macro definitions as the IDE auto formatting does not see the macro as a separate statement and indents it incorrectly.
It also doesn't match the use of other macros in the ETL code. e.g. `ETL_ASSERT`

rolandreichweinbmw: Yes, I understood that you previously left out the `;` intentionally.

From clang-tidy perspective, they are missing. :-)

I hope that current IDEs will handle the macros correctly with `;`. Otherwise, `;` seems to be widely accepted after parameterized macros.

Also, `ETL_ASSERT()` is typically ended with a `;`? So in this regard, consistently using `;` seems reasonable?

jwellbelove: It would be consistent with the usage of the majority of other macros if the `;` was not part of the definition.
Could you use the `(void)` cast trick to keep clang-tidy happy?

rolandreichweinbmw: With `((void)0)` it works. Doing so now. Thanks for the hint!

jwellbelove (approved):

PR #1410, opened 2026-04-24
Add begin() and end() to etl::expected

### LINKED ISSUE #1243: Add begin() and end() to etl::expected
Match functionality with C++26 std::expected

rolandreichweinbmw:
Fixes #1243


drewr95 (approved):

include/etl/expected.h
```diff
@@ -725,6 +725,39 @@ namespace etl
...
+      return has_value() ? &etl::get<value_type>(storage) : ETL_NULLPTR;
+    }
+
+    //*******************************************
+    /// Returns a const pointer past the value if has_value(), otherwise returns nullptr.
+    //*******************************************
+    ETL_NODISCARD ETL_CONSTEXPR14 const value_type* end() const ETL_NOEXCEPT
+    {
+      return has_value() ? &etl::get<value_type>(storage) + 1 : ETL_NULLPTR;
+    }
```
serges147, line 807: Should we also have`cbegin` & `cend` versions?

rolandreichweinbmw: I think that the const overloads are sufficient, comparing with other interfaces like etl::optional / std::optional.

However, now that you mentioned it :-) in #1243, @jwellbelove proposed "Match functionality with C++26 std::expected". But std::expected doesn't have begin() and end().

Anyway, keeping the PR's change of adding them, as an ETL specific extension. However, I found that error_or() as defined in std was missing. Now also adding it.

jwellbelove (approved):

PR #1414, opened 2026-04-25
Swap can be part of the etl namespace and let ADL do its thing

jputcu:
Same as other recent code improvements: cleaning up the global namespace


rolandreichweinbmw (approved):

serges147:
Yep, I'm in the process to fix the same issue for `swap(etl::optional<T>&, etl::optional<T>&)` (in addition to other issues of the optional).

drewr95 (approved):

jwellbelove (approved):

PR #1427, opened 2026-05-03
Remove dead code

rolandreichweinbmw:
Removing private class members, code unused by ETL in "private"
namespaces, code unreachable via preprocessor guards (C++11 inside
C++03).

For code still to be kept, even though unused at first sight, add
tests.

Note: Based on development branch, to prevent conflicts.


jwellbelove (dismissed):

rolandreichweinbmw:
It was not easy to provide a test for the not-unused wipe_on_destruct. But it was worth it. :-)

jwellbelove (approved):

PR #1387, opened 2026-04-06
Rename _current to _current_it in ranges.h

rolandreichweinbmw:
Resolves conflict with _current macro from Zephyr and improves self-explanation of variable.

Issue #1386, opened 2026-04-06
ranges iterator member _current clashes with zephyr _current macro


benedekkupper:
Zephyr RTOS defines a `_current` macro in [`include/zephyr/kernel_structs.h`](https://github.com/zephyrproject-rtos/zephyr/blob/149c8b1758a8ac6bd20ce230d2b1ae43c483ef30/include/zephyr/kernel_structs.h#L260), which collides with the naming of `_current` member variable of various iterators in `etl/ranges.h`

benedekkupper:
@jwellbelove to avoid similar issues in the future, it would make sense to add a zephyr build to the CI, did this come up before?

jwellbelove:
I'm fine with adding that, if someone wants to do a PR for it.

jwellbelove:
Fixed 20.48.0

PR #1437, opened 2026-05-12
ranges: Fix confict with Zephyr preprocessor definition

surligas:
Classic naming pollution through the preprocessor! 

Zephyr defines the `_current` as a preprocessor macro (https://github.com/zephyrproject-rtos/zephyr/blob/f0ce29dd53bc4d4271bd13113691807cd55ea55f/include/zephyr/kernel_structs.h#L256). In the `etl/ranges.h` file, the `_current` is used as a member in various places. This causes a conflict and prevents the code from compiling.
This commit fixes this issue by renaming all members named `_current` to `_cur`.

I understand that this is not an issue of ETL, but Zephyr, so feel free to reject this PR unconditionally :D

### DIFF
--- a/include/etl/ranges.h
+++ b/include/etl/ranges.h
@@ -2975,7 +2975,7 @@ namespace etl
         iterator() = default;
 
         iterator(const_iterator_type current, const_iterator_type segment_end, bool is_end)
-          : _current(current)
+          : _cur(current)
           , _segment_end(segment_end)
           , _is_end(is_end || (current == segment_end))
         {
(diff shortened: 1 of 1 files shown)


rolandreichweinbmw:
Unfortunately, this PR is duplicated effort, since the issue is already fixed in the development branch.

See also issue #1386 and PR #1387.

surligas:
Aaahh perfect then! I didn't checked the development branch. I am closing this one. Thanks!

Issue #1390, opened 2026-04-08
etl::variant uses too much flash size


jputcu:
Since some years I've been using my own, naive, implementation of `std::expected` for my Atmega128. I really wanted to use the `etl::expected`, so I have an alias to swap implements, however, doing so makes my code explode: **77kb becomes 83kb**. (Note: I'm currently not using `lto`)

So starting small I just replaced the use in my smallest library: **+600b**.

As an experiment, I made a copy of `expected.h` but replaced `etl::variant` with `std::variant` which is provided by https://github.com/modm-io/avr-libstdcpp.

Result, just **+10b**.

So somehow `etl::variant` generates much more code than `std::variant`.

I recall I had the same experience many years ago, wanting to use `etl::variant`. Due to the huge code size increase I kept using `etl::legacy::variant` until I switched to `std::variant`.

rolandreichweinbmw:
The root cause is that `etl::variant` stores a `void(*)(int, char*, const char*)` function pointer for copy/move/destroy, forcing the compiler to emit and retain separate non-inlinable function bodies for every type alternative. A `std::variant`-style approach that uses an index + switch at the call site would let the compiler see all the code paths, enabling inlining, dead-code elimination, and constant folding — dramatically reducing code size especially on small targets without LTO.

Maybe we should reconsider this for `etl::variant`?

jwellbelove:
There may also be some scope for moving non-type dependent functionality into helper classes.

drewr95:
I've noticed this as well while using it. Would like to decrease flash usage.

rolandreichweinbmw:
Can you confirm that https://github.com/ETLCPP/etl/pull/1441 solves the issue?

jputcu:
@rolandreichweinbmw I replaced `expected.h` and `type_traits.h` from your other pull request (for the missing type traits) with yours and can confirm the flash usage dropped 500 byte. My unit tests are also still ok.

jputcu:
@rolandreichweinbmw My previous comment compared it to my copy of `etl/expected.h`, where I replace `etl::variant` with `std::variant`. When I compare it to the unchanged `etl/expected.h` I saw a reduction of 3k (for a firmware of 80k), so a big improvement!

jwellbelove:
Fixed 20.48.0

Issue #1310, opened 2026-02-21
Request: Make variant capable of ROM placement


drewr95:
I have some use cases where it would be desirable to store a variant in ROM like I can with the STL.  Currently I have to resort back to using tagged unions and NOLINT all of the union access.

I think this would require possibility any of the following:
1. If the types are all trivially destructible then you do not have to implement the destructor, or maybe just not call the destructor on the byte buffer.
2. Make the destructor constexpr

Current I can do this (from my unittests which are C++20 but application uses C++14):
`constexpr std::variant<int, float> var{42};`

However the following fails to compile:
`constexpr etl::variant<int, float> var{42};`

> [build] _unittests.cpp:50:36: error: constexpr variable cannot have non-literal type 'const etl::variant<int, float>'
> [build]    50 | constexpr etl::variant<int, float> dummyVariant{42};
> [build]       |                                    ^
> [build] /home/ubuntu/boom_ground/build_clang/_deps/etl-src/include/etl/private/variant_variadic.h:609:5: note: 'variant<int, float>' is not literal because its destructor is not constexpr
> [build]   609 |     ~variant()
> [build]       |     ^
> [build] 1 error generated.

jwellbelove:
`etl::variant` cannot be `constexpr` as it uses placement-new to initialise the elements.
In theory it could be `constexpr` if you are using C++20 and the STL, as `std::construct_at` is `constexpr`.
Internally it uses placement-new but the compiler gives the STL special status to do things us mere mortals are not allowed to do.

It should also be possible the specialise `etl::variant` for trivially constructible/destructible types, thus removing the need for placement-new in those circumstances, though any solution would require a union of the types, rather than a char buffer as storage, to avoid using a non-`constexpr` compatible `reinterpret_cast`.

I managed to convince some compilers that they had a `std::initializer_list` by creating a compatible clone in the `std` namespace. I did try a similar experiment with `std::construct_at` some time ago, with mixed results.

drewr95:
That was my initial research too, that specialization would have to be done for trivially destructible types. 

Do you think this would be a big effort? At least the union storage would be isolated in one place instead of using unions for any ROM placement. 

Thank you

jwellbelove:
I think a union solution may have to be implemented quite differently to the current version.
I'm not sure yet how a variadic union would be implemented. I think it would have to be some sort of recursive definition.

A bit like how `etl::type_list` and `etl::tuple` are defined.

[...]

PR #1441, opened 2026-05-16
Make etl::variant capable for ROM placement and optimize runtime size

rolandreichweinbmw:
Via a variadic_union, the internal storage is adjusted to be able to be constexpr.

This PR addresses https://github.com/ETLCPP/etl/issues/1310 and https://github.com/ETLCPP/etl/issues/1390

### DIFF
--- a/include/etl/private/variant_variadic.h
+++ b/include/etl/private/variant_variadic.h
@@ -47,6 +47,7 @@ SOFTWARE.
 #include "../visitor.h"
 
 #include <stdint.h>
+#include <string.h>
 
 #if defined(ETL_COMPILER_KEIL)
   #pragma diag_suppress 940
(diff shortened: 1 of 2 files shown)


drewr95 (approved):
Wow that is crazy! Thank you so much for implementing this!

Is this c++14 compliant? I just wanted to make sure but it looks like it. 

We heavily use variants and I am excited to get rid of our tagged unions in ROM :)

rolandreichweinbmw:
> Is this c++14 compliant? I just wanted to make sure but it looks like it.

Good question. For historical reasons, variant_variadic.h is C++11 and the tests do C++14 only. But since this is only needed in 2 places for lamda auto parameters, I just adjusted the tests to C++11 (except for the 2 auto tests), so the variant is now available even more widely.

jwellbelove (approved):

Issue #1457, opened 2026-06-08
FR: priority_queue needs a non-sized base class like other containers do


positron96:
Other queue classes have base classes that are not dependent on size. They are very useful, e.g. to make functions that take these containers as arguments, but are decoupled from decisions of the size of the containers (i.e. they can be library functions while the project choses container size).
In general, they help to decouple interfaces from implementation details.

For example, iqueue or ideque are such non-sized templates. Here are their class hierarchies:
```
template <size_t MEMORY_MODEL = etl::memory_model::MEMORY_MODEL_LARGE>
class queue_base;

template <typename T, const size_t MEMORY_MODEL = etl::memory_model::MEMORY_MODEL_LARGE>
class iqueue : public etl::queue_base<MEMORY_MODEL>;

template <typename T, const size_t SIZE, const size_t MEMORY_MODEL = etl::memory_model::MEMORY_MODEL_LARGE>
class queue : public etl::iqueue<T, MEMORY_MODEL>
```

```
class deque_base;

template <typename T>
class ideque : public etl::deque_base;

template <typename T, const size_t MAX_SIZE_>
class deque : public etl::ideque<T>;
```

A `etl::priority_queue` lacks such a base class. It extends ipriority_queue, but that must be parametrized by an underlying container type, which contains size and in general is an implementation-specific detail.

```
template <typename T, typename TContainer, typename TCompare = etl::less<T> >
class ipriority_queue;

template <typename T, const size_t SIZE, typename TContainer = etl::vector<T, SIZE>,
          typename TCompare = etl::less<typename TContainer::value_type> >
class priority_queue : public etl::ipriority_queue<T, TContainer, TCompare>;
```

It would be very useful for priority_queue to have such an abstract base class that does not depend on queue size (and container type for that matter). In fact, it would be nice for priority queue to implement an interface of a queue, but it doesn't look like it aligns with a matching std:: equivalent.

rolandreichweinbmw:
Like this?

https://github.com/rolandreichweinbmw/etl/tree/priority-queue-base-class

(See last commit, on top of current development branch.)

positron96:
Yes, I think that about does it!

jwellbelove:
Implemented 20.48.0

positron96:
Thanks for a great job!

PR #1459, opened 2026-06-11
 Make ipriority_queue non-sized base class

rolandreichweinbmw:
Fixes issue #1457

Move from operator[] access to operator+ and operator- to prevent
limited compatibility. This worked before for vector because of
iterators being ptrdiff_t (signed), but not for unsigned access
like size_t as in etl::queue.


include/etl/algorithm.h
```diff
@@ -880,14 +880,14 @@ namespace etl
       TDistance parent = (value_index - 1) / 2;
 
 #include "etl/private/diagnostic_array_bounds_push.h"
-      while ((value_index > top_index) && compare(first[parent], value))
+      while ((value_index > top_index) && compare(*(first + parent), value))
```
jwellbelove, line 883: Was this giving you warnings?

rolandreichweinbmw: Yes, exactly: It warns about unsigned vs. signed indexing. E.g. argument signed while operator signature takes size_t (unsigned).

This was discovered by using deque instead of vector, since here we have a "real" operator (with size_t index) instead of just using ptrdiff_t as iterator.

rolandreichweinbmw: It's also described in the commit message.

jwellbelove (approved):

PR #1462, opened 2026-06-15
Add emplace() to etl::map and further containers

### LINKED ISSUE #961: emplace for etl::map
Hello!

I just noticed that there is no support for `emplace` on `etl::map` like https://en.cppreference.com/w/cpp/container/map/emplace.
However `etl::vector` supports `emplace`.
Is there a specific reason for not supporting this on `etl::map` or is it 'just not done(, yet)'?

rolandreichweinbmw:
Fixes #961

### DIFF
--- a/include/etl/map.h
+++ b/include/etl/map.h
@@ -1319,6 +1319,188 @@ namespace etl
       }
     }
 
+#if ETL_USING_CPP11 && ETL_NOT_USING_STLPORT
+    //*********************************************************************
+    /// Emplaces a value to the map.
+    /// Constructs the value_type in place from the given arguments.
+    //*********************************************************************
+    template <typename... Args>
+    ETL_OR_STD::pair<iterator, bool> emplace(Args&&... args)
+    {
+      ETL_ASSERT(!full(), ETL_ERROR(map_full));
+
+      // Get next available free node
+      Data_Node& node = allocate_data_node_from_args(etl::forward<Args>(args)...);
+
+      // Obtain the inserted node (might be ETL_NULLPTR if node was a duplicate)
+      Node* inserted_node = insert_node(root_node, node);
+      bool  inserted      = inserted_node == &node;
+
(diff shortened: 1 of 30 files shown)


include/etl/map.h
```diff
@@ -1319,6 +1319,188 @@ namespace etl
       }
     }
 
+#if ETL_USING_CPP11 && ETL_NOT_USING_STLPORT
```
drewr95, line 1322: I'm curious why there is an additional check for not using the STL port?  Why should this method not exist if using c++11 but was also using the STL port in conjunction?  If this guard is going to exist I'm not sure there is a need to call `ETL_OR_STD::` macro instead of just directly invoking `etl`.  If it's preferred to use `ETL_OR_STD` for implementation choice, shouldn't it also be done as well on the below line?

`ETL_OR_STD::forward<Args>(args)...)`?

Thank you,

drewr95 (approved):

include/etl/map.h
rolandreichweinbmw: STLport is a legacy underlying STL C++ library, besides libc++ and libstdc++. Still supported in ETL but limited support for variadic templates or perfect forwarding. Typical pattern in ETL to support either modern C++ (11+) or not.

Don't confuse with using STD in ETL.

drewr95: I completely forgot about that. Thanks!

drewr95 (approved):

drewr95 (approved):

jwellbelove:
There are also 
```
flat_map
flat_multimap
flat_set
flat_multiset
```
```
reference_flat_map
reference_flat_multimap
reference_flat_set
reference_flat_multiset
```

rolandreichweinbmw:
> There are also
> 
> ```
> ...
> ```

Right. Just added for those also.

jwellbelove (approved):

Issue #1372, opened 2026-04-02
Stable Partition - 98


wsciaroni:
# Background

I would like to request that two stable partition algorithms are implemented to complement std::stable_partition.

std::stable_partition rearranges elements in a range so that those satisfying a predicate come before those that do not, while preserving their original relative order.

# Implementation

Implementations typically use an adaptive algorithm that chooses a strategy based on available memory:

## With Extra Memory (O(N) Time):
If enough temporary memory is available, the algorithm copies elements satisfying the predicate into a temporary buffer, then copies the remaining elements, and finally moves them back to the original range.

In the ETL this could look like a pair of iterators is passed in to the stable_partition method to be the temporary buffer.

## In-Place (O(N log N) Time):
The ETL is very explicit on how memory is managed often requiring in-place implementation. For in place behavior, a divide-and-conquer approach is used. The range is split in two; each half is partitioned recursively, and the resulting middle sections (the "false" group of the first half and the "true" group of the second half) are swapped using a rotate operation.

jwellbelove:
Implemented 20.48.0

PR #1466, opened 2026-06-17
Add etl::stable partition

jwellbelove:
Added two versions of `stable_partition` to `algorithm.h`.
O(NlogN) time, O(1) space. (In-place)
O(N) time O(N) space. (User defined external buffer)

Added etl::`partition_move`.

Updated the tests and documentation.

Implements #1372

### DIFF
--- a/docs/utilities/algorithms.md
+++ b/docs/utilities/algorithms.md
@@ -940,6 +940,61 @@ bool is_permutation(TIterator1       begin1,
 Returns `true` if there exists a permutation of the elements in the range [`begin1`, `end1`) that makes that range equal to the second range.  
 Uses `predicate` to checks elements in the range.
 
+## partition
+```cpp
+ETL_CONSTEXPR14 TIterator partition(TIterator first, TIterator last, TPredicate predicate)
+```
+**Description**  
+Reorders elements in a range based on a predicate.  
+It moves all elements that satisfy the predicate to the front, and those that do not to the back.
+Does not guarantee to keep the original relative order.  
+
+**Return**  
+An iterator to the start of the group that do not satisfy the predicate.
+
+## stable_partition
+
+### In-place
+```cpp
+template <typename TIterator, typename TPredicate>
+ETL_CONSTEXPR14 TIterator stable_partition(TIterator first, 
+                                           TIterator last, 
+                                           TPredicate predicate)
+```
+**Description**  
+Reorders elements in a range based on a predicate.  
+It moves all elements that satisfy the predicate to the front, and those that do not to the back.
+Keeps the original relative order.  
+In-place  
+O(NlogN) time.  
+O(1) space.
+
+**Return**  
+An iterator to the start of the group that do not satisfy the predicate.
+
+### External buffer
+```cpp
+template <typename TIterator, typename TPredicate>
+ETL_CONSTEXPR14 TIterator stable_partition(TIterator first, 
+                                           TIterator last, 
+                                           TIterator buffer_first, 
+                                           TIterator buffer_last, 
+                                           TPredicate predicate)
+```
+**Description**  
+Reorders elements in a range based on a predicate.  
+It moves all elements that satisfy the predicate to the front, and those that do not to the back.
+Keeps the original relative order.  
+
+Requires a user supplied buffer that must be at least the same size as the range [`first`, `last`).  
+O(N) time.  
+O(N) space.  
+
+Raises an `etl::stable_partition_buffer_too_small` is the buffer is not large enough to accommodate the range [`first`, `last`).
+
+**Return**  
+An iterator to the start of the group that do not satisfy the predicate.
+
 ## is_partitioned
 
 ```cpp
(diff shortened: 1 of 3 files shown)


drewr95 (approved):

include/etl/algorithm.h
```diff
@@ -3531,6 +3576,83 @@ namespace etl
...
+      else
+      {
+        *buffer_false++ = etl::move(*first);
+      }
+
+      ++first;
+    }
+
+    // Move them back to the original range.
+    etl::move(buffer_first, buffer_last, input_first);
```
rolandreichweinbmw: This moves the entire buffer [[buffer_first, buffer_last) back into the source starting at input_first == first. The docs and the assert both explicitly allow a buffer larger than the range ("at least the same size as the range"). When the buffer is larger than n, this writes buffer_size - n elements past last — an out-of-bounds write into the source container.

The tests don't catch it because every test uses a buffer of exactly the same size as the data. Fix: move only n elements back, e.g.

```
TIterator buffer_end = buffer_first;
etl::advance(buffer_end, n);
etl::move(buffer_first, buffer_end, input_first);
```

test/test_algorithm.cpp
```diff
@@ -2853,18 +2853,65 @@ namespace
...
+      bool complete = false;
+
+      while (!complete)
+      {
+        auto boundary = std::stable_partition(std::begin(std_input), std::end(std_input), std::bind(std::greater<int>(), std::placeholders::_1, 4));
+        std::move(std::begin(std_input), boundary, std::begin(std_true));
+        std::move(boundary, std::end(std_input), std::begin(std_false));
+
+        etl::partition_move(std::begin(etl_input), std::end(etl_input), std::begin(etl_true), std::begin(etl_false),
+                            std::bind(std::greater<int>(), std::placeholders::_1, 4));
```
rolandreichweinbmw: The test runs on TestDataM<int> but uses std::bind(std::greater<int>(), _1, 4) as the predicate. TestDataM<int> has no operator int — only data.h:231. So each element converts TestDataM → bool → int (1), and the predicate is effectively 1 > 4, i.e. always false.

Consequences:

* Every element is classified "false", so destination_true is never exercised — the core branch of partition_move is untested.
* std::stable_partition with an always-false predicate leaves data in place, so the comparison passes trivially across all 40,320 permutations, giving false confidence.

The move-only stable_partition tests do it correctly with [](const TestDataM<int>& t){ return etl::is_even(t.value); }. The partition_move test should use the same .value-based predicate.

include/etl/algorithm.h
```diff
@@ -3531,6 +3576,83 @@ namespace etl
...
+  //***************************************************************************
+  template <typename TIterator, typename TPredicate>
+  ETL_CONSTEXPR14 TIterator stable_partition(TIterator first, TIterator last, TPredicate predicate)
+  {
+    typename etl::iterator_traits<TIterator>::difference_type n = etl::distance(first, last);
+
+    if (n <= 1)
+    {
+      // Empty or single-element range: trivially partitioned either way.
+      return last;
```
rolandreichweinbmw: std::stable_partition returns the start of the false group. For a one-element range where the predicate is false, that should be first, not last. It only affects top-level calls with exactly one element (the recursion discards these return values, so the main algorithm is unaffected and the 8-element tests don't expose it).

include/etl/algorithm.h
```diff
@@ -98,6 +98,22 @@ namespace etl
...
+    stable_partition_exception(string_type reason_, string_type file_, numeric_type line_)
+      : etl::exception(reason_, file_, line_)
+    {
+    }
+  };
+
+  struct stable_partition_buffer_too_small : stable_partition_exception
+  {
+    stable_partition_buffer_too_small(string_type file_, numeric_type line_)
+      : stable_partition_exception(ETL_ERROR_TEXT("stable_partiton:buffer too small", ETL_ALGORITHM_FILE_ID"A"), file_, line_)
```
rolandreichweinbmw: sp: stable_partiton -> stable_partition

include/etl/algorithm.h
```diff
@@ -2116,6 +2132,35 @@ namespace etl
     return ETL_OR_STD::pair<TDestinationTrue, TDestinationFalse>(destination_true, destination_false);
   }
 
+  //***************************************************************************
+  /// Moves the elements from the range (begin, end) to two different ranges
+  /// depending on the value returned by the predicate.<br>
+  ///< a href="http://en.cppreference.com/w/cpp/algorithm/partition_move"></a>
```
rolandreichweinbmw: This one does not exist, does it?

docs/utilities/algorithms.md
```diff
@@ -940,6 +956,68 @@ bool is_permutation(TIterator1       begin1,
...
+```
+**Description**  
+Reorders elements in a range based on a predicate.  
+It moves all elements that satisfy the predicate to the front, and those that do not to the back.  
+Keeps the original relative order.  
+In-place  
+
+**Complexity**  
+O(NlogN) time.  
+O(1) space.
```
rolandreichweinbmw: O(1) space is only true if you ignore space on the stack. I think it is O(log N) due to the recursive calls, when partitioning.

jwellbelove: Good point

drewr95 (approved):

PR #1467, opened 2026-06-18
Only enable the assign and insert member functions for class ivector<…

### LINKED ISSUE #1464: etl::vector::insert allows inserting an incompatible type
I was refactoring some code and it started segfaulting. After some debugging, I realized a made a basic type error (inserting a vector of values into a value of pointers). However, I am very surprised that this was able to compile in the first place as it seems to violate basic type rules.

Here's a minimal example:

```cpp
struct SomeStruct
{
    int foo;
};

etl::vector<SomeStruct, 4> myStructs{{1}, {2}, {3}, {4}};

etl::vector<SomeStruct *, 4> structPointers;
structPointers.insert(structPointers.end(), myStructs.begin(), myStructs.end());
printf("structPointers[0] = %p\n", structPointers[0]);
```

This _compiles_, and on my 64-bit desktop prints
```
structPointers[0] = 0000000200000001
```

If I change it to instead use std::vector, we get the expected compile error:

```
C:/msys64/ucrt64/include/c++/15.2.0/bits/stl_algobase.h:407:16: error: cannot convert 'SomeStruct' to 'SomeStruct*' in assignment
  407 |         *__out = *__in;
      |         ~~~~~~~^~~~~~~
```

It looks like the issue here is that the underlying insert implementation (`etl::pvoidvector::insert`) erases the type information of the source iterators and just treats them as `void **` (rather than, in this case, SomeStruct *). This leads to it basically memcpy-ing the SomeStruct entries into the destination vector.

I think the best fix here would be to add some checks in the upper level class, `etl::ivector<T*>`, that the iterators passed to `insert()` (and probably other functions that accept iterators) are compatible with the type of the vector's elements.

jwellbelove:
Only enable the `assign` and `insert` member functions for `class ivector<T*>` & `class ivector<const T*>` when the `value_type` the iterator references is a pointer to `T`.

Fixes #1464

### DIFF
--- a/include/etl/private/ivectorpointer.h
+++ b/include/etl/private/ivectorpointer.h
@@ -70,6 +70,13 @@ namespace etl
 
     typedef pvoidvector base_t;
 
+    template <typename TIterator>
+    struct is_compatible_iterator
+      : etl::integral_constant<bool, etl::is_pointer<typename etl::iterator_traits<TIterator>::value_type>::value
+                                       && etl::is_convertible<typename etl::iterator_traits<TIterator>::value_type, value_type>::value>
+    {
+    };
+
   public:
 
     //*********************************************************************
@@ -320,10 +327,7 @@ namespace etl
     ///\param last  The iterator to the last element + 1.
     //*********************************************************************
     template <typename TIterator>
-    typename etl::enable_if< etl::is_pointer<typename etl::iterator_traits<TIterator>::value_type>::value
-                               && etl::is_convertible<typename etl::iterator_traits<TIterator>::value_type, value_type>::value,
-                             void>::type
-      assign(TIterator first, TIterator last)
+    typename etl::enable_if< is_compatible_iterator<TIterator>::value, void>::type assign(TIterator first, TIterator last)
     {
       base_t::assign(first, last);
     }
@@ -444,10 +448,7 @@ namespace etl
     ///\param last     The last + 1 element to add.
     //*********************************************************************
     template <typename TIterator>
-    typename etl::enable_if< etl::is_pointer<typename etl::iterator_traits<TIterator>::value_type>::value
-                               && etl::is_convertible<typename etl::iterator_traits<TIterator>::value_type, value_type>::value,
-                             void>::type
-      insert(const_iterator position, TIterator first, TIterator last)
+    typename etl::enable_if< is_compatible_iterator<TIterator>::value, void>::type insert(const_iterator position, TIterator first, TIterator last)
     {
       base_t::insert(base_t::iterator(position), first, last);
     }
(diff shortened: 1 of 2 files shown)


rolandreichweinbmw:
@coderabbit full review

include/etl/private/ivectorpointer.h
```diff
@@ -320,7 +320,12 @@ namespace etl
     ///\param last  The iterator to the last element + 1.
     //*********************************************************************
     template <typename TIterator>
-    void assign(TIterator first, TIterator last)
+    typename etl::enable_if<
+      etl::is_pointer<typename etl::iterator_traits<TIterator>::value_type>::value
+        && etl::is_same<typename etl::remove_cv<typename etl::remove_pointer<typename etl::iterator_traits<TIterator>::value_type>::type>::type,
+                        T>::value,
+      void>::type
```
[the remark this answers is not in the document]
rolandreichweinbmw: That's a valid concern.

Can be solved like e.g.

```
template <typename TIterator>
    typename etl::enable_if<
      etl::is_pointer<typename etl::iterator_traits<TIterator>::value_type>::value
        && etl::is_same<typename etl::remove_pointer<typename etl::iterator_traits<TIterator>::value_type>::type,
                        T>::value,
      void>::type
      assign(TIterator first, TIterator last)
```

include/etl/private/ivectorpointer.h
```diff
@@ -901,9 +916,16 @@ namespace etl
...
-    void insert(const_iterator position, TIterator first, TIterator last)
+    template <typename TIterator>
+    typename etl::enable_if<
+      etl::is_pointer<typename etl::iterator_traits<TIterator>::value_type>::value
+        && etl::is_same<typename etl::remove_cv<typename etl::remove_pointer<typename etl::iterator_traits<TIterator>::value_type>::type>::type,
+                        T>::value,
+      void>::type
+      insert(const_iterator position, TIterator first, TIterator last)
     {
+      ETL_STATIC_ASSERT(etl::is_pointer<typename etl::iterator_traits<TIterator>::value_type>::value, "TIterator must dereference to a pointer type");
```
rolandreichweinbmw: This ETL_STATIC_ASSERT is redundant since the enable_if ensures it already.

If the intention was to get a better diagnostic/error message, it does not work since the whole function is already discarded by SFINAE and the message will never be printed.

jwellbelove: That was mistakenly left in. I was trying to decide whether I should go with 'Disabled & not-found' or 'Explicit static_assert' error messages.

jwellbelove: One shows up at compile time and the other can highlighted by the IDE in the editor.

rolandreichweinbmw (approved):

PR #1469, opened 2026-06-23
Documentation updates

jwellbelove:
Fixed some formatting and renamed files to match the ETL convention.


docs/containers/binary/bitset.md
```diff
@@ -86,11 +86,11 @@ etl::bitset<>::npos
 template <size_t Size, typename TElement>
 etl::bitset<Size, TElement>::npos
 ```
-`Number_Of_Elements`
-`Bits_Per_Element`
-`Allocated_Bits`
-`All_Set_Element`
-`All_Clear_Element`
+`Number_Of_Elements`
```
rolandreichweinbmw, line 89: Are you sure you only want to add trailing space to those lines?

A cleanup would rather remove them?

docs/containers/binary/bitset.md
```diff
@@ -224,7 +224,7 @@ template <typename T>
 T value() const
 ```
 **Description**  
-Returns the value corresponding to the bitset.
+Returns the value corresponding to the bitset.
```
rolandreichweinbmw, line 227: dito.

docs/containers/binary/bitset.md
```diff
@@ -234,8 +234,8 @@ unsigned long to_ulong() const
 unsigned long long to_ullong() const
 ```
 **Description**  
-Functions for compatibility with the STL.
-Calls value<unsigned long>() or value<unsigned long long>().
+Functions for compatibility with the STL.
```
rolandreichweinbmw, line 237: dito.

docs/containers/binary/bitset.md
```diff
@@ -246,7 +246,7 @@ template <typename TString>
 TString to_string(typename TString::value_type zero = typename TString::value_type('0'),
                   typename TString::value_type one  = typename TString::value_type('1')) const
 ```
-Returns the value as a string of `0` and `1` characters.
+Returns the value as a string of `0` and `1` characters.
```
rolandreichweinbmw, line 249: dito.

docs/containers/binary/bitset.md
```diff
@@ -301,7 +301,7 @@ For C++98/03
 bool operator[](size_t position) const
 ```
 **Description**  
-Returns the boolean state of the indexed bit.
+Returns the boolean state of the indexed bit.
```
rolandreichweinbmw, line 304: ... even more.

Can you please check the whole diff?

docs/containers/binary/bitset.md
jwellbelove: A double space at the end of a line forces a 'newline'.
Hugo's Markdown ignores simple returns in the text if it is not a code block.

Without trailing double spaces

`Allocated_Bits`
`All_Set_Element`
`All_Clear_Element`

is rendered as 
`Allocated_Bits` `All_Set_Element` `All_Clear_Element`

rolandreichweinbmw: Interesting! Was not aware of that. Until now, I was handling trailing whitespace as redundant.

Thanks for the note!

drewr95 (approved):

docs/containers/binary/bitset.md
rolandreichweinbmw: I had a look at Hugo + Goldmark (the Markdown formatter in use):

The double space at the end of a line is somehow counterintuitive, and editors and git commits would misread it as wrong (i.e. to be corrected/removed).

As I understand, we can even without trailing spaces:
* Use double newline for new paragraph
* Use trailing backslash for hard newline. Basically, the same result as with double spaces, but more obvious.

I would suggest to do without the trailing spaces approach. I volunteer to convert current documentation.

What do you think?

jwellbelove: > The double space at the end of a line is somehow counterintuitive,

And also invisible when editing.

> Use double newline for new paragraph

I already use that.

> Use trailing backslash for hard newline

That sounds like a better option

rolandreichweinbmw: Great! Sounds like we can do without trailing spaces for now.

PR #1477, opened 2026-06-24
Fix for 1405: etl::exchange doesn't work with unique_ptr

vcoselev:
- etl::exchange was using the copy constructor. Changed for etl::move.
- The assignment to the new object is now done with forward.
- Added a test with unique_ptr.


include/etl/utility.h
```diff
@@ -493,19 +493,11 @@ namespace etl
...
-  T exchange(T& object, const T& new_value)
-  {
-    T old_value = object;
-    object      = new_value;
-    return old_value;
-  }
-
-  template <typename T, typename U>
-  T exchange(T& object, const U& new_value)
+  template <typename T, typename U = T>
```
jwellbelove, line 515: You will need to conditionally compile the old and new versions of `exchange` with `#if ETL_USING_CPP11`

vcoselev: I just realised that the library has syntax checking for C++03. Done!

test/test_utility.cpp
```diff
@@ -315,6 +316,16 @@ namespace
...
     }
 
+    //*************************************************************************
+    TEST(test_exchange_unique_ptr)
+    {
+      etl::unique_ptr<int> p1(new int(1));
+      etl::unique_ptr<int> p2 = etl::exchange(p1, nullptr);
+
+      CHECK(p1.get() == nullptr);
+      CHECK_EQUAL(*p2.get(), 1);
```
serges147:
```suggestion
      CHECK_FALSE(p1);
      CHECK_TRUE(p2);
      CHECK_EQUAL(*p2, 1);
```
See `operator bool()` and `operator*`.
Nitpicking, but IMO with these operators it's more C++ ideomatic (rather than dealing with raw pointers).

include/etl/utility.h
```diff
@@ -508,6 +509,18 @@ namespace etl
     object      = new_value;
     return old_value;
   }
+  #else
+  //***************************************************************************
+  /// exchange
+  //***************************************************************************
+  template <typename T, typename U = T>
+  T exchange(T& object, U&& new_value)
```
serges147: I would add also:
1. `ETL_CONSTEXPR20_STL`
2. `ETL_NOEXCEPT_IF(etl::is_nothrow_move_constructible<T>::value &&
        etl::is_nothrow_assignable<T&, U>::value)`

See [possible implementation](https://cppreference.com/cpp/utility/exchange) at cppreference.
You can find similar 1. & 2. at `etl::optional`.

vcoselev: 1. `etl::is_nothrow_assignable` doesn't exist in the source code.
2. `etl::optional` doesn't use ETL_NOEXCEPT_IF or noexcept.

vcoselev: @jwellbelove Maybe a future enhancement should be implement is_nothrow_assignable. IMO the PR is ready to go.

test/test_utility.cpp
vcoselev: Makes sense. Done!

include/etl/utility.h
```diff
@@ -508,6 +509,18 @@ namespace etl
...
+  {
+    T old_value = etl::move(object);
+    object      = etl::forward<U>(new_value);
+    return old_value;
+  }
+  #endif // ETL_CPP11_NOT_SUPPORTED
 #else
   //***************************************************************************
   /// exchange (const)
```
rolandreichweinbmw: The STL branch should be fixed also.

test/test_utility.cpp
```diff
@@ -315,6 +316,17 @@ namespace
       CHECK_EQUAL(1, c);
     }
 
+    //*************************************************************************
+    TEST(test_exchange_unique_ptr)
+    {
+      etl::unique_ptr<int> p1(new int(1));
+      etl::unique_ptr<int> p2 = etl::exchange(p1, nullptr);
```
rolandreichweinbmw, line 323: Assigning nullptr selects unique_ptr::operator=(std::nullptr_t) (a reset). The deleted operator=(const unique_ptr&) — the one the const U& wrapper would drag in — is never instantiated.

I.e. the issue is silently ignored, therefore the STL case does not fail.

rolandreichweinbmw (changes requested):

rolandreichweinbmw:
I would propose sth. like:

```
diff --git a/include/etl/utility.h b/include/etl/utility.h
index 93157d14..0863e958 100644
--- a/include/etl/utility.h
+++ b/include/etl/utility.h
@@ -523,12 +523,12 @@ namespace etl
   #endif // ETL_CPP11_NOT_SUPPORTED
 #else
   //***************************************************************************
-  /// exchange (const)
+  /// exchange
   //***************************************************************************
   template <typename T, typename U = T>
-  T exchange(T& object, const U& new_value)
+  T exchange(T& object, U&& new_value)
   {
-    return std::exchange(object, new_value);
+    return std::exchange(object, etl::forward<U>(new_value));
   }
 #endif // ETL_NOT_USING_STL || ETL_CPP14_NOT_SUPPORTED

diff --git a/test/test_utility.cpp b/test/test_utility.cpp
index 40c07812..5b824aa3 100644
--- a/test/test_utility.cpp
+++ b/test/test_utility.cpp
@@ -327,6 +327,21 @@ namespace
       CHECK_EQUAL(*p2, 1);
     }

+    TEST(test_exchange_unique_ptr_move_in)
+    {
+      etl::unique_ptr<int> a(new int(10));
+      etl::unique_ptr<int> b(new int(20));
+
+      etl::unique_ptr<int> old = etl::exchange(a, etl::move(b));
+
+      CHECK_TRUE(old);
+      CHECK_EQUAL(*old, 10);
+      CHECK_TRUE(a);
+      CHECK_EQUAL(*a, 20);
+      CHECK_FALSE(b);
+    }
+
     TEST(test_as_const)
     {
```

(On top of your changes.)

I.e. covering the STL case also and fix the test to cover the case when a move-only value is assigned into an object.

I have a similar solution here: https://github.com/rolandreichweinbmw/etl/tree/fix-exchange (see last commit)

include/etl/utility.h
serges147: > 2\. `etl::optional` doesn't use ETL_NOEXCEPT_IF or noexcept.

Sorry, but there are 30 instances of ETL_NOEXCEPT_IF at `etl::optional`, and 42 more of `ETL_NOEXCEPT`.

vcoselev: @serges147 Is it better to do it with ETL_CONSTEXPR, ETL_CONSTEXPR20 or some other one? I've seen that STD implemented the constexpr in C++20.

test/test_utility.cpp
vcoselev: Makes sense. I'll change it.

include/etl/utility.h
serges147: I believe that `ETL_CONSTEXPR20_STL` should be used. At least cppreference [mentions](https://cppreference.com/cpp/utility/exchange) it since c++20.

vcoselev: `ETL_CONSTEXPR20_STL` doesn't make sense because that part of the code isn't using STL right?. I suppose you meant `ETL_CONSTEXPR20` which will not do anything for the non-STL code and put a constexpr for the STL code.

Image

rolandreichweinbmw: Please note that `ETL_CONSTEXPR20_STL` defines `constexpr` only in case of STL build combined with C++20. Otherwise dropping `constexpr`. If defining it because we are in the STL compile branch, we don't really need the `_STL` suffix since it would be redundant.

Also note that we don't strictly need to stick to the C++ standard in the sense of defining `constexpr` only starting from a certain C++ standard since it is only defined there. In contrast, we enable certain feature intentionally even earlier "because we can". E.g. ranges available in ETL and C++17, even if C++ has it only from C++20 on.

I.e. maybe try to use `ETL_CONSTEXPR` and add special casing only if technical reasons like compile errors or missing compiler support speak against it.

rolandreichweinbmw (commented):
Regarding the commit message: "The assignment to the new object is now done without forward" : I guess you meant "with forward" instead of "without"?

Nitpick: etl:move -> etl::move

include/etl/utility.h
```diff
@@ -508,16 +509,35 @@ namespace etl
...
+  }
+  #endif // ETL_CPP11_NOT_SUPPORTED
 #else
   //***************************************************************************
-  /// exchange (const)
+  /// exchange
   //***************************************************************************
   template <typename T, typename U = T>
-  T exchange(T& object, const U& new_value)
+  T exchange(T& object, U&& new_value)
```
rolandreichweinbmw: The STL path can have constexpr/noexcept also?

vcoselev: I thought that std would have implemented that. I took that else branch as a wrapper. If necessary I can put it there.

include/etl/utility.h
serges147: Yes, Roland, I agree. I mentioned C++ standard 20 only b/c assumed that it wouldn't work on earlier standard (like e.g. constexpr destructors since C++20). But it definitely worth to try. I didn't noticed `_STL` at all.

include/etl/utility.h
rolandreichweinbmw: The internal call to std::exchange does not propagate to the etl function. You need to specify it explicitly.

vcoselev: Makes sense!

include/etl/utility.h
```diff
@@ -508,16 +509,35 @@ namespace etl
     object      = new_value;
     return old_value;
   }
+  #else
+  //***************************************************************************
+  /// exchange
+  //***************************************************************************
+  template <typename T, typename U = T>
+  #if ETL_CPP14_SUPPORTED
```
jwellbelove: You can use ETL_CONSTEXPR14

The ETL defines 
ETL_CONSTEXPR  / ETL_CONSTEXPR11
ETL_CONSTEXPR14
ETL_CONSTEXPR17
ETL_CONSTEXPR20

jwellbelove:
I'd like to push a new release soon, so if you can fix the last few issues I can include this in the update.

include/etl/utility.h
vcoselev: I've used ETL_CONSTEXPR14 instead of ETL_CONSTEXPR because C++11 constexpr functions cannot define variables (they must consist of a single return statement). ETL_CONSTEXPR14 conditionally expands to constexpr only in C++14 and later.

vcoselev:
@jwellbelove I think it's ready. Although this [action](https://github.com/ETLCPP/etl/actions/runs/29085477124) failed before.

include/etl/utility.h
```diff
@@ -500,24 +501,36 @@ namespace etl
...
+  }
+  #endif // ETL_CPP11_NOT_SUPPORTED
 #else
   //***************************************************************************
-  /// exchange (const)
+  /// exchange
   //***************************************************************************
   template <typename T, typename U = T>
-  T exchange(T& object, const U& new_value)
+  T exchange(T& object, U&& new_value)
```
rolandreichweinbmw: Still missing the ETL_CONSTEXPR14 and ETL_NOEXCEPT_IF in this STL branch as discussed. I guess this was just forgotten instead of being intentional?

Otherwise looks good!

rolandreichweinbmw (approved):

include/etl/utility.h
vcoselev: I went in to deep with the constexpr and completely forgot about it 😓. Just made the change!

Thank you!

jwellbelove (approved):

PR #1478, opened 2026-06-24
Drop problematic `T& emplace(const optional_impl& other)`

serges147:
- For both fundamental and non-fundamental specializations.
- This also fixes incorrect `return storage.u.value;` in the fundamental specializations.
- Simplify `emplace` overloads (drop SFINAE).
- Update test cases to:
  - verify returned reference
  - verify `nothrow`
- Add `emplace(std::initializer_list, ...)` overload for non-fundamental types.

### DIFF
--- a/include/etl/optional.h
+++ b/include/etl/optional.h
@@ -198,7 +198,7 @@ namespace etl
       //*******************************************
       template <typename U, typename... TArgs >
       ETL_CONSTEXPR20_STL optional_impl(etl::in_place_t, std::initializer_list<U> ilist, TArgs&&... args)
-        ETL_NOEXCEPT_IF((etl::is_nothrow_constructible<T, std::initializer_list<U>, TArgs...>::value))
+        ETL_NOEXCEPT_IF((etl::is_nothrow_constructible<T, std::initializer_list<U>&, TArgs...>::value))
       {
         storage.construct(ilist, etl::forward<TArgs>(args)...);
       }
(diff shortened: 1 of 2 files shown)


test/test_optional.cpp
```diff
@@ -1120,11 +1148,11 @@ namespace
       CHECK_EQUAL(1, (*opt2)[0]);
       CHECK_EQUAL(20, (*opt2)[1]);
 
-      etl::optional<const ItemType> opt3;
+      etl::optional<etl::optional<const ItemType>> opt3;
```
serges147, line 1151: This is b/c `emplace(const optional_impl& other)` api is not available anymore, so if you want emplace an optional into another optional then target type should be `optional<optional<T>>`.

Just in case, I kept `create_optional_issue_1171` test almost as is (except `opt3` type, and double dereferencing now needed).

include/etl/optional.h
```diff
@@ -198,7 +198,7 @@ namespace etl
       //*******************************************
       template <typename U, typename... TArgs >
       ETL_CONSTEXPR20_STL optional_impl(etl::in_place_t, std::initializer_list<U> ilist, TArgs&&... args)
-        ETL_NOEXCEPT_IF((etl::is_nothrow_constructible<T, std::initializer_list<U>, TArgs...>::value))
+        ETL_NOEXCEPT_IF((etl::is_nothrow_constructible<T, std::initializer_list<U>&, TArgs...>::value))
```
serges147, line 201: Adding missing `&` reflects the way how we construct, namely with lvalue `ilist`. See cppreference how they define overload resolution constraints. E.g. for [emplace(initializer_list)](https://cppreference.com/cpp/utility/optional/emplace):
>2) Initializes the contained value by calling its constructor with ilist, std::forward<Args>(args)... as parameters. This overload participates in overload resolution only if std::is_constructible<T, std::initializer_list<U>&, Args&&...>::value is true.

test/test_optional.cpp
```diff
@@ -94,7 +94,7 @@ namespace
     {
     }
 
-    ETL_CONSTEXPR20 TestIL(std::initializer_list<int> il, int a_, int b_, int c_)
+    ETL_CONSTEXPR20 TestIL(std::initializer_list<int>& il, int a_, int b_, int c_)
```
rolandreichweinbmw: How about keeping the by-value version?

More idiomatic C++, and more real world example.

test/test_optional.cpp
```diff
@@ -1365,7 +1393,7 @@ namespace
       NothrowAtAll() noexcept {}
       NothrowAtAll(const NothrowAtAll&) noexcept {}
       NothrowAtAll(NothrowAtAll&&) noexcept {}
-      NothrowAtAll(std::initializer_list<int>) noexcept {}
+      NothrowAtAll(std::initializer_list<int>&) noexcept {}
```
rolandreichweinbmw: Dito.

test/test_optional.cpp
```diff
@@ -1408,11 +1436,11 @@ namespace
...
     {
-      ThrowingAll() {}                           // may throw
-      ThrowingAll(const ThrowingAll&) {}         // may throw
-      ThrowingAll(ThrowingAll&&) {}              // may throw
-      ThrowingAll(std::initializer_list<int>) {} // may throw
-      ThrowingAll& operator=(const ThrowingAll&) // may throw
+      ThrowingAll() {}                            // may throw
+      ThrowingAll(const ThrowingAll&) {}          // may throw
+      ThrowingAll(ThrowingAll&&) {}               // may throw
+      ThrowingAll(std::initializer_list<int>&) {} // may throw
```
rolandreichweinbmw: Dito.

test/test_optional.cpp
serges147: Yes, normally it's by-value. I made it here (in test entities) as lvalue ref to specifically verify that various `is_[nothrow_]constructible` in `optional.h` are using `<T, std::initializer_list<U>&, ...` - see my above [comment for optional.h](https://github.com/ETLCPP/etl/pull/1478/files/48ed439b36200867410fa3ae54ba92c2a80d059d#diff-bda7929c3eb7b537b60c97adc1062f5c6e55624753d5be9f7509312e7fca8d54).

rolandreichweinbmw: You could test it additionally, to demonstrate?

serges147: What do you mean exactly, like what?

rolandreichweinbmw: Two versions:
1. `ETL_CONSTEXPR20 TestIL(std::initializer_list<int> il, int a_, int b_, int c_)`
2. `ETL_CONSTEXPR20 TestIL(std::initializer_list<int>& il, int a_, int b_, int c_)`

In case of doubt, keep 1. for the idiomatic C++ case.

serges147: Ok, I'll keep 1. only. Thanks!

BTW, now I noticed that in the same `is_constructable<T, ..., Args...>` expressions I'm missing also `&&` for the `Args...`. Should be `Args&&...` according to the cppreference. I'll try reproduce it (make it fail to be constructible just b/c of missing &&). I bet we are missing these `&&` in many similar places at ETL.

serges147: Ok, there is no difference between `Args...` and `Args&&...` when it comes to `is_constructable`-friends.
This is because `is_constructible` models arguments using `declval<Arg>()`, so
```
std::declval<Args>()      // gives Args&&
std::declval<Args&&>()    // also gives Args&& after reference collapsing
```

rolandreichweinbmw (approved):

rolandreichweinbmw:
When "squash and merge" (or in case another PR push), we should correct the typo in the commit message: "Ror" -> "For"

jwellbelove (approved):

PR #1492, opened 2026-07-03
Allow etl::vector to store pointers to functions and member functions

### LINKED ISSUE #1479: etl::vector is incompatible with pointers to functions
I expect this example to compile:

```cpp
#include <etl/vector.h>

using fn_ptr = int(*)();

int example_fn() { return 42; }

int main() {
    etl::vector<fn_ptr, 4> v;
    v.push_back(&example_fn);
    return 0;
}
```

but it doesn't, as of 5eed38e25a092f84cae262b0f2320091879f19a6:

```
In file included from test.cpp:1:
In file included from etl/include/etl/vector.h:1265:
etl/include/etl/private/ivectorpointer.h:356:25: error:
      cannot initialize a parameter of type 'value_type' (aka 'void *') with an
      lvalue of type 'parameter_t' (aka 'int (*)()')
  356 |       base_t::push_back(value);
      |                         ^~~~~
test.cpp:9:7: note: in instantiation of member function
      'etl::ivector<int (*)()>::push_back' requested here
    9 |     v.push_back(&example_fn);
      |       ^
etl/include/etl/private/pvoidvector.h:412:31: note:
      passing argument to parameter 'value' here
  412 |     void push_back(value_type value)
      |                               ^
```

That is, function pointers can't be converted to object pointers, and the implementation delegates to `pvoidvector` even for function pointers.

jwellbelove:
This PR fixes #1479

Previously, `etl::vector` could only store pointers to objects.
This change allows vectors of pointers to functions and member functions.

**Changes**

- Removed the `etl::ivector` specialisations for `T*` & `const T*`.
- Added enables to `etl::ivector` to select between two implementations; one for object pointers, the second for all other types.
- `etl::ivector` for object pointers uses the original `pvoidvector` implementation.
- Removed the intermediate class `etl::ivectorpointer`.
- Added `etl::is_object_pointer` and `etl::is_object_pointer_v` to `type_traits.h`.
- Updated `type_traits.h`.
- Added tests for function and member function pointers.


test/vs2022/etl.vcxproj
```diff
@@ -11462,6 +11462,10 @@
     <ClCompile Include="..\test_variant_pool.cpp" />
     <ClCompile Include="..\test_vector.cpp" />
     <ClCompile Include="..\test_vector_external_buffer.cpp" />
+    <ClCompile Include="..\test_vector_function_pointer.cpp" />
+    <ClCompile Include="..\test_vector_function_pointer_external_buffer.cpp" />
+    <ClCompile Include="..\test_vector_member_function_pointer.cpp" />
+    <ClCompile Include="..\test_vector_member_function_pointer_external_buffer.cpp" />
```
rolandreichweinbmw, line 11467: Those 4 should also be added to `test/meson.build`.

docs/types/type-traits.md
```diff
@@ -459,3 +459,20 @@ template <typename T, template <typename...> class Template>
...
+template <typename T>
+struct is_object_pointer
+```
+Checks if `T` is a pointer to an object.  
+Pointers to functions and member functions return `false`.  
+The result is found in the member `value`.  
+
+```cpp
+template <typename T>
+inline constexpr bool is_is_object_pointer_v = etl::is_object_pointer<T>::value;
```
rolandreichweinbmw: is_is_object_pointer_v -> is_object_pointer_v

drewr95 (approved):

include/etl/vector.h
```diff
@@ -1840,23 +2100,42 @@ namespace etl
...
+
+        typename etl::ivector<T>::iterator itr = rhs.begin();
+        while (itr != rhs.end())
+        {
+          this->push_back(etl::move(*itr));
+          ++itr;
+        }
+
+        rhs.initialise();
+      }
```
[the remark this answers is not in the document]
jwellbelove: Checks and asserts added

jwellbelove:
The GCC C++23 tests with optimisation are failing within the compiler's own STL implementation.

rolandreichweinbmw:
Two issues at once:

1. The warnings from
```
test/test_vector_function_pointer.cpp
test/test_vector_function_pointer_external_buffer.cpp
test/test_vector_member_function_pointer.cpp
test/test_vector_member_function_pointer_external_buffer.cpp
```
can be suppressed by surrounding those files with the common

```
#include "etl/private/diagnostic_uninitialized_push.h"
...
#include "etl/private/diagnostic_pop.h"
```

2. A compiler bug in GCC 12 and 13 (gone in GCC 14). Can be fixed like this in `test/CMakeLists.txt`:
```
# ---------------------------------------------------------------------------
# GCC 12 and 13 miscompile the inter-procedural mod/ref analysis
# (-fipa-modref) of std::vector<pointer-to-member-function> operations. The
# corruption makes the std::vector reference data used by these tests report a
# bogus size, throwing spurious std::length_error / std::bad_alloc (or crashing
# with SIGSEGV) at -O2 and above. The defect is fixed in GCC 14; disabling the
# pass for just these two translation units restores correct code generation.
# The flag is applied only when the compiler accepts it, so Clang and GCC < 12
# (which do not know -fno-ipa-modref) are left untouched.
# ---------------------------------------------------------------------------
if (CMAKE_CXX_COMPILER_ID MATCHES "GNU")
	include(CheckCXXCompilerFlag)
	check_cxx_compiler_flag("-fno-ipa-modref" ETL_HAS_FNO_IPA_MODREF)
	if (ETL_HAS_FNO_IPA_MODREF)
		set_source_files_properties(
			test_vector_member_function_pointer.cpp
			test_vector_member_function_pointer_external_buffer.cpp
			PROPERTIES COMPILE_OPTIONS "-fno-ipa-modref")
	endif ()
endif ()
```

See also the last 2 commits in this branch: `https://github.com/rolandreichweinbmw/etl/tree/issue/etl-vector-is-incompatible-with-pointers-to-functions`

Alternatively, we could also switch to the newer GCC (via a newer distribution environment), as we are already doing with the C++26.

jwellbelove:
Thanks, I'll give that a try.
My research did seem to point to a compiler bug in GCC13

I've been attempting, on my local setup this afternoon, to get the compiler to ignore warnings in its own STL, but that seems to be either difficult, or impossible. The next step was to upgrade the compiler.

jwellbelove:
Moving to GCC14 fixed some things, but has a number of `-Werror=maybe-uninitialized` and `-Wnull-dereference` issues on files that didn't used to have them (`test_map.cpp` and more).

jwellbelove:
@rolandreichweinbmw 
I'm more than a bit confused.

`development` and `master` are are at the same commit.
I based my new code changes on `development` and compiling with GCC C++23 -O2 gives multiple warnings, even in files that I've not edited.
Pushing to the PR fails the checks for GCC C++23.

Your PRs are based on `master`, and the checks don't fail with any warnings.

I switched to `master` locally and ran GCC C++23 -O2 tests and I still get multiple warnings.
The original push to `master` passed all checks.

I'm at a loss to understand what's going on here.
Any ideas?

rolandreichweinbmw:
> I'm at a loss to understand what's going on here. Any ideas?

@jwellbelove it all makes sense when you realize that reproducing the issue is not deterministic. As long as it is present, it sometimes hits, sometimes not.

Then it makes sense under the following conditions:
* No matter if based on master or development (as you correctly pointed out)
* When my above described fixes are applied, it should not be reproducable anymore

rolandreichweinbmw:
As documented in https://developers.redhat.com/articles/2023/06/21/new-c-features-gcc-13, C++23 is considered experimental in GCC 13. Maybe it's not worth it to waste too much energy into this combination and update the compiler for our C++23 checks. (As mentioned above, we can do similar as for C++26.)

jwellbelove:
I'm running GCC14 locally and I've changed the GCC C++23 yaml file to use GCC14.
I am now getting many warnings on the 'ranges' tests in `test_algorithm.cpp`.

I've spent more time chasing GCC's false positives than writing the PR in the first place.

.github/workflows/gcc-c++23.yml
```diff
@@ -87,11 +87,11 @@ jobs:
...
-        export CXX=g++
+        export CC=gcc-14
+        export CXX=g++-14
         cmake -DBUILD_TESTS=ON -DNO_STL=ON -DETL_USE_TYPE_TRAITS_BUILTINS=OFF -DETL_USER_DEFINED_TYPE_TRAITS=OFF -DETL_FORCE_TEST_CPP03_IMPLEMENTATION=ON -DETL_CXX_STANDARD=23 -DETL_OPTIMISATION=-O3 ./
         gcc --version
         make -j "$(getconf _NPROCESSORS_ONLN)"
-    
+
     - name: Run tests
       run: ./test/etl_tests -v
```
drewr95: This file is missing a newline

drewr95 (approved):

drewr95 (approved):

.github/workflows/gcc-c++23.yml
```diff
@@ -21,12 +21,12 @@ jobs:
     - name: Build
       run: |
         export ASAN_OPTIONS=alloc_dealloc_mismatch=0,detect_leaks=0
-        export CC=gcc
-        export CXX=g++
+        export CC=gcc-14
+        export CXX=g++-14
         cmake -DBUILD_TESTS=ON -DNO_STL=OFF -DETL_USE_TYPE_TRAITS_BUILTINS=OFF -DETL_USER_DEFINED_TYPE_TRAITS=OFF -DETL_FORCE_TEST_CPP03_IMPLEMENTATION=OFF -DETL_CXX_STANDARD=23 -DETL_OPTIMISATION=-O3 ./
         gcc --version
```
rolandreichweinbmw: Maybe also change to `gcc-14 --version` here and below?

jwellbelove: Do you mean change all .yml files for GCC to use 14?

rolandreichweinbmw: Only this single file (where gcc-14 and g++-14 is used explicitly), at several occurrences of "gcc --version".

jwellbelove: All occurrences in this file have already been changed to gcc-14 and g++-14.

rolandreichweinbmw: As far as I can see, it still reads `gcc --version` in 4 places.

Might be confusing if `gcc` in the search path is different from `gcc-14`?

jwellbelove: Oh, I see you mean the `gcc --version`.
I thought you were talking about the `export` lines.

rolandreichweinbmw (approved):

PR #1494, opened 2026-07-05
Refactor variant::emplace to perfect forward args instead of copy/move construct

drewr95:
Fixes #1494

This pull request refactors the construction logic in the `etl::variant` implementation to improve efficiency and support move-hostile types. The main change is the introduction of a new `do_emplace` method, which directly constructs the variant alternative in place, avoiding unnecessary temporaries and moves. The existing code paths that previously used `do_construct` have been updated to use `do_emplace` instead.

**Variant construction improvements:**

* Added the `do_emplace` method, which directly constructs the variant alternative in place from forwarded arguments, supporting both trivially and non-trivially destructible types and avoiding unnecessary temporaries.
* Updated all variant emplacement code paths to use `do_emplace` instead of `do_construct`, ensuring more efficient and safer construction for all types. [[1]](diffhunk://#diff-55c450f94b98abc7922e61cd75fafa56d69999aae96bed375a9da754f9e80244L718-R718) [[2]](diffhunk://#diff-55c450f94b98abc7922e61cd75fafa56d69999aae96bed375a9da754f9e80244L738-R738) [[3]](diffhunk://#diff-55c450f94b98abc7922e61cd75fafa56d69999aae96bed375a9da754f9e80244L758-R758) [[4]](diffhunk://#diff-55c450f94b98abc7922e61cd75fafa56d69999aae96bed375a9da754f9e80244L778-R778)

### DIFF
--- a/include/etl/private/variant_variadic.h
+++ b/include/etl/private/variant_variadic.h
@@ -715,7 +715,7 @@ namespace etl
       using type = etl::remove_cvref_t<T>;
 
       do_destroy();
-      do_construct<type>(type(etl::forward<TArgs>(args)...));
+      do_emplace<type>(etl::forward<TArgs>(args)...);
 
       type_id = index_of_type<T>::value;
 
@@ -735,7 +735,7 @@ namespace etl
       using type = etl::remove_cvref_t<T>;
 
       do_destroy();
-      do_construct<type>(type(il, etl::forward<TArgs>(args)...));
+      do_emplace<type>(il, etl::forward<TArgs>(args)...);
 
       type_id = index_of_type<T>::value;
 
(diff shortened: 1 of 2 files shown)


drewr95:
@jwellbelove I tried bumping to the latest version of etl (20.48.0) and was met with some issues when I have underlying types that are non-movable / non-copyable because emplace was creating a temporary object and using move/copy construction.

It would be great if this PR could make it into a 20.48.1 release.

drewr95:
I'm not capable of assigning reviewers @rolandreichweinbmw but I know you handled implementing the constexpr + ROM variadic union implementation, so I'd like you to see this.

rolandreichweinbmw (commented):
Looks good in general!

However, tests are missing. I.e. reverting this change would get unnoticed.

rolandreichweinbmw:
E.g. the last commit I added to https://github.com/rolandreichweinbmw/etl/tree/1493-emplace-args-construct

drewr95:
@rolandreichweinbmw Thank you for adding the tests. I went ahead and cherry-picked it in from your implementation into this.

rolandreichweinbmw (approved):

PR #1495, opened 2026-07-06
Separate bit order and endianness in bit_stream.h

rolandreichweinbmw:
See also discussion in the Slack channel with Ivan Zrno

### DIFF
--- a/docs/utilities/bit-stream-reader.md
+++ b/docs/utilities/bit-stream-reader.md
@@ -3,11 +3,11 @@ title: "bit_stream_reader"
 ---
 
 {{< callout type="info">}}
-  Header: `bit_stream_reader.h`  
+  Header: `bit_stream.h`  
   Since: `TBC`  
 {{< /callout >}}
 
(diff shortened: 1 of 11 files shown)


rolandreichweinbmw:
In both `read_value()` and `write_data()`, we have `nbits = (nbits > (CHAR_BIT * sizeof(T))) ? (CHAR_BIT * sizeof(T)) : nbits;`

Supporting `nbits > CHAR_BIT * sizeof(T)` at all silently ignores the invalid values and clamps them to this maximum. Since technically, it's already a comparison and branch, how about ETL_ASSERT() instead? So it's not even a performance penalty. (Even faster in case of `ETL_NO_CHECKS`.)

jwellbelove:
> Since technically, it's already a comparison and branch, how about ETL_ASSERT() instead?

That would be a reasonable change, rather than failing silently.

rolandreichweinbmw:
Ok, adjusted accordingly.

drewr95 (approved):

jwellbelove (approved):

PR #1496, opened 2026-07-08
Fix non-assignable alternatives in variant

rolandreichweinbmw:
When using non-assignable alternatives in variant, the new variant implementation led to compile errors because an assignment was used instead of a construction.

This is a regression from the previous ETL version.

Note: This PR is based on the development branch since a rebase from master to development already conflicted. Therefore, starting at development already.


drewr95 (approved):
Looks good!  Thank you

jwellbelove (approved):

PR #1507, opened 2026-07-14
Various fixes and cleanup

rolandreichweinbmw:
Typo fixes

Fix bit_stream: Guard nbits > 0 to prevent division by zero

Fix test_vector to check for the correct iterator position after erase

This was done before, but a recent change dropped it.

Fix several test_vector_* for valid data

Instead of comparing the moved-from data which is invalid, compare with
the valid initial_data.

Fix histogram test

Use different Start index so skipped start_index copy would not get
unnoticed in the test.


drewr95 (approved):

include/etl/type_traits.h
```diff
@@ -3034,7 +3034,7 @@ namespace etl
 
   //*********************************************
   // Force the user to provide specialisations for
-  // anything other than arithmetics and pointers.
+  // anything other than arithmetic and pointers.
```
jwellbelove: That doesn't read right to me.

test/test_vector.cpp
```diff
@@ -1143,9 +1143,12 @@ namespace
       Compare_Data compare_data(initial_data.begin(), initial_data.end());
       Data         data(initial_data.begin(), initial_data.end());
 
-      compare_data.erase(compare_data.begin() + 2, compare_data.begin() + 4);
+      Compare_Data::iterator cdi = compare_data.erase(compare_data.begin() + 2, compare_data.begin() + 4);
 
-      data.erase(data.begin() + 2, data.begin() + 4);
+      Data::iterator di = data.erase(data.begin() + 2, data.begin() + 4);
+
+      CHECK(cdi == compare_data.begin() + 2);
```
jwellbelove, line 1150: You're checking that `std::vector` works?

include/etl/type_traits.h
jwellbelove: "other than arithmetics and pointers."

or

"other than arithmetic and pointer types."

rolandreichweinbmw: Agree. Adjusting.

test/test_vector.cpp
rolandreichweinbmw: The test is (also) supposed to check the correct iterator position after etl::vector erase() (Data data).

This was dropped in a recent change in test_vector.cpp.

This change re-introduces the check.

jwellbelove (approved):

PR #1511, opened 2026-07-14
Fix deep const-cast in ivector pointer repair for const pointee types

rolandreichweinbmw:
to_void_pptr() previously stripped const from the pointed-to type via element_type, breaking pointer-to-const element types such as etl::vector<const int*>. Convert through void* instead, removing only the buffer's top-level const and avoiding reinterpret_cast.


include/etl/vector.h
```diff
@@ -1728,7 +1726,10 @@ namespace etl
     // Convert from pointer to void**
     static ETL_CONSTEXPR14 void** to_void_pptr(pointer p_buffer) ETL_NOEXCEPT
     {
-      return reinterpret_cast<void**>(const_cast<element_type*>(p_buffer));
+      ETL_STATIC_ASSERT(!etl::is_const<typename etl::remove_pointer<pointer>::type>::value,
```
jwellbelove, line 1729: In what circumstance would this static assert be triggered?
Is this redundant?

drewr95 (approved):

include/etl/vector.h
rolandreichweinbmw: It actually filters out the case when T is a top-level const-qualified pointer.

Admittedly, an unusual case in practical projects, but it makes sure to give you even clearer diagnostics in case the user got it wrong.

jwellbelove (approved):

PR #1513, opened 2026-07-17
string_view: clamp n in remove_prefix and remove_suffix

mike919192:
I was comparing ETL to the STL hardening [proposal](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2025/p3471r4.html).  One check that I found was missing from ETL was string_view remove_prefix and remove_suffix.

[array_view](https://github.com/ETLCPP/etl/blob/7f290365a9f1b72bf89b4b77ec7dbe39b228bd8e/include/etl/array_view.h#L577) already had a check to effectively clamp n to the size of the view.  I think that is a good approach so I copied it to string_view.  Another option could be to assert if n > size().


rolandreichweinbmw:
Regarding

https://en.cppreference.com/cpp/string/basic_string_view/remove_prefix
https://en.cppreference.com/cpp/string/basic_string_view/remove_suffix

an `ETL_ASSERT()` would be more standard conforming, as defined in the "hardened" implementation, preventing UB.

Admittedly, deviates from `array_view`, but the latter is non-standard anway. Also, `array_view` is mostly redundant to `etl::span`, being the newer, standard-conforming interface.

So with reference to `etl::span` which also follows the `ETL_ASSERT` path, the `ETL_ASSERT` path would be consistent.

include/etl/string_view.h
```diff
@@ -425,18 +425,26 @@ namespace etl
 
     //*************************************************************************
     /// Shrinks the view by moving its start forward.
+    /// When n >= size(), the view is set to empty.
     //*************************************************************************
     ETL_CONSTEXPR14 void remove_prefix(size_type n) ETL_NOEXCEPT
     {
-      mbegin += n;
+      if (n < size())
```
rolandreichweinbmw, line 433: Formally, `n <= size()` would be the correct assert statement.

mike919192:
> Regarding
> 
> https://en.cppreference.com/cpp/string/basic_string_view/remove_prefix https://en.cppreference.com/cpp/string/basic_string_view/remove_suffix
> 
> an `ETL_ASSERT()` would be more standard conforming, as defined in the "hardened" implementation, preventing UB.
> 
> Admittedly, deviates from `array_view`, but the latter is non-standard anway. Also, `array_view` is mostly redundant to `etl::span`, being the newer, standard-conforming interface.
> 
> So with reference to `etl::span` which also follows the `ETL_ASSERT` path, the `ETL_ASSERT` path would be consistent.

A better comparison that I should have mentioned, would be to [etl::string_view.substr()](https://github.com/ETLCPP/etl/blob/7f290365a9f1b72bf89b4b77ec7dbe39b228bd8e/include/etl/string_view.h#L412).  That function also clamps the parameters instead of asserting.

rolandreichweinbmw:
Let @jwellbelove decide :-) Either way would be an improvement.

jwellbelove:
A while ago I started to use `ETL_ASSERT_OR_RETURN_VALUE` so that if the assert does not trigger an exception or 'halt' (such as by using `ETL_NO_CHECKS`) the result will never be undefined.

```
ETL_ASSERT_OR_RETURN_VALUE(n <= size(), ETL_ERROR(string_view_bounds), view);
```

mike919192:
> A while ago I started to use `ETL_ASSERT_OR_RETURN_VALUE` so that if the assert does not trigger an exception or 'halt' (such as by using `ETL_NO_CHECKS`) the result will never be undefined.
> 
> ```
> ETL_ASSERT_OR_RETURN_VALUE(n <= size(), ETL_ERROR(string_view_bounds), view);
> ```

remove_prefix and remove_suffix are void return though.  So I'm thinking that you want me to insert `ETL_ASSERT(n <= size(), ETL_ERROR(string_view_bounds);` at the beginning of the functions, and then the current clamping logic after that in case error checking is disabled.

Also updating NOEXCEPT for the possible exception throw.

include/etl/string_view.h
```diff
@@ -425,18 +425,28 @@ namespace etl
 
     //*************************************************************************
     /// Shrinks the view by moving its start forward.
+    /// When n >= size(), the view is set to empty.
     //*************************************************************************
-    ETL_CONSTEXPR14 void remove_prefix(size_type n) ETL_NOEXCEPT
+    ETL_CONSTEXPR14 void remove_prefix(size_type n) ETL_NOEXCEPT_IF(ETL_NOT_USING_EXCEPTIONS)
```
jwellbelove: The definition of `ETL_NOEXCEPT` already depends on `ETL_USING_EXCEPTIONS`.

include/etl/string_view.h
mike919192: The assert is added to both functions

include/etl/string_view.h
mike919192: I see this now, but I'm a bit confused about this.  There's a bunch of other examples of this being done for example [array](https://github.com/ETLCPP/etl/blob/7f290365a9f1b72bf89b4b77ec7dbe39b228bd8e/include/etl/array.h#L114) and [span](https://github.com/ETLCPP/etl/blob/7f290365a9f1b72bf89b4b77ec7dbe39b228bd8e/include/etl/span.h#L251).  Before your comment, I thought the intention was that these examples should be marked noexcept if ETL_NOT_USING_EXCEPTIONS is true.  But if noexcept is always disabled, should it just be removed?

jwellbelove:
The use of `ETL_NOEXCEPT_EXPR(ETL_NOT_USING_EXCEPTIONS)` is redundant and a review oversight in `array` and `span`. They should be simplified to `ETL_NOEXCEPT`.

If `ETL_NOT_USING_EXCEPTIONS == 1` then `ETL_NOEXCEPT_EXPR` is defined as 'nothing'.

include/etl/string_view.h
mike919192: I removed the `ETL_NOEXCEPT_IF`.  I did a search for files that have `ETL_NOEXCEPT_IF(ETL_NOT_USING_EXCEPTIONS)` or `ETL_NOEXCEPT_EXPR(ETL_NOT_USING_EXCEPTIONS)` and there were 5 hits.

`include/etl/not_null.h` `include/etl/alignment.h` `include/etl/signal.h` `include/etl/array.h` and `include/etl/span.h`

After that there's also other uses that could be simplified.

jwellbelove (approved):

PR #1514, opened 2026-07-17
Fix variant copy operations enabled with non-copyable alternatives

### LINKED ISSUE #1512: etl::variant copy constructor remains enabled with non-copyable alternatives
`etl::variant` can be copied even when one of its alternatives is not copy-constructible.

When the non-copyable alternative is active, the copy compiles, but the destination variant appears to copy the active index without constructing the active value. This can leave the copied variant in an invalid state.

## Reproducer

```cpp
#include <etl/variant.h>

struct MoveOnly
{
  MoveOnly() = default;
  MoveOnly(const MoveOnly&) = delete;
  MoveOnly& operator=(const MoveOnly&) = delete;
  MoveOnly(MoveOnly&&) = default;
  MoveOnly& operator=(MoveOnly&&) = default;

  ~MoveOnly() {}
};

using Variant = etl::variant<int, MoveOnly>;

int main()
{
  Variant source(etl::in_place_type_t<MoveOnly>{});

  Variant copy = source; // Compiles, but copy is invalid.
}
```

The equivalent `std::variant` code is rejected at compile time because its copy constructor is deleted.

## Expected behavior

`etl::variant<T...>` should not be copy-constructible unless every alternative is copy-constructible.

For example:

```cpp
static_assert(!etl::is_copy_constructible<Variant>::value);
```

Copy assignment may have the same issue and should be checked as well.

rolandreichweinbmw:
etl::variant was copy-constructible and copy-assignable even when an alternative was not copyable, and copying a non-copyable active alternative silently copied only the index, leaving the variant in an invalid state.

Move the copy/move construct and assign logic into variant_base and add a variant_copy_move_control mixin that conditionally deletes the copy and move special members based on whether every alternative is copy/move constructible. The variant's special members are now defaulted, so they are deleted unless all alternatives support the operation.

Fixes: #1512


rolandreichweinbmw:
@coderabbit full review

drewr95 (approved):
Wow, thank you for doing this! Looks great :)

jwellbelove (approved):

PR #1501, opened 2026-07-12
Topic/optional monadic operations

davmf:
* topic/optional-monadic-operations:
	- added and_then, or_else, and transform to etl::optional with unit tests
	- added is_optional to optional.h, used in and_then and or_else to ensure that the returned type is an optional
	- implemented as ref-qualified overloads with C++11 compatible trailing return types, following the etl::expected implementation

See #1500

PR #1515, opened 2026-07-18
Clean up the use of type traits by using C++11 forms wherever possible.

jwellbelove:
Replaced lines such as

`typename U = typename etl::remove_cvref< typename etl::invoke_result<F, void, TValue&&>::type>::type`

with

`typename U = etl::remove_cvref_t< etl::invoke_result_t<F, TValue&&>>`

This change would also be applicable to #1501


rolandreichweinbmw (approved):

drewr95 (approved):
Nice cleanup :)

PR #1519, opened 2026-07-21
fix(devcontainer): resolve missing packages in bullseye image

mkilivan:
This PR fixes devcontainer provisioning on Debian Bullseye by removing unavailable apt packages and installing `cogapp` in a distro-independent way.

**What changed**
* Updated Dockerfile to:
  * Remove `python3-cogapp` from `apt-get install` (not available on Bullseye)
  * Remove `clang-format-18` from `apt-get install` (not available on Bullseye)
  * Keep `clang-format` installation via apt
  * Add `python3 -m pip install --no-cache-dir cogapp`

**Why**
The previous package list referenced Bullseye-unavailable packages, causing `clang:18` to `clang:21` devcontainer setup failures. Installing cogapp via pip avoids distro package availability issues.

**Impact**
* Devcontainer builds succeed on Bullseye.
* Existing tool/version verification behavior is preserved.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -36,13 +36,12 @@ RUN set -eux \
     && apt-get -y install --no-install-recommends \
         python3-full \
         python3-pip \
-        python3-cogapp \
         git \
         wget \
         cmake \
         clang-format \
-        clang-format-18 \
         lcov \
+    && python3 -m pip install --no-cache-dir cogapp \
     && rm -rf /var/lib/apt/lists/*
 
 RUN set -eux; \


jwellbelove:
What version of `clang-format` does bullseye pull in?

drewr95 (approved):

.devcontainer/Dockerfile
```diff
@@ -36,13 +36,12 @@ RUN set -eux \
...
         python3-full \
         python3-pip \
-        python3-cogapp \
         git \
         wget \
         cmake \
         clang-format \
-        clang-format-18 \
         lcov \
+    && python3 -m pip install --no-cache-dir cogapp \
```
jwellbelove, line 44: My research suggests that, to keep this compatible with Debian 12 and above, this line should be 
`&& python3 -m pip install --no-cache-dir --break-system-packages cogapp \`

mkilivan:
> What version of `clang-format` does bullseye pull in?

Thanks for the question. I checked Debian package metadata, and Bullseye pulls in clang-format 11 (`1:11.0-51+nmu5`), which is older than the clang-format 18 this repository expects.

Only the devcontainer configs for clang 18-21 pin Bullseye; the others use the generic `clang` image tags. Is there a reason to pin Bullseye specifically? If not, I’d suggest replacing `silkeh/clang:18-bullseye` with `silkeh/clang:18` so no Dockerfile change would be needed. The same applies to `silkeh/clang:19`, `silkeh/clang:20`, and `silkeh/clang:21`.

rolandreichweinbmw:
I also vote for dropping bullseye support. It is outdated by latest Debian be several years and versions.

rolandreichweinbmw:
Closing this PR in favour of the other PR, removing bullseye support. See discussion.

Issue #1029, opened 2025-02-11
Add etl::counting_semaphore and etl::binary_semaphore for GCC/Clang


jwellbelove:
Add `etl::counting_semaphore` and `etl::binary_semaphore` using GCC/Clang atomic builtins.

PR #1520, opened 2026-07-22
Add etl::counting_semaphore and etl::binary_semaphore

julien-hecht-netatmo:
As part of my work, I had create a FreeRTOS semaphore in ETL, so I thought I might as well try to upstream it. I wrote a bit more code to solve #1029 at the same time.

I took inspiration from `etl::mutex` which doesn't have any unit test as far as I saw so I didn't include any either, though if you'd rather I write some I can do that.


include/etl/semaphore/semaphore_freertos.h
```diff
@@ -0,0 +1,119 @@
...
+    // On FreeRTOS, LeastMaxValue = 0 causes an assert. This deviates slightly from the C++ spec, but it should be fine.
+    ETL_STATIC_ASSERT(LeastMaxValue > 0, "LeastMaxValue should be 1 or more");
+    ETL_STATIC_ASSERT(LeastMaxValue <= numeric_limits<BaseType_t>::max(), "LeastMaxValue is too high");
+
+  public:
+
+    counting_semaphore(ptrdiff_t desired)
+    {
+      sem = xSemaphoreCreateCountingStatic(max(), desired, &semaphore_allocation);
+    }
```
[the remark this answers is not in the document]
julien-hecht-netatmo: In general I'd agree with this, but there are a few issues:

- FreeRTOS is mostly used in contexts where we don't have or want exceptions, so throwing is impractical
- This can't be made constexpr (deviating from std) because of `xSemaphoreCreateCountingStatic`
- I can't make a constructor "fail", So checking whether `desired` is negative wouldn't help much

I'm unsure how to solve that cleanly.

include/etl/semaphore/semaphore_freertos.h
```diff
@@ -0,0 +1,119 @@
...
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
+******************************************************************************/
+
+#ifndef ETL_SEMAPHORE_FREERTOS_INCLUDED
+#define ETL_SEMAPHORE_FREERTOS_INCLUDED
+
+#include "../platform.h"
```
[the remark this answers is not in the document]
julien-hecht-netatmo: I don't think this is applicable to headers in subfolders of `include/etl`.

include/etl/semaphore/semaphore_freertos.h
```diff
@@ -0,0 +1,119 @@
...
+#include "../chrono.h"
+#include "../limits.h"
+
+#include "FreeRTOS.h"
+#include <semphr.h>
```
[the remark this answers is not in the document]
julien-hecht-netatmo: I'm not entirely sure how to do that cleanly. I can't put it in `semaphore.h` obviously, and it is impractical to put in in `semaphore_freertos.h`. `mutex.h` also does not have this gate.

include/etl/semaphore/semaphore_std.h
```diff
@@ -0,0 +1,47 @@
...
+#include <semaphore>
+
+namespace etl
+{
+  //***************************************************************************
+  ///\ingroup semaphore
+  ///\brief These semaphore classes are implemented using std::counting_semaphore and std::binary_semaphore.
+  //***************************************************************************
+  template <ptrdiff_t LeastMaxValue>
+  using counting_semaphore = std::counting_semaphore<LeastMaxValue>;
```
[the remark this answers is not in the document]
julien-hecht-netatmo: I'm not entirely sure how to solve this problem: on GCC for example this could be `INT_MAX`, `SEM_VALUE_MAX` or `POSIX_SEM_VALUE_MAX`, all of which have potentially different values.
`POSIX_SEM_VALUE_MAX` is `INT16_MAX` which is (typically) the lower of the 3, but this feels like a band-aid rather than a proper fix.

Maybe this could be put in `platform.h` or a similar "configuration" header with a default value to let the library end-user change it?

include/etl/semaphore/semaphore_gcc_sync.h
```diff
@@ -0,0 +1,116 @@
...
+  {
+  public:
+
+    ETL_STATIC_ASSERT(LeastMaxValue >= 0, "LeastMaxValue shall be non-negative");
+    ETL_STATIC_ASSERT(LeastMaxValue <= numeric_limits<ptrdiff_t>::max(), "LeastMaxValue is too high");
+
+    ETL_CONSTEXPR14 explicit counting_semaphore(ptrdiff_t desired)
+      : counter(desired)
+    {
+      ETL_ASSERT(desired >= 0 && desired <= max(), "Preconditions: desired >= 0 is true and desired <= max() is true");
```
jwellbelove: An `ETL_ASSERT` should use an object derived from `etl::exception`, not a C string.

i.e.

```
namespace etl
{
  struct semaphore_error : etl::exception
  {
  ...
  };
}

ETL_ASSERT(desired >= 0 && desired <= max(), ETL_ERROR(etl::semaphore_error));
```

jwellbelove:
The semaphore headers appear in the wrong filter in VS2022. I've created a patch to fix this.

[etl-#1520-vs2022-filter.patch](https://github.com/user-attachments/files/30385108/etl-.1520-vs2022-filter.patch)

julien-hecht-netatmo:
> The semaphore headers appear in the wrong filter in VS2022. I've created a patch to fix this.
> 
> [etl-#1520-vs2022-filter.patch](https://github.com/user-attachments/files/30385108/etl-.1520-vs2022-filter.patch)

Thank you for your patch. As I don't work on Windows, I couldn't get that to work. I just added it to the PR.

include/etl/semaphore/semaphore_gcc_sync.h
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    void release(ptrdiff_t update = 1)
+    {
+      ptrdiff_t old = __atomic_fetch_add(&counter, update, __ATOMIC_RELEASE);
+
+      // Preconditions are tested afterwards because otherwise it makes the atomic code needlessly complex.
+      ETL_ASSERT(update >= 0, ETL_ERROR(semaphore_out_of_bounds));
+      ETL_ASSERT(update <= max() - old, ETL_ERROR(semaphore_out_of_bounds));
+    }
```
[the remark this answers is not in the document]
jwellbelove: The `ETL_ASSERT` should also be changed to `ETL_ASSERT_OR_RETURN` so that the `__atomic_fetch_add` is not executed if either assert fails.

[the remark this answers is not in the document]
julien-hecht-netatmo: Moving the asserts to before the atomic operation would mean:

- Doing a first atomic operation to fetch the value,
- Running the asserts,
- Doing a CAS operation to check whether it's actually possible to decrement the counter,
- And all of that in a loop until it succeeds (atomic number didn't change between the read/fetch/CAS and the next CAS) or fails (counter can't be decremented).

This is a lot of complexity for a release operation. As far as I can read, `libstdc++` chose to keep the operation simple and do the assert after the atomic too: [waiting uses a loop](https://github.com/gnutools/gcc/blob/7c38c56214bf2809399a2198441bb48ee1b00512/libstdc%2B%2B-v3/include/bits/atomic_wait.h#L355), but not [releasing the semaphore](https://github.com/gnutools/gcc/blob/7c38c56214bf2809399a2198441bb48ee1b00512/libstdc%2B%2B-v3/include/std/semaphore#L71) (see also [the underlying implementation](https://github.com/gnutools/gcc/blob/7c38c56214bf2809399a2198441bb48ee1b00512/libstdc%2B%2B-v3/include/bits/semaphore_base.h#L297).

In general, C++ implementations seem to prioritize speed and efficiency over correctness, so I thought implementing it this way would be fine. In addition, the way the standard is written makes me think `release` is not supposed to be written like this as it only mentions doing the fetch+add, not any kind of "retry" or loop.
I would write the precondition checks another way, but I'm not sure how. I imagine it's possible with templates? In general I agree with this not being perfectly safe, but I'm also really not sure how to make it safer in a way that doesn't make it needlessly complex as an operation.

Edit: found [the source of `__glibcxx_assert`](https://github.com/gnutools/gcc/blob/7c38c56214bf2809399a2198441bb48ee1b00512/libstdc%2B%2B-v3/include/bits/c%2B%2Bconfig#L661). It's just a regular assert (`if (!cond) abort()`).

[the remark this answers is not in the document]
jwellbelove: > Moving the asserts to before the atomic operation would mean:
> 
> * Doing a first atomic operation to fetch the value,
> * Running the asserts,
> * Doing a CAS operation to check whether it's actually possible to decrement the counter,
> * And all of that in a loop until it succeeds (atomic number didn't change between the read/fetch/CAS and the next CAS) or fails (counter can't be decremented).

Ah yes, I see. The post check is probably the better solution.

julien-hecht-netatmo:
With further testing, I have noticed an issue where `semaphore_std.h` is basically unusable in its current form because the compiler cannot convert between `etl::ratio`, `etl::duration` etc and their `std` counterparts. I've pushed the new version.

include/etl/semaphore/semaphore_std.h
```diff
@@ -0,0 +1,120 @@
...
+
+      std::chrono::time_point<private_semaphore_std::CompatClock<Clock>, CompatDuration> time_point{
+        CompatDuration(abs_time.time_since_epoch().count())};
+
+      return sem.try_acquire_until(time_point);
+    }
+
+    void release(ptrdiff_t update = 1)
+    {
+      return sem.release(update);
```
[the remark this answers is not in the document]
julien-hecht-netatmo: There is no need to verify because it is verified by the std implementation. This assumes the std implementation is correct which I sure hope it is lol

include/etl/semaphore.h
```diff
@@ -0,0 +1,60 @@
...
+#define ETL_SEMAPHORE_INCLUDED
+
+#include "platform.h"
+
+#include "semaphore/semaphore_exceptions.h"
+
+#if defined(ETL_TARGET_OS_FREERTOS)
+  #include "semaphore/semaphore_freertos.h"
+  #define ETL_HAS_SEMAPHORE 1
+#elif ETL_USING_STL && ETL_USING_CPP20 && ETL_USING_STD_SEMAPHORE
```
rolandreichweinbmw, line 39: ETL_USING_STD_SEMAPHORE is never defined. Convention in ETL is to have a macro like ETL_NO_STD_SEMAPHORE which causes `platform.h` to define ETL_USING_STD_SEMAPHORE. See mutex.h for a similar pattern.

This issue causes `semaphore_std.h` not being used at all.

include/etl/semaphore/semaphore_std.h
```diff
@@ -0,0 +1,120 @@
...
+    {
+      std::chrono::duration<Rep, std::ratio<Period::num, Period::den>> duration(rel_time.count());
+
+      return sem.try_acquire_for(duration);
+    }
+
+    template <class Clock, class Duration>
+    bool try_acquire_until(const etl::chrono::time_point<Clock, Duration>& abs_time)
+    {
+      using CompatDuration = std::chrono::duration<typename Duration::rep, std::ratio<Duration::period::num, Duration::period::den>>;
```
rolandreichweinbmw, line 97: The CompatClock shim declares duration from Clock::period, so CompatClock::now() returns a different duration type than the time_point passed in. libstdc++'s timed-wait requires both to share _Duration1. acquire/release/try_acquire/try_acquire_for compile and run fine; only try_acquire_until is broken. Clang with libc++ accepts it, so this is a libstdc++-specific break — i.e. it would fail on GCC.

include/etl/semaphore/semaphore_freertos.h
```diff
@@ -0,0 +1,126 @@
...
+#include "../limits.h"
+#include "semaphore_exceptions.h"
+
+namespace etl
+{
+  //***************************************************************************
+  ///\ingroup semaphore
+  ///\brief This counting_semaphore class is implemented using FreeRTOS's semaphores
+  //***************************************************************************
+  template <ptrdiff_t LeastMaxValue = numeric_limits<BaseType_t>::max()>
```
rolandreichweinbmw, line 49: Some FreeRTOS types like BaseType_t, TickType_t, StaticSemaphore_t, SemaphoreHandle_t, xSemaphoreCreateCountingStatic, xSemaphoreTake/Give, portMAX_DELAY, portTICK_PERIOD_MS are use here but not defined via header include.

muex_freertos.h already does include FreeRTOS.h and semphr.h. Maybe do in this header also?

include/etl/semaphore/semaphore_freertos.h
```diff
@@ -0,0 +1,126 @@
...
+    template <class Rep, class Period>
+    bool try_acquire_for(const chrono::duration<Rep, Period>& rel_time)
+    {
+      using URep_t = make_unsigned_t<Rep>;
+
+      TickType_t           ticks_to_wait = 0;
+      chrono::milliseconds ms_to_wait{rel_time};
+      if (ms_to_wait > chrono::duration<Rep, Period>::zero())
+      {
+        ticks_to_wait = divide_round_to_ceiling(static_cast<URep_t>(ms_to_wait.count()), portTICK_PERIOD_MS);
```
rolandreichweinbmw, line 85: divide_round_to_ceiling() needs ../rounded_integral_division.h

include/etl/semaphore/semaphore_freertos.h
```diff
@@ -0,0 +1,126 @@
...
+
+    bool try_acquire()
+    {
+      return try_acquire_for(0_s);
+    }
+
+    template <class Rep, class Period>
+    bool try_acquire_for(const chrono::duration<Rep, Period>& rel_time)
+    {
+      using URep_t = make_unsigned_t<Rep>;
```
rolandreichweinbmw, line 79: make_unsigned_t needs ../type_traits.h

include/etl/semaphore/semaphore_freertos.h
```diff
@@ -0,0 +1,126 @@
...
+    void acquire()
+    {
+      // portMAX_DELAY=block forever
+      // Looping guarantees waiting forever even if the configuration does not have `INCLUDE_vTaskSuspend == 1`
+      while (xSemaphoreTake(sem, portMAX_DELAY) != pdTRUE);
+    }
+
+    bool try_acquire()
+    {
+      return try_acquire_for(0_s);
```
rolandreichweinbmw: Suffix _s only works when ETL_USE_VERBOSE_CHRONO_LITERALS are off. It maybe works here in tests because our test profile defines it, but not in general.

chrono::seconds(0) would be the portable solution.

test/test_etl_traits.cpp
```diff
@@ -69,6 +70,7 @@ namespace
       // Has...
       CHECK_EQUAL((ETL_HAS_ATOMIC == 1), etl::traits::has_atomic);
       CHECK_EQUAL((ETL_HAS_MUTEX == 1), etl::traits::has_mutex);
+      CHECK_EQUAL((ETL_HAS_SEMAPHORE == 1), etl::traits::has_semaphore);
```
rolandreichweinbmw, line 73: This test line is the only test added.

I strongly recommend adding a separate test_semaphore.cpp which includes:
* Instantiation of counting_semaphore, exercising acquire/release counting, try_acquire on empty/non-empty, both timed forms hitting timeout and success
* Instantiation of binary_semaphore, doing likewise

include/etl/semaphore/semaphore_freertos.h
```diff
@@ -0,0 +1,126 @@
...
+    static constexpr ptrdiff_t max() noexcept
+    {
+      return LeastMaxValue;
+    }
+
+  private:
+
+    // Non-copyable
+    counting_semaphore(const counting_semaphore&) ETL_DELETE;
+    counting_semaphore& operator=(const counting_semaphore&) ETL_DELETE;
```
rolandreichweinbmw, line 119: In semaphore_std.h, those lines are missing.

include/etl/semaphore/semaphore_freertos.h
```diff
@@ -0,0 +1,126 @@
...
+  //***************************************************************************
+  ///\ingroup semaphore
+  ///\brief This counting_semaphore class is implemented using FreeRTOS's semaphores
+  //***************************************************************************
+  template <ptrdiff_t LeastMaxValue = numeric_limits<BaseType_t>::max()>
+  class counting_semaphore
+  {
+    // On FreeRTOS, LeastMaxValue = 0 causes an assert. This deviates slightly from the C++ spec, but it should be fine.
+    ETL_STATIC_ASSERT(LeastMaxValue > 0, "LeastMaxValue should be 1 or more");
+    ETL_STATIC_ASSERT(LeastMaxValue <= numeric_limits<BaseType_t>::max(), "LeastMaxValue is too high");
```
rolandreichweinbmw, line 54: Those two lines are missing in semaphore_std.h

include/etl/semaphore/semaphore_gcc_sync.h
```diff
@@ -0,0 +1,117 @@
...
+
+    static constexpr ptrdiff_t max() noexcept
```
rolandreichweinbmw: ETL_CONSTEXPR and ETL_NOEXCEPT?

include/etl/semaphore/semaphore_std.h
```diff
@@ -0,0 +1,120 @@
...
+{
+  //***************************************************************************
+  ///\ingroup semaphore
+  ///\brief These semaphore classes are implemented using std::counting_semaphore and std::binary_semaphore.
+  //***************************************************************************
+
+  namespace private_semaphore_std
+  {
+    template <class Clock>
+    struct CompatClock
```
rolandreichweinbmw: compat_clock ? (snake_case convention for struct names)

test/test_etl_traits.cpp
julien-hecht-netatmo: Will do. This requires a bit more dev on my side. The only reason I did not write tests was because mutexes don't have tests either.

include/etl/semaphore/semaphore_std.h
julien-hecht-netatmo: When I wrote this, I was thinking of "translating" everything from ETL to STD namespace.
So, creating the STD clock from the ETL clock can only rely on the provided `Clock`, and creating the STD duration from the ETL duration can only rely on the provided `Duration`.

My thought was, adding an additional check here would be redundant: everything is forwarded to STD anyway, and STD should be responsible for checking consistency and validity of parameters. If it somehow doesn't, it's fair to add a check here.

I think I understand the issue, but I'm really not sure how to fix it, I don't have much intuition when it comes to templates.
To be honest, reading the libstdc++ implementation details is difficult for me atm, I'm really not used to the writing style and the heavy use of templates.

If you would be so kind to provide a patch, I would be grateful.

julien-hecht-netatmo:
I have pushed my new changes. Like the last time, please review the Visual Studio project files as I cannot test them on my devices.

jwellbelove:
It looks like you need to conditionally compile parts of your code to deal with when `ETL_HAS_ATOMIC` is `0`.

PR #1522, opened 2026-07-24
Fix ODR violation in optional

HazardyKnusperkeks:
Detected by GCC 15, when using in a module context.

### DIFF
--- a/include/etl/optional.h
+++ b/include/etl/optional.h
@@ -59,7 +59,7 @@ namespace etl
 
     // Convertible to any type of null non-member pointer.
     template <class T>
-    operator T*() const
+    ETL_CONSTEXPR operator T*() const ETL_NOEXCEPT
     {
       return 0;
     }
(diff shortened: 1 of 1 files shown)


include/etl/optional.h
```diff
@@ -74,7 +74,7 @@ namespace etl
   /// A null option.
   ///\ingroup utilities
   //*****************************************************************************
-  const nullopt_t nullopt = {};
+  ETL_INLINE_VAR const nullopt_t nullopt = {};
```
jwellbelove: The `const` could also be changed to `ETL_CONSTANT`.

This would require changes to `nullopt_t`.

```
class nullopt_t
{
public:

  // Convertible to any type of null non-member pointer.
  template <class T>
  ETL_CONSTEXPR operator T*() const ETL_NOEXCEPT
  {
    return 0;
  }

private:

  // Can't take address of nullopt.
  void operator&() const ETL_DELETE;
};
```

HazardyKnusperkeks: Done

drewr95 (approved):

include/etl/optional.h
```diff
@@ -74,7 +74,7 @@ namespace etl
   /// A null option.
   ///\ingroup utilities
   //*****************************************************************************
-  const nullopt_t nullopt = {};
+  ETL_INLINE_VAR ETL_CONSTANT nullopt_t nullopt = {};
```
drewr95, line 77: [nitpick] This could be simplified to brace initialization instead of assignment.

```suggestion
  ETL_INLINE_VAR ETL_CONSTANT nullopt_t nullopt{};
```

PR #1524, opened 2026-07-25
Fixed signed/unsigned mismatch warnings in VS2022

jwellbelove:
Fixed VS2022 signed/unsigned mismatch warnings with specialisations of `cmp_less_impl` that are called from `cmp_less`.

### DIFF
--- a/include/etl/utility.h
+++ b/include/etl/utility.h
@@ -231,6 +231,35 @@ namespace etl
     struct enable_if_comparable : etl::enable_if<is_standard_integer<T>::value && is_standard_integer<U>::value, TReturn>
     {
     };
+
+    //***********************************
+    /// cmp_less implementation helpers, selected by SFINAE to avoid
+    /// signed/unsigned comparison warnings in dead code branches.
+    //***********************************
+
+    // Both same signedness.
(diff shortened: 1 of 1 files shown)


include/etl/utility.h
```diff
@@ -231,6 +231,35 @@ namespace etl
...
     };
+
+    //***********************************
+    /// cmp_less implementation helpers, selected by SFINAE to avoid
+    /// signed/unsigned comparison warnings in dead code branches.
+    //***********************************
+
+    // Both same signedness.
+    template <typename T, typename U>
+    ETL_NODISCARD ETL_CONSTEXPR typename etl::enable_if<etl::is_signed<T>::value == etl::is_signed<U>::value, bool>::type cmp_less_impl(T t, U u)
```
serges147, line 242: John, could we use `etl::enable_if_t`? Thnx.

drewr95: And for the rest :)

jwellbelove: I would not work for C++03.
Until the day that no one requires the ETL to support C++03, if it can be complied with C++03, then it should work.

drewr95 (approved):

PR #1526, opened 2026-07-31
Fix make_array and make_deque rejecting const-qualified lvalues of the element type

rleigh-electradx:
`make_array` and `make_deque` forwarded each argument as `T` (the element type) rather than as its own deduced type. `etl::forward<T>` accepts only a non-const `T&` or a `T&&`, and a `T&&` cannot bind to a reference-related const lvalue, so
any call passing a const-qualified lvalue of type `T` failed to compile:

```cpp
static const char c = 42;
etl::make_array<char>(c, c);   // error: would lose const qualifier
```

Arguments of *other* types slipped through via temporary materialisation, which is why the existing conversion and move tests never caught it.

Forward each argument as its own deduced type and convert explicitly with `static_cast<T>`. The explicit conversion is required rather than plain forwarding: function parameters are not constant expressions, so narrowing initialisers such as `make_array<char>(0, 1)` - which the old code accepted and the tests cover - would otherwise be rejected by the braced initialiser list.

Adds regression tests for const/non-const lvalues of the element type and for constexpr evaluation. Verified with clang 22 (C++14/17/20) and GCC 14 (C++11/20); existing conversion and move-only patterns still compile.

Note: This fix and the unit test additions were prepared using Claude Fable.  If this isn't suitable for direct inclusion, I hope that the fix and the tests provide a suitable example for what to correct.  Other than `make_array` and `make_deque` I didn't find any other instances of this defect; a separate defect recorded in https://github.com/ETLCPP/etl/issues/1527 for `etl::list`.

Kind regards,
Roger Leigh


jwellbelove:
Does this issue also affect `make_list`, `make_forward_list`, `make_map` etc?

rleigh-electradx:
> Does this issue also affect `make_list`, `make_forward_list`, `make_map` etc?

Not that I can see. I checked and this only affects `make_array` and `make_deque` as far as I could determine, the others are all OK.

rleigh-electradx:
By the way, if you would like const lvalue testcases for all of the other make_ functions to properly verify they are compliant, I'm happy to add them either in this PR, or a separate one, whichever you prefer.

jwellbelove:
I tried it for `etl::list` and this fails.

```
    TEST(test_make_list_from_const_lvalues_of_element_type)
    {
      static const char static_const_lvalue = 42;
      const char        local_const_lvalue  = 43;
      char              mutable_lvalue      = 44;

      auto data = etl::make_list<char>(static_const_lvalue, local_const_lvalue, mutable_lvalue);

      auto itr = data.begin();

      using Type = etl::remove_reference_t<decltype(*itr)>;
      CHECK((std::is_same<char, Type>::value));

      CHECK_EQUAL(42, *itr++);
      CHECK_EQUAL(43, *itr++);
      CHECK_EQUAL(44, *itr++);
    }
```

rleigh-electradx:
I got Claude Fable to do a complete review of all the `make_xxx` functions for all other containers, in commit 7812a85, and they all pass.  `make_list` tests were not added here, because they were already added in #1528, which is the source of the behaviour change.

The difference between the testcases and the example above is the template specialisation.  Claude's take on this:

> `make_list<char>(...)` was never a supported calling form. I verified the reviewer's exact test empirically on both trees:
> 
> On make-array-fix: it compiles and works, because `make_list` there still takes its pack by value, which copies const lvalues. So the defect doesn't actually manifest on this branch.
> On `etl/development` (with merged [#1528](https://github.com/rleigh-electradx/etl/issues/1528)): it fails to compile — but for a different reason than the `make_array` bug. `make_list` deduces its element type as the common type of the arguments; writing `make_list<char>` doesn't select the element type, it pins the first pack parameter to `char`, making it `char&&`, which can't bind to any lvalue. The sibling factories `make_vector`, `make_forward_list`, and `make_indirect_vector` have always rejected this form (verified on the pre-change tree), and the existing `test_make_vector` even documents the pinning behaviour: it calls `make_vector<char>(0, 1, ...)` and asserts the element type is `int`. The docs were corrected in [#1530](https://github.com/rleigh-electradx/etl/issues/1530) to show the supported form, `make_list(0, 1, ...)`.
> The natural calling style with `const` lvalues — `make_list(static_const_lvalue, ...)` — works on development and is already covered by `test_make_list_from_const_lvalues_of_element_type`, which we merged as part of [#1528](https://github.com/rleigh-electradx/etl/issues/1528). So in summary, the `const`-lvalue defect doesn't exist for `make_list`; the test's explicit `<char>` is the unsupported first-argument-pinning form (and worked on the old code only as an accident of the by-value signature).

What could be done is to support an equivalent of the Library Fundamentals TS which solved exactly this problem for `std::experimental::make_array` with a defaulted leading template parameter plus a lazily-evaluated element-type trait:

```cpp
// Element type is D when supplied; otherwise the common type of the arguments.
// The specialisation keeps common_type uninstantiated when D is supplied.
template <typename D, typename... TValues>
struct make_list_element { using type = D; };

template <typename... TValues>
struct make_list_element<void, TValues...> : etl::common_type<typename etl::decay<TValues>::type...> { };

template <typename D = void, typename... TValues>
constexpr auto make_list(TValues&&... values)
  -> etl::list<typename make_list_element<D, TValues...>::type, sizeof...(TValues)>
{
  using TElement = typename make_list_element<D, TValues...>::type;
  return {static_cast<TElement>(etl::forward<TValues>(values))...};
}
```

Because the element type comes from a defaulted parameter rather than the pack, `make_list<char>(...)` no longer pins the first argument's deduced type — the pack is always fully deduced, so lvalues (`const` or not) bind fine, and `static_cast` handles conversions exactly like the `make_array`/`make_deque` fix.

However,

* this is safe for `make_list` as an additive change
* applying the same pattern to `make_vector`/`make_forward_list`/`make_indirect_vector` would change existing behaviour. Today `make_vector<char>(0, 1, ...)` pins the first argument and deduces the element type as `int` — and the existing `test_make_vector` explicitly asserts that. Under this pattern it would produce `vector<char>`.  But it's arguably the correct and least surprising thing to do.

If that's something that would be useful to implement as a separate pull request, I'll be happy to do so.

Kind regards,
Roger

jwellbelove:
I feel that there is an inconsistency in the various `make_xxx` container implementations that should be looked at and resolved, if necessary.

_Today make_vector<char>(0, 1, ...) pins the first argument and deduces the element type as int — and the existing test_make_vector explicitly asserts that. Under this pattern it would produce vector<char>. But it's arguably the correct and least surprising thing to do._

I agree.

BTW, your fix to `etl::list` should definitely apply to `etl::forward_list`.

I only have limited time to look at this, as I will be away in two days time, and I will have minimal access to the internet for 12 days. I shall be back on the 20th August.
I think this needs a bit of time to think about solutions anyway.

rleigh-electradx:
> BTW, your fix to `etl::list` should definitely apply to `etl::forward_list`.

I have added a commit which adds the tests for `etl::forward_list`, but no code changes were required (it was already compliant).

rleigh-electradx:
> I feel that there is an inconsistency in the various `make_xxx` container implementations that should be looked at and resolved, if necessary.
> 
> _Today make_vector(0, 1, ...) pins the first argument and deduces the element type as int — and the existing test_make_vector explicitly asserts that. Under this pattern it would produce vector. But it's arguably the correct and least surprising thing to do._
> 
> I agree.

I have implemented this on the branch https://github.com/rleigh-electradx/etl/tree/container-libts-make-parameter in this commit: https://github.com/rleigh-electradx/etl/commit/212d31ef2e0448b84a4f8540660b433e541280fd.  Once the current PR is merged, I'll be able to rebase it and open a PR properly for it.  In the meantime, it's there to review and test.

This implements the Library Fundamentals TS for all sequence container types, including tests and documentation examples.  It's completely compatible with the current implementation, with the addition of explicit specialisation as in your test case, as well as the current automatic type deduction when the type is not specified.

Hope that's useful, and please let me know if you want any changes.  I'll be happy to open a new PR for it once this PR merged in.

Kind regards,
Roger

rolandreichweinbmw (approved):

drewr95 (approved):

jwellbelove (approved):

drewr95 (approved):

PR #1529, opened 2026-08-01
Fix etl::common_type decaying both arguments to T2

rleigh-electradx:
# Fix `etl::common_type` decaying both arguments to `T2`

## Problem

The two-type specialization of `etl::common_type` in `include/etl/type_traits.h` is based on the sample implementation on [cppreference](https://en.cppreference.com/w/cpp/types/common_type): if either argument is not already decayed, it should decay both arguments and recurse.

The recursion contained a typo — it passed `etl::decay<T2>::type` **twice**, discarding `T1` entirely:

```cpp
template <typename T1, typename T2>
struct common_type<T1, T2>
  : etl::conditional<etl::is_same<T1, typename etl::decay<T1>::type>::value && etl::is_same<T2, typename etl::decay<T2>::type>::value,
                     private_common_type::common_type_2_impl<T1, T2>,
                     common_type<typename etl::decay<T2>::type,   // <-- should be decay<T1>
                     typename etl::decay<T2>::type>>::type
{
};
```

As a result, whenever either argument is a reference (or otherwise non-decayed) type, the result is computed from `T2` alone:

```cpp
etl::common_type_t<long&, int&>  // yields int  — std::common_type_t gives long
```

The bug is easy to miss because many orderings are correct by accident: `common_type_t<int&, long&>` recurses as `common_type<long, long>` and happens to give the right answer; only cases where the wider (or differing) type is in the first position expose it.

### Impact

Any caller passing non-decayed types is affected. Notably `etl::make_vector`, `etl::make_forward_list`, and `etl::make_indirect_vector` take forwarding-reference packs and pass the deduced reference types straight to `etl::common_type_t` to compute the element type, so mixed-type *lvalue* arguments could deduce the wrong (narrower) element type:

```cpp
long l = 1; int i = 2;
auto v = etl::make_vector(l, i);  // was etl::vector<int, 2>, now etl::vector<long, 2>
```

## Solution

Recurse with **both** decayed types, matching the cppreference sample implementation the code cites:

```cpp
common_type<typename etl::decay<T1>::type,
            typename etl::decay<T2>::type>
```

## Tests

Added `TEST(test_common_type)` to `test/test_type_traits.cpp` (guarded by `ETL_USING_CPP11`), comparing `etl::common_type_t` against both the expected type and `std::common_type` for reference-type arguments:

- `common_type_t<long&, int&>` → `long` (the failing case)
- `common_type_t<int&, long&>` → `long`
- `common_type_t<const char&, char&>` → `char`
- `common_type_t<int, long, int>` → `long` (already-decayed multi-type sanity check)

Verified with clang 22: the `<long&, int&>` assertions fail before the fix and the full `test_type_traits` suite passes after it (also compile-checked at C++14 and C++17).

Note: As for #1526 this change was prepared with Claude Fable.


rleigh-electradx:
The single failing check is a docker timeout issue, not a problem with the code as far as I can tell.  It probably just needs restarting, but I don't have permission to do that.

drewr95 (approved):

jwellbelove (approved):

PR #1535, opened 2026-08-06
Make etl_profile Incompatible by Default

christophruethingbmw:
At the moment for Bazel when using the `etl` library it by default uses the `NO_PROFILE`. This can be dangerous since consumers might accidently fall back to this default. Therefore, this change makes the default incompatible, in case a consumer wants to use the `NO_PROFILE` that's possible by using the still existing `no_profile` library target.


rolandreichweinbmw (approved):

rolandreichweinbmw:
recheck

drewr95 (approved):

drewr95 (approved):

jwellbelove (approved):

Issue #1532, opened 2026-08-05
Const and temporary expected cannot be dereferenced


denravonska:
```Cpp
#include <etl/expected.h>

etl::expected<int, const char *> Make() {
    return 42;
}

int FromLvalue() {
    auto e = Make();
    return *e;        // operator*() &        -- compiles
}

int FromConstRef(const etl::expected<int, const char *> & e) {
    return *e;        // operator*() const &  -- does not compile
}

int FromTemporary() {
    return *Make();   // operator*() &&       -- does not compile
}
```

Tested on [Godbolt](https://godbolt.org/z/ef47x8KjY):
```
In file included from /opt/compiler-explorer/libs/etl/include/etl/expected.h:37,
                 from <source>:1:
/opt/compiler-explorer/libs/etl/include/etl/expected.h: In instantiation of 'const etl::expected<TValue, TError>::value_type& etl::expected<TValue, TError>::operator*() const & [with TValue = int; TError = const char*; value_type = int]':
<source>:13:13:   required from here
/opt/compiler-explorer/libs/etl/include/etl/error_handler.h:520:22: error: invalid initialization of reference of type 'const etl::expected<int, const char*>::value_type&' {aka 'const int&'} from expression of type 'std::nullptr_t'
  520 |             return (v);                           \
      |                      ^
/opt/compiler-explorer/libs/etl/include/etl/expected.h:749:7: note: in expansion of macro 'ETL_ASSERT_OR_RETURN_VALUE'
  749 |       ETL_ASSERT_OR_RETURN_VALUE(has_value(), ETL_ERROR(expected_invalid), ETL_NULLPTR);
      |       ^~~~~~~~~~~~~~~~~~~~~~~~~~
/opt/compiler-explorer/libs/etl/include/etl/expected.h: In instantiation of 'etl::expected<TValue, TError>::value_type&& etl::expected<TValue, TError>::operator*() && [with TValue = int; TError = const char*; value_type = int]':
<source>:17:18:   required from here
/opt/compiler-explorer/libs/etl/include/etl/error_handler.h:520:22: error: invalid initialization of reference of type 'etl::expected<int, const char*>::value_type&&' {aka 'int&&'} from expression of type 'std::nullptr_t'
  520 |             return (v);                           \
      |                      ^
/opt/compiler-explorer/libs/etl/include/etl/expected.h:760:7: note: in expansion of macro 'ETL_ASSERT_OR_RETURN_VALUE'
[...]

PR #1538, opened 2026-08-07
Fix expected::operator*() const& and && failing to compile

rolandreichweinbmw:
ETL_ASSERT_OR_RETURN_VALUE tried to return ETL_NULLPTR from functions returning references, which cannot be initialized from nullptr. Use plain ETL_ASSERT instead, matching the other operator*() overloads.

Add regression test for dereferencing const and temporary expected.


denravonska:
This closes #1532

drewr95 (approved):

jwellbelove (approved):

drewr95 (approved):

PR #1546, opened 2026-08-11
Fix delegate clear through assignment

HazardyKnusperkeks:
Because the ptr was not nulled, it did not compare equal to a default constructed delegate anymore.

### DIFF
--- a/test/test_delegate.cpp
+++ b/test/test_delegate.cpp
@@ -2167,6 +2167,20 @@ namespace
     #endif
     }
   #endif
+
+    TEST(test_delegate_clear_equals_default_constructed)
+    {
+      using delegate_type = etl::delegate<void(int, int)>;
+
+      delegate_type       delegate = {};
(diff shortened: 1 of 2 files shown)


test/test_delegate.cpp
```diff
@@ -2167,6 +2167,20 @@ namespace
     #endif
     }
   #endif
+
+    TEST(test_foo)
```
rolandreichweinbmw: test_delegate_clear_equals_default_constructed
?

rolandreichweinbmw (commented):
Generally looks good to me!

Before merging, the test name should get a reasonable name.

test/test_delegate.cpp
HazardyKnusperkeks: Of course, done.

rolandreichweinbmw (approved):

drewr95 (approved):

jwellbelove (approved):

PR #1548, opened 2026-08-11
delegate: restore argument deduction behaviour

### LINKED ISSUE #1547: `etl::delegate` fails to deduce arguments types
`etl::delegate` has been broken since 30ccdea62974fffa0e416c9926b295cf1b9ae829

> I'm not sure if this is the correct behaviour. For one, it breaks argument-type deduction for delegates. The following code compiles as expected with ETL 20.43.4 but not any version after:
> ```c++
> // delegate that takes a span
> etl::delegate<void(std::span<int>)> fn = <...>;
> 
> // invoke the delegate with an empty span
> fn({});
> 
> // or with a constructed span
> int *int_ptr = <...>;
> fn({int_ptr, 5});
> ```
> 
> Specifically, deducing constructors from braced initializer lists is what I've noticed is broken.
> 
> Additionally, the signature of `std::function`'s `operator()` _does not_ use perfect forwarding for the args:
> ```c++
> template <typename R ,typename Args...>
> class function<R(Args...)>
> {
>     <...>
>     R operator()(Args... args) const {...}
> };
> ```
> see https://en.cppreference.com/w/cpp/utility/functional/function/operator().html 

 _Originally posted by @edenfrosst in [#1203](https://github.com/ETLCPP/etl/issues/1203#issuecomment-4033790459)_

edenfrosst:
include new test that verifies arg deduction works

fixes #1547


rolandreichweinbmw (approved):
Well done, thanks!

drewr95 (approved):

jwellbelove (approved):

PR #1550, opened 2026-08-13
Add span deduction guide for types with data() and size().

denravonska:
This enables span construct with type deduction from containers which support .data() and .size(), primarily vector.


test/test_span_dynamic_extent.cpp
```diff
@@ -983,6 +983,53 @@ namespace
       CHECK((std::is_same_v<const int* const, std::remove_reference_t<decltype(pspan4_ext.front())>>));
     }
 
+  #if ETL_USING_STL
```
rolandreichweinbmw: This guard can be removed. It is rather intended for ETL production code.

In the tests, we can use std types etc. unconditionally (i.e. in both STL and non-STL builds) since we can assume the C++ environment supports the C++ standard std headers.

denravonska: Oki. It seems to be honored in other tests so it could be a good idea to go through them at some point. Will axe it here.

rolandreichweinbmw (approved):

jwellbelove (approved):

PR #1552, opened 2026-08-17
Add Macro for Inline Control

christophruethingbmw:
In specific situations there might be the need to enforce that a method is not inlined (when mapping a method to a specific section) or always inlined (when using e.g. raw assembly). Therefore, we provide two macros ETL_NO_INLINE and ETL_ALWAYS_INLINE.

Note, that we explicitly decided to throw an `#error` in case the compiler is not supported, since we otherwise might silently run into issues in case the programmer relies on the fact that a method is not inlined or is explicitly inlined.


include/etl/platform.h
```diff
@@ -524,6 +524,20 @@ SOFTWARE.
...
+// inline
+#if ETL_USING_GCC_COMPILER || ETL_USING_CLANG_COMPILER
+  #define ETL_NO_INLINE     __attribute__((noinline))
+  #define ETL_ALWAYS_INLINE __attribute__((always_inline))
+#elif ETL_USING_MICROSOFT_COMPILER
+  #define ETL_NO_INLINE     __declspec(noinline)
+  #define ETL_ALWAYS_INLINE __forceinline
+#else
+  // will trigger a compilation error if used
+  #define ETL_NO_INLINE     unsupported compiler
```
jwellbelove, line 537: #error?

rolandreichweinbmw:
The powerpc check will be fixed by https://github.com/ETLCPP/etl/pull/1554

drewr95 (approved):

include/etl/platform.h
rolandreichweinbmw: I think @christophruethingbmw decided against because `platform.h` would cause a compile error in this case just by inclusion, resulting in basically all ETL being unusable then.

jwellbelove (approved):

PR #1553, opened 2026-08-17
Adopt the Library Fundamentals TS `make_array` design for the sequence factories

rleigh-electradx:
Followup for #1526.  Based on development, rather than master, because it depends upon #1526 and related MRs.

`make_array`, `make_deque`, `make_vector`, `make_list`, `make_forward_list` and `make_indirect_vector` now follow the Library Fundamentals TS s`td::experimental::make_array` design: a defaulted leading template parameter selects the element type explicitly, falling back to the decayed common type of the arguments when it is not supplied. A shared trait (`etl::private_make::element_type`) performs the selection lazily, so `common_type` is not instantiated when the element type is explicit.

This fixes the explicit-element-type calling style for the four common-type factories, where `make_vector<char>(lvalue)` never compiled: the explicit argument pinned the first pack parameter to `char`, making it a `char&&` that no lvalue can bind to. It also gives `make_array` and `make_deque` the deduced form (e.g. `make_array(1, 2L)` is `array<long, 2>`) for free. All arguments are converted with `static_cast`, as in the `make_array` const-lvalue fix, so const-qualified lvalues and narrowing initialisers work in both forms.

Intentional behaviour change: an explicit element type now converts every argument instead of pinning only the first. `make_vector<char>(0, 1, ...)` previously produced `vector<int>` - the explicit char affected only the first argument and the element type was still the common type
- and now produces `vector<char>`. *test_make_vector* documented the old behaviour and has been updated.

The keyed `set` and `map` factories are unchanged: their key/mapped types are separate mandatory template parameters ahead of a fully-deduced pack, so the explicit-type style already works, and retrofitting deduction there would interact with the defaulted comparator/hash parameters.

Adds explicit-element-type tests for the four common-type factories (`const`/non-`const` lvalues plus narrowing initialisers) and deduced-element-type tests for `make_array` and `make_deque`, and updates the four documented signatures.

### DIFF
--- a/docs/containers/arrays/array.md
+++ b/docs/containers/arrays/array.md
@@ -43,7 +43,10 @@ constexpr auto make_array(TValues&&... values) -> etl::array</*element-type*/, s
 ```
 
 The element type is `T` when supplied, otherwise the decayed common type of
-the arguments (the Library Fundamentals TS `make_array` design).
+the arguments (the Library Fundamentals TS `make_array` design). An empty
+array requires the element type to be supplied explicitly, e.g.
+`etl::make_array<int>()`; a call with neither an element type nor arguments
+is ill-formed, as there is nothing to deduce the element type from.
 
 ### Example
 ```cpp
--- a/docs/containers/arrays/array.md
+++ b/docs/containers/arrays/array.md
@@ -38,13 +38,20 @@ Defines data as an array of 'int', of length '10', containing the supplied data.
 **C++11 and above**
 
 ```cpp
-template <typename T, typename... TValues>
-constexpr auto make_array(TValues&&... values) -> etl::array<T, sizeof...(TValues)>
+template <typename T = void, typename... TValues>
+constexpr auto make_array(TValues&&... values) -> etl::array<etl::private_make::element_type_t<T, TValues...>, sizeof...(TValues)>
 ```
+
+The element type is `T` when supplied, otherwise the decayed common type of
+the arguments (the Library Fundamentals TS `make_array` design).
 
 ### Example
 ```cpp
-auto data = etl::make_array<int>(0, 1, 2, 3, 4, 5, 6, 7, 8, 9);
+// Deduced element type: etl::array<int, 10>.
+auto data = etl::make_array(0, 1, 2, 3, 4, 5, 6, 7, 8, 9);
+
+// Explicit element type: etl::array<char, 10>. Each argument is converted.
+auto chars = etl::make_array<char>(0, 1, 2, 3, 4, 5, 6, 7, 8, 9);
 ```
 
 ## Member types
--- a/docs/containers/arrays/array.md
+++ b/docs/containers/arrays/array.md
@@ -39,7 +39,7 @@ Defines data as an array of 'int', of length '10', containing the supplied data.
 
 ```cpp
 template <typename T = void, typename... TValues>
-constexpr auto make_array(TValues&&... values) -> etl::array<etl::private_make::element_type_t<T, TValues...>, sizeof...(TValues)>
+constexpr auto make_array(TValues&&... values) -> etl::array</*element-type*/, sizeof...(TValues)>
 ```
 
 The element type is `T` when supplied, otherwise the decayed common type of
--- a/include/etl/array.h
+++ b/include/etl/array.h
@@ -1053,16 +1053,17 @@ namespace etl
   /// Make
   //*************************************************************************
 #if ETL_HAS_INITIALIZER_LIST
-  template <typename T, typename... TValues>
-  constexpr auto make_array(TValues&&... values) ETL_NOEXCEPT -> etl::array<T, sizeof...(TValues)>
+  template <typename T = void, typename... TValues>
+  constexpr auto make_array(TValues&&... values) ETL_NOEXCEPT -> etl::array<etl::private_make::element_type_t<T, TValues...>, sizeof...(TValues)>
   {
-    // Forward each argument as its own deduced type, then convert explicitly to T.
-    // Forwarding as T (the old behaviour) made the call ill-formed for any const-qualified
-    // lvalue of type T: forward<T> requires a non-const T&, and a T&& cannot bind to a
-    // reference-related const lvalue. The explicit conversion also keeps narrowing
-    // initialisers (e.g. make_array<char>(0, 1)) working, which plain forwarding into the
-    // braced initialiser list would reject.
-    return {static_cast<T>(etl::forward<TValues>(values))...};
+    // Library Fundamentals TS make_array design: the element type is T when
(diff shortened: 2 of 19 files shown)


jwellbelove (approved):

drewr95 (approved):

docs/containers/arrays/array.md
```diff
@@ -38,13 +38,20 @@ Defines data as an array of 'int', of length '10', containing the supplied data.
 **C++11 and above**
 
 ```cpp
-template <typename T, typename... TValues>
-constexpr auto make_array(TValues&&... values) -> etl::array<T, sizeof...(TValues)>
+template <typename T = void, typename... TValues>
+constexpr auto make_array(TValues&&... values) -> etl::array<etl::private_make::element_type_t<T, TValues...>, sizeof...(TValues)>
```
rolandreichweinbmw: The documentation is not supposed to leak the internal/implementation-specific namespace etl::private_make

include/etl/array.h
```diff
@@ -1053,16 +1053,17 @@ namespace etl
   /// Make
   //*************************************************************************
 #if ETL_HAS_INITIALIZER_LIST
-  template <typename T, typename... TValues>
-  constexpr auto make_array(TValues&&... values) ETL_NOEXCEPT -> etl::array<T, sizeof...(TValues)>
+  template <typename T = void, typename... TValues>
+  constexpr auto make_array(TValues&&... values) ETL_NOEXCEPT -> etl::array<etl::private_make::element_type_t<T, TValues...>, sizeof...(TValues)>
```
rolandreichweinbmw: When we have conversions now, better drop ETL_NOEXCEPT. The other make_*() also don't have it.

include/etl/type_traits.h
```diff
@@ -3751,6 +3751,29 @@ namespace etl
...
+  namespace private_make
+  {
+    template <typename TDesired, typename... TArgs>
+    struct element_type
+    {
+      using type = TDesired;
+    };
+
+    template <typename... TArgs>
+    struct element_type<void, TArgs...> : etl::common_type<typename etl::decay<TArgs>::type...>
```
rolandreichweinbmw: etl::common_type already decays, so this one is redundant. Harmless, but it adds an instantiation per argument

include/etl/array.h
```diff
@@ -1053,16 +1053,17 @@ namespace etl
...
-    // braced initialiser list would reject.
-    return {static_cast<T>(etl::forward<TValues>(values))...};
+    // Library Fundamentals TS make_array design: the element type is T when
+    // supplied explicitly, otherwise the decayed common type of the arguments.
+    // Each argument is forwarded as its own deduced type and converted
+    // explicitly, which accepts const-qualified lvalues (forwarding as the
+    // element type rejected them) and keeps narrowing initialisers such as
+    // make_array<char>(0, 1) working.
+    using TElement = etl::private_make::element_type_t<T, TValues...>;
+    return {static_cast<TElement>(etl::forward<TValues>(values))...};
```
rolandreichweinbmw: When TElement equals the argument type this creates a prvalue that, pre-C++17, requires an accessible move/copy constructor before the element is constructed. Types that are constructible-into-place but not movable would regress.

You could add a pass-through conversion helper like this:
https://github.com/rolandreichweinbmw/etl/commit/d9c83d82b6a2a2a19f04b60d9512730494e4da16

This just forwards in the non-conversion case, and does not limit the possible types involved.

docs/containers/arrays/array.md
rleigh-electradx: This has been corrected, thanks.

include/etl/array.h
rleigh-electradx: Done.

include/etl/type_traits.h
rleigh-electradx: This should now also be addressed.

include/etl/array.h
rleigh-electradx: This has now been implemented, but for `etl::array` only.

The non-movable fix is fully effective only for `etl::array` (aggregate init). For the node/buffer-based containers, the `initializer_list` constructors themselves still require movable or `is_convertible` types (e.g. `etl::list`'s init-list ctor does `push_back(etl::move(*itr))` pre-C++17, and `etl::vector`'s iterator assign is SFINAE-constrained on `etl::is_convertible`, which is false for move-deleted types). `convert` still strictly relaxes the factory layer for those containers, but it can't fix the existing behaviour of their internals.  This could potentially be done as a separate piece of work.

drewr95 (approved):

rolandreichweinbmw (approved):
Minor: etl::make_array<T>() is not producing an empty array as expected, but is invalid now.

Feel free to fix here or in a follow-up.

rleigh-electradx:
> Minor: etl::make_array() is not producing an empty array as expected, but is invalid now.

The `make_array()` invocation which was not explicitly-typed did not work before this change, and is ill-formed by TS design, because there are no parameters to deduce any type from.  The empty-array form that can exist (`make_array<T>()`) works and is now tested with an extra test-case added in the latest commit along with some additional documentation of the behaviour.

rolandreichweinbmw:
> The `make_array()` invocation which was not explicitly-typed did not work before this change, and is ill-formed by TS design, because there are no parameters to deduce any type from. The empty-array form that can exist (`make_array<T>()`) works and is now tested with an extra test-case added in the latest commit along with some additional documentation of the behaviour.

Great! Thanks for the clarification and new test.

include/etl/array.h
```diff
@@ -1053,16 +1053,17 @@ namespace etl
   /// Make
   //*************************************************************************
 #if ETL_HAS_INITIALIZER_LIST
```
rolandreichweinbmw, line 1055: Even if not strictly part of your change, the guard for ETL_HAS_INITIALIZER_LIST is wrong here, since for etl::array at least, we don't have initializer-list initialization but aggregate initialization which is available always in C++11.

When you adjust here to ETL_USING_CPP11, you can remove/change the guards for ETL_HAS_INITIALIZER_LIST  also in your new tests in test_array.cpp.

Like this: https://github.com/rolandreichweinbmw/etl/commit/44d10c80d8c2ae11ad52c113d397e619da68095c

You can pick this to your branch/PR before we merge.

rleigh-electradx: @rolandreichweinbmw Thanks, I have cherry-picked this commit.

jwellbelove (approved):

include/etl/array.h
rleigh-electradx: This failed to push due to a re-merge of development.  I corrected this now.

rolandreichweinbmw: I added it as a follow-up PR here:
https://github.com/ETLCPP/etl/pull/1566

PR #1554, opened 2026-08-18
Fix PowerPC Dockerfile to use Snapshot consistently

christophruethingbmw:
Using Debian sid as a base image and updating to latest package sources via `apt-get update` was bound to fail sooner or later in conjunction with using powerpc from debian-ports (not sid but older), getting stuck in dependency conflicts.

Now, we use a consistent (hash-pinned) version of sid, and only updating from the respective known-working snapshot.


rolandreichweinbmw (approved):
This fixes the build issues in the other PRs on the powerpc cross build + checks.

drewr95 (approved):

jwellbelove (approved):

PR #1556, opened 2026-08-19
circular_buffer: add initialization and assignment overloads

drewr95:
## Summary
Implemented count/value constructors and assign overloads for internal and external circular buffers.

Excess elements retains the newest elements.

## Context
I posted a [discussion](https://github.com/ETLCPP/etl/discussions/1523) awhile back that I thought this would be useful but got no other comments.  So I figured I'd add these and see if others, like myself, want these additions.

### DIFF
--- a/include/etl/circular_buffer.h
+++ b/include/etl/circular_buffer.h
@@ -901,6 +901,22 @@ namespace etl
     }
 
     //*************************************************************************
+    /// Assigns an iterator range to the circular buffer.
+    //*************************************************************************
+    void assign(iterator first, const iterator& last)
+    {
+      assign_range(first, last);
+    }
+
+    //*************************************************************************
+    /// Assigns a const iterator range to the circular buffer.
+    //*************************************************************************
+    void assign(const_iterator first, const const_iterator& last)
+    {
+      assign_range(first, last);
+    }
+
+    //*************************************************************************
     /// Assigns a range to the circular buffer.
     //*************************************************************************
     template <typename TIterator>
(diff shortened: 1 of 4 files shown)


include/etl/circular_buffer.h
```diff
@@ -900,6 +900,29 @@ namespace etl
...
+    //*************************************************************************
+    /// Assigns 'n' copies of a value to the circular buffer.
+    //*************************************************************************
+    void assign(size_type n, const_reference value)
+    {
+      clear();
+
+      while (n-- != 0U)
+      {
+        push(value);
```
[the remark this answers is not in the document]
drewr95: I’ve addressed both undefined-behavior cases.

assign(n, value) now makes a defensive copy of value before clearing the buffer, so references such as data.front() remain safe.

Iterator and const-iterator self-ranges are now handled in place without destroying retained elements before reading them. This supports complete, partial, and physically wrapped self-ranges. Non-self ranges continue through the normal clear-and-copy path.

I added regression tests for both fixed and external buffers covering:

- data.assign(data.begin(), data.end())
- Partial wrapped self-ranges using const iterators
- data.assign(n, data.front()) with a non-trivial element type

include/etl/circular_buffer.h
```diff
@@ -1127,6 +1211,24 @@ namespace etl
     {
     }
 
+    //*************************************************************************
+    /// Constructor, with size.
+    //*************************************************************************
+    explicit circular_buffer(size_t initial_size)
```
jwellbelove, line 1217: These new constructors need to be documented in `etl\docs\containers\queues & stacks\circular-buffer.md`

drewr95: Good catch, I also documented the assign methods.

jwellbelove (approved):

PR #1560, opened 2026-08-21
Set default compare for iconst_map

HazardyKnusperkeks:
This is consistent with (i)flat_map, and allows to just write iconst_map<K, V>.

### DIFF
--- a/include/etl/const_map.h
+++ b/include/etl/const_map.h
@@ -48,7 +48,7 @@ SOFTWARE.
 
 namespace etl
 {
-  template <typename TKey, typename TMapped, typename TKeyCompare>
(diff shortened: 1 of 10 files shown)


rolandreichweinbmw (commented):
Looks good, generally!

Adding a small test would be helpful.

A simple compile time check to demonstrate the interface syntax would be enough. E.g.

static_assert(etl::is_same<etl::iconst_map<Key, int>, etl::iconst_map<Key, int, etl::less<Key>>>::value, "...");

HazardyKnusperkeks:
Didn't really know where to put it. I hope this is acceptable.
Also adapted `iconst_set`, there it was missing too. The `multi` variants had the default.

rolandreichweinbmw (approved):

PR #1561, opened 2026-08-21
Allow easier construction of const_(multi_)map

HazardyKnusperkeks:
When forcing to feed into value_types it has to be an ETL_OR_STD::pair<const K, V>, but when you want to use CTAD on the pairs the const is missing and the assert triggers.

### DIFF
--- a/test/test_const_map.cpp
+++ b/test/test_const_map.cpp
@@ -257,6 +257,27 @@ namespace
       CHECK_TRUE(data.max_size() == Max_Size);
       CHECK_FALSE(data.begin() == data.end());
     }
+
+    //*************************************************************************
+    TEST(test_cpp17_deduced_constructor_of_plain_pair)
+    {
+      static const etl::const_map   data{ETL_OR_STD::pair{Key('A'), 0}, ETL_OR_STD::pair{Key('B'), 1}, ETL_OR_STD::pair{Key('C'), 2},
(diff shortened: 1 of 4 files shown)


test/test_const_map.cpp
```diff
@@ -257,6 +257,27 @@ namespace
...
     }
+
+    //*************************************************************************
+    TEST(test_cpp17_deduced_constructor_of_plain_pair)
+    {
+      static const etl::const_map   data{ETL_OR_STD::pair{Key('A'), 0}, ETL_OR_STD::pair{Key('B'), 1}, ETL_OR_STD::pair{Key('C'), 2},
+                                       ETL_OR_STD::pair{Key('D'), 3}, ETL_OR_STD::pair{Key('E'), 4}, ETL_OR_STD::pair{Key('F'), 5},
+                                       ETL_OR_STD::pair{Key('G'), 6}, ETL_OR_STD::pair{Key('H'), 7}, ETL_OR_STD::pair{Key('I'), 8},
+                                       ETL_OR_STD::pair{Key('J'), 9}};
+      etl::const_map<Key, int, 10U> check{ETL_OR_STD::pair{Key('A'), 0}, ETL_OR_STD::pair{Key('B'), 1}, ETL_OR_STD::pair{Key('C'), 2},
```
rolandreichweinbmw, line 268: A little bit odd since `check` is unused in this scope, and inconsistency between static `data`, and non-static `check`. Consistent with the pre-existing test, so ok in this case.

rolandreichweinbmw (approved):

jwellbelove (approved):

PR #1564, opened 2026-08-22
Update etl::expected documentation

myrkle11:
I updated the documentation for expected to reflect the current state of the code. This meant adding the transform(), and_then(), or_else(), and transform_error() methods. Also added begin() and end() which weren't in the docs, and fixed a couple of formatting errors and a typo.

I noticed that none of the transform, etc. methods have the ETL_NODISCARD, ETL_CONSTEXPR14, or ETL_NOEXCEPT defines in the source code. At least the constexpr ought to be there, right? Anyway, this is just a doc PR so I didn't pursue that.


rolandreichweinbmw:
Thanks for the documentation fixes. coderabbitai raised some valid points. Can you please address them?

jwellbelove (approved):

Issue #894, opened 2024-05-24
Add an integrator to the library


drewr95:
I didn't see one when searching your library, if I missed it my apologies.

I have implemented my own that I think would do well here :).  It is very similar to _cyclic_value_ except it's clamped to the min and max range.  It also has a compile-time MIN and MAX similar to cyclic value to save RAM.

Let me know if you'd like me to add it.

jwellbelove:
Could you show an example of how it would be used?

drewr95:
> Could you show an example of how it would be used?

Sure thing.  I'll just paste the main incrementing and decrementing unittests.  One for compile-time known min and maxes and run-time:

```c++
/**
 * @brief Tests preincrementing the integrator.
 */
TYPED_TEST(CompileTimeIntegratorTest, PreIncrement)
{
    constexpr TypeParam min{2};
    constexpr TypeParam max{10};
    common::Integrator<TypeParam, min, max> integrator{min};
    for (auto i = min; i < max; ++i)
    {
        EXPECT_EQ(i + 1, ++integrator);
    }
    EXPECT_EQ(max, integrator);

    // Try to increment past max.
    EXPECT_EQ(max, ++integrator);
}

/**
 * @brief Tests postincrementing the integer.
 */
TYPED_TEST(CompileTimeIntegratorTest, PostIncrement)
{
    constexpr TypeParam min{2};
    constexpr TypeParam max{10};
    common::Integrator<TypeParam, min, max> integrator{min};
    for (auto i = min; i < max; ++i)
    {
        EXPECT_EQ(i, integrator++);
    }
    EXPECT_EQ(max, integrator);

    // Try increment past max.
    integrator++;
    EXPECT_EQ(max, integrator);
}

/**
 * @brief Tests predecrementing the integrator.
 */
TYPED_TEST(CompileTimeIntegratorTest, PreDecrement)
{
    constexpr TypeParam min{2};
    constexpr TypeParam max{10};
    common::Integrator<TypeParam, min, max> integrator{max};
    for (auto i = max; i > min; --i)
    {
        EXPECT_EQ(i - 1, --integrator);
    }
    EXPECT_EQ(min, integrator);

    // Try to decrement past min.
    EXPECT_EQ(min, --integrator);
}

/**
 * @brief Tests the special handling when the integrator min is 0.  This is a
 * special case because the value can't be subtracted normally when the type is
 * unsigned.  This is because of unsigned integeger wrapping back to the max
 * value.
 */
[...]

PR #1565, opened 2026-08-23
Implement clamped_value utility

drewr95:
## Summary
This addresses (at least partially) issue #894 

This pull request introduces the new `clamped_value` utility to the codebase, along with comprehensive documentation, tests, and build system integration. The main focus is on providing a value type that is always constrained between specified minimum and maximum bounds, supporting both compile-time and runtime configuration. The changes also ensure that `clamped_value` is included in all relevant test and syntax check builds.

**Addition of the `clamped_value` utility:**

* Added documentation for `clamped_value` in `docs/utilities/clamped-value.md`, detailing its API, usage, constructors, modifiers, accessors, operators, and operations.
* Integrated the `clamped_value` header (`etl/clamped_value.h`) into the codebase and build/test systems, including CMake, Meson, and Visual Studio project files. [[1]](diffhunk://#diff-18a7d1cd045954d8309900e7c5757a9fd984cfbacb1fba5828f9a06bc0ca4cd6R64) [[2]](diffhunk://#diff-248415ab710ccd66a6ef1e64ecb43c65065275cdcba07b879beec6a377272cd2R3768) [[3]](diffhunk://#diff-248415ab710ccd66a6ef1e64ecb43c65065275cdcba07b879beec6a377272cd2R6307) [[4]](diffhunk://#diff-248415ab710ccd66a6ef1e64ecb43c65065275cdcba07b879beec6a377272cd2R10451) [[5]](diffhunk://#diff-4b570dde260b00bb48119818f7fe598bd16a79ec8ec38958d8f1f6d12145ccd3R267-R269)

**Testing and validation:**

* Added a comprehensive unit test suite for `clamped_value` in `test/test_clamped_value.cpp`, covering initialization, clamping behavior, operators, and edge cases.
* Added a syntax check test for `clamped_value` in `test/syntax_check/clamped_value.h.t.cpp` to verify basic compilation and usage.

**Build system updates:**

* Registered `test_clamped_value.cpp` and `clamped_value.h.t.cpp` in all test runners and build scripts (`test/CMakeLists.txt`, `test/meson.build`, `test/syntax_check/CMakeLists.txt`, and Visual Studio project/filter files). [[1]](diffhunk://#diff-33394812ba204689144fd2f80832db83853ba1cb32403edb4e15fe4893e675fdR167) [[2]](diffhunk://#diff-fb327c87db6b2c02c67999cd120ccf98eb09d654b487f2612366306f9d6ffd9cR155) [[3]](diffhunk://#diff-ba1c12c2b2f53e9a892902a3943a34ffca38ee972c6f6115843fd8b04f1d63cfR206) [[4]](diffhunk://#diff-248415ab710ccd66a6ef1e64ecb43c65065275cdcba07b879beec6a377272cd2R10451) [[5]](diffhunk://#diff-4b570dde260b00bb48119818f7fe598bd16a79ec8ec38958d8f1f6d12145ccd3R2096-R2098) [[6]](diffhunk://#diff-248415ab710ccd66a6ef1e64ecb43c65065275cdcba07b879beec6a377272cd2R6307) [[7]](diffhunk://#diff-4b570dde260b00bb48119818f7fe598bd16a79ec8ec38958d8f1f6d12145ccd3R2921-R2923)

These changes ensure that `clamped_value` is fully documented, tested, and integrated into the project's build and validation pipelines.


include/etl/clamped_value.h
```diff
@@ -0,0 +1,556 @@
...
+    {
+      value = Max;
+    }
+
+    //*************************************************************************
+    /// Advances to value by a number of steps.
+    /// Saturates at the Min/Max range.
+    ///\param n The number of steps.
+    //*************************************************************************
+    ETL_CONSTEXPR14 void advance(int n) ETL_NOEXCEPT
```
serges147: O(n) complexity of this `advance` is worrying... what if I want advance by e.g. 1_000_000?

include/etl/clamped_value.h
```diff
@@ -0,0 +1,556 @@
...
+#include "platform.h"
+#include "algorithm.h"
+
+///\defgroup clamped_value clamped_value
+/// Provides a value that is clamped between two limits.
+/// \ingroup utilities
+
+namespace etl
+{
+  template <typename T, T Min = 0, T Max = 0, bool EtlRuntimeSpecialisation = ((Min == 0) && (Max == 0))>
```
rolandreichweinbmw: Maybe we don't need "Etl" as Prefix in Parameter names?

test/syntax_check/clamped_value.h.t.cpp
```diff
@@ -0,0 +1,41 @@
...
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
+******************************************************************************/
+
+#include <etl/clamped_value.h>
+
+void test_clamped_value()
```
rolandreichweinbmw: Convention here is to just include the header, and leave actual tests to the unit tests.

This should cover all relevant cases, or do you need this since the unit test build doesn't cover certain non-ETL_IN_UNIT_TEST case or similar?

include/etl/clamped_value.h
```diff
@@ -0,0 +1,556 @@
...
+    /// Constructor.
+    /// Sets the value to the minimum of the range.
+    ///\param min_ The minimum value.
+    ///\param max_ The maximum value.
+    //*************************************************************************
+    ETL_CONSTEXPR clamped_value(T min_, T max_) ETL_NOEXCEPT
+      : value(min_)
+      , min_value(min_)
+      , max_value(max_)
+    {
```
rolandreichweinbmw, line 504: Add an ETL_ASSERT(Min <= Max, ...)?

Same for set(min, max) below.

include/etl/clamped_value.h
```diff
@@ -0,0 +1,556 @@
...
+  public:
+
+    //*************************************************************************
+    /// Default constructor.
+    /// Sets the value and bounds to the template parameter values.
+    //*************************************************************************
+    ETL_CONSTEXPR clamped_value() ETL_NOEXCEPT
+      : value(Min)
+      , min_value(Min)
+      , max_value(Max)
```
rolandreichweinbmw: Are you sure this is the right / expected behavior? Then, it could be tested in test_run_time_initialisation() for min() == 0 and max() == 0, to be added.

Otherwise, maybe better use numeric_limits<T> bounds for defaults?

include/etl/clamped_value.h
```diff
@@ -0,0 +1,556 @@
...
+
+    //*************************************************************************
+    /// Constructor.
+    /// Set to an initial value.
+    /// Clamped to the range.
+    ///\param initial The initial value.
+    //*************************************************************************
+    ETL_CONSTEXPR14 explicit clamped_value(T initial) ETL_NOEXCEPT
+    {
+      set(initial);
```
rolandreichweinbmw: By not initializing the `value` member, this is ill-formed in C++14 constant evaluation and rists -Wmaybe-uninitialized. Maybe Init `value` first.

Same for 3-arg runtime ctor, below.

include/etl/clamped_value.h
```diff
@@ -0,0 +1,556 @@
...
+      {
+        ++(*this);
+        --n;
+      }
+
+      while ((n < 0) && (value > Min))
+      {
+        --(*this);
+        ++n;
+      }
```
rolandreichweinbmw: This implements O(n), but could be O(1) by using addition/subtraction.

include/etl/clamped_value.h
rolandreichweinbmw: When T is 64 bit, `int` is quite a limitation here. Maybe find a more general solution via difference_type / make_signed?

include/etl/clamped_value.h
```diff
@@ -0,0 +1,556 @@
...
+    //*************************************************************************
+    ETL_CONSTEXPR T get() const ETL_NOEXCEPT
+    {
+      return value;
+    }
+
+    //*************************************************************************
+    /// Gets the minimum value.
+    //*************************************************************************
+    static ETL_CONSTEXPR T min() ETL_NOEXCEPT
```
rolandreichweinbmw: Since this is `static` but in the runtime implementation below, it's `const`. Limits generic use. Better align both?

include/etl/clamped_value.h
```diff
@@ -0,0 +1,556 @@
...
+    {
+      lhs.swap(rhs);
+    }
+
+    //*************************************************************************
+    /// Operator ==.
+    //*************************************************************************
+    friend ETL_CONSTEXPR bool operator==(const clamped_value& lhs, const clamped_value& rhs) ETL_NOEXCEPT
+    {
+      return (lhs.value == rhs.value) && (lhs.min_value == rhs.min_value) && (lhs.max_value == rhs.max_value);
```
rolandreichweinbmw: There is an asymmetry here since in the compile time's implementation, only `value` is compared.

include/etl/clamped_value.h
```diff
@@ -0,0 +1,556 @@
...
+    ETL_CONSTEXPR14 operator T() ETL_NOEXCEPT
+    {
+      return value;
+    }
+
+    //*************************************************************************
+    /// Const conversion operator.
+    /// \return The value of the underlying type.
+    //*************************************************************************
+    ETL_CONSTEXPR operator const T() const ETL_NOEXCEPT
```
rolandreichweinbmw: Maybe this is redundant with the above `ETL_CONSTEXPR14 operator T() ETL_NOEXCEPT` since returning a const-qualified prvalue is a no-op and trips -Wignored-qualifiers on some compilers. `cyclic_value` also has the single one.

include/etl/clamped_value.h
```diff
@@ -0,0 +1,556 @@
...
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
+******************************************************************************/
+
+#ifndef ETL_CLAMPED_VALUE_INCLUDED
+#define ETL_CLAMPED_VALUE_INCLUDED
+
+#include "platform.h"
+#include "algorithm.h"
+
```
rolandreichweinbmw, line 41: Maybe also include static_assert.h for ETL_STATIC_ASSERT, and utility.h for swap()

include/etl/clamped_value.h
```diff
@@ -0,0 +1,556 @@
...
+      return value;
+    }
+
+    //*************************************************************************
+    /// ++ operator.
+    /// Saturates at Max.
+    //*************************************************************************
+    ETL_CONSTEXPR14 clamped_value& operator++() ETL_LVALUE_REF_QUALIFIER ETL_NOEXCEPT
+    {
+      value = etl::clamp(static_cast<T>(value + (value < Max)), Min, Max);
```
rolandreichweinbmw: This semantics and implementation limits T to integral types. Can we ensure this via static_assert for the class or similar?

rolandreichweinbmw (commented):
Test coverate: How about:
* negative Min
* unsigned runtime range
* reversed bounds
* wide T (int64_t, ...)

And how about API:
* relational operators
* comparison against T
* operator+= / -=
* ETL_NODISCARD on get(), min(), max()

include/etl/clamped_value.h
jwellbelove: Maybe we could do with a type trait that detects whether a type can be incremented and/or decremented, or stepped by an arbitrary amount?

Basically, detecting if operator--, operator++, operator+, or operator- exist.

include/etl/clamped_value.h
drewr95: removed

test/syntax_check/clamped_value.h.t.cpp
drewr95: Fixed.

include/etl/clamped_value.h
drewr95: Good suggestions, addressed.

include/etl/clamped_value.h
drewr95: Yeah I forgot about that, thank you for catching that.

include/etl/clamped_value.h
drewr95: Another good suggestion.  My initial implementation was to make this closely similar to `cyclic_value`.  Applied your suggestion.

include/etl/clamped_value.h
drewr95: See https://github.com/ETLCPP/etl/pull/1565#discussion_r3842494720

include/etl/clamped_value.h
drewr95: There is no need to compare against min and max for the compile-time version because they are a part of the type and cannot be compared to a different type (a compile time version with different min and max).

include/etl/clamped_value.h
drewr95: You are correct. I removed the redundant version

include/etl/clamped_value.h
drewr95: Good suggestion, applied.

include/etl/clamped_value.h
drewr95: Yes, I now static_assert it :)

include/etl/clamped_value.h
rolandreichweinbmw: Since we have conversion operators to `T`, a comparison still compiles and compares the plain values directly.

You can add a test for that. E.g.

```
    TEST(test_different_compile_time_instances)
    {
      etl::cyclic_value<int, 2, 7> value1(3);
      etl::cyclic_value<int, 2, 8> value2(3);
      CHECK(value1 == value2);
    }
```

Now, we still have a conflict between the compile-time and runtime implementation. I recommend aligning by either
* Comparing only the value in both cases
* Or changing the comparison operator for the compile time version to take a differently parameterized Min/Max setup for comparing with other compile-time cyclic_values with different parameters

include/etl/clamped_value.h
drewr95: I agree in this case it's better to align both.  I removed the static.

include/etl/clamped_value.h
drewr95: I understand now, thank you for explaining.  I think it's best for now to compare only by value and keep them both aligned.

include/etl/clamped_value.h
drewr95: I addressed this.  See: https://github.com/drewr95/etl/blob/87f60035f13008d68f0b98bc6b7c4eaeb4667511/include/etl/clamped_value.h#L64-L94

include/etl/clamped_value.h
```diff
@@ -0,0 +1,874 @@
...
+    /// Operator <.
+    ///\param lhs The left-hand operand.
+    ///\param rhs The right-hand operand.
+    ///\return `true` if lhs is less than rhs.
+    /// @{
+    friend ETL_CONSTEXPR bool operator<(const clamped_value& lhs, const clamped_value& rhs) ETL_NOEXCEPT
+    {
+      return lhs.value < rhs.value;
+    }
+    friend ETL_CONSTEXPR bool operator<(const clamped_value& lhs, T rhs) ETL_NOEXCEPT
```
rolandreichweinbmw, line 790: The class has an implicit operator T() const and friend operator< / <= / > / >= overloads taking a bare T. Whenever the other operand isn't exactly T, the friend and the built-in operator (reached via operator T()) are both viable and neither wins:

```
etl::clamped_value<uint8_t, 0, 200> v(10);
bool a = (v < 20);    // error
etl::clamped_value<int, 0, 200> w(10);
bool b = (w < 20L);   // error
bool c = (w < 20u);   // error
```

* Clang: hard error, use of overloaded operator '<' is ambiguous.
* GCC: error: ISO C++ says that these are ambiguous... under -Werror (which CI uses).
So clamped_value is unusable with relational operators for every type narrower than int (int8_t/uint8_t/int16_t/uint16_t/char), and for any literal whose type differs from T. The unit tests miss it because they only use T = int with plain int literals. Maybe add respective tests?

Note operator==/!= deliberately have no mixed overloads and work fine through operator T() — and cyclic_value (the model for this class) likewise only declares ==/!=. The consistent fix is to delete the six mixed T relational overloads and let operator T() handle mixed comparisons, keeping only the clamped_value/clamped_value forms. (Making operator T() explicit would also work but breaks the documented conversion behaviour and the existing tests.)

Sorry for the confusion from the previous suggestion for mixed comparisons.

include/etl/clamped_value.h
```diff
@@ -0,0 +1,874 @@
...
+
+    T value;
+  };
+
+  //***************************************************************************
+  /// Provides a value that is clamped between two runtime limits.
+  /// Supports incrementing, decrementing and arbitrary advance.
+  ///\tparam T   The type of the value.
+  ///\tparam Min The default minimum value.
+  ///\tparam Max The default maximum value.
```
rolandreichweinbmw, line 470: Min and Max documentation is copy&paste error: Actually ignored in the runtime implementation.

include/etl/clamped_value.h
```diff
@@ -0,0 +1,874 @@
...
+    typedef typename private_clamped_value::traits<T>::limits_type            limits_type;
+    typedef typename private_clamped_value::traits<T>::difference_limits_type difference_limits_type;
+
+    //*************************************************************************
+    /// Default constructor.
+    /// Sets the value to the minimum representable value of `T` and the
+    /// bounds to the full range of `T`.
+    //*************************************************************************
+    ETL_CONSTEXPR clamped_value() ETL_NOEXCEPT
+      : value(limits_type::lowest())
```
rolandreichweinbmw, line 488: Why not using min()? I can imagine this stems from the idea of supporting floating point types, but at the same time, you ensure integral type via ETL_STATIC_ASSERT().

include/etl/clamped_value.h
```diff
@@ -0,0 +1,874 @@
...
+    ETL_NODISCARD ETL_CONSTEXPR14 T advance(T value, T min_value, T max_value, typename traits<T>::difference_type n) ETL_NOEXCEPT
+    {
+      typedef typename traits<T>::unsigned_type unsigned_type;
+
+      if ((n > 0) && (value < max_value))
+      {
+        const unsigned_type distance = static_cast<unsigned_type>(max_value) - static_cast<unsigned_type>(value);
+        const unsigned_type step     = static_cast<unsigned_type>(n);
+
+        value = (step >= distance) ? max_value : static_cast<T>(static_cast<unsigned_type>(value) + step);
```
serges147, line 82: IMO these `static_cast<unsigned_type>(xxx)` (all of them except with `n`) could overflow. Do you have test for e.g. `clamped_value<int8_t, -128, -10>` ?

rolandreichweinbmw (commented):
Since you mentioned DRY: Instead of implementing nearly everything twice (compile-time vs. runtime versions), you could implement a common base including value, and specialize just regarding min() / max().

Further, you can still improve test coverage:
* operator-=(difference_limits_type::lowest()) — the special-cased ETL_UNLIKELY branch in both specialisations is never exercised.
* operator+=/-= for runtime version
* .swap()
* ETL_ASSERT(min_ <= max_) paths in runtime implementation
* Types narrower than int

And:
* swap is the only operation not marked ETL_CONSTEXPR14; operator T() lacks ETL_NODISCARD while get() has it.
* The runtime 3-arg ctor initialises value(initial) in the mem-init list and then calls set(initial) — a redundant double initialisation.
* New entries in CMakeLists.txt, meson.build, CMakeLists.txt and the vcxproj are placed after cyclic_value, breaking the alphabetical ordering of those lists (clamped < cyclic). maincpp03check.cpp is correctly sorted.
* Docs synopsis shows template <typename T, const T Min = 0, const T Max = 0>, omitting the 4th (dispatch) parameter and adding a const the header doesn't have.

include/etl/clamped_value.h
drewr95: Yes, you are correct.  I used `lowest()` because it's what I default to.  In this case, I'd like to keep it because I'm not convinced long-term that this implementation is final.  Float values should be able to be clamped, I was just trying to avoid a huge PR and thought integral type clamping was easiest path forward and then opening up a refactor PR to add support for floating point types.  

If you really want `min()` let me know and I can make the change in my next batch.

include/etl/clamped_value.h
drewr95: AFAIK unsigned types are a safe way to deal with overflow.  I will double check and see if I can add tests to stress it.

include/etl/clamped_value.h
rolandreichweinbmw: To prevent others from stumbling about the same question that I found, you can intentionally support (and test :-) ) float types, while implementing/enabling operator++/-- etc. accordingly.

drewr95: Okay, I will add support for float types in this PR.  Thank you again for your reviews

include/etl/clamped_value.h
```diff
@@ -0,0 +1,874 @@
...
+    //*************************************************************************
+    /// Advances a value and saturates it at the supplied bounds.
+    ///\param value The current value.
+    ///\param min_value The minimum value.
+    ///\param max_value The maximum value.
+    ///\param n The number of steps.
+    ///\return The advanced value.
+    //*************************************************************************
+    template <typename T>
+    ETL_NODISCARD ETL_CONSTEXPR14 T advance(T value, T min_value, T max_value, typename traits<T>::difference_type n) ETL_NOEXCEPT
```
jwellbelove, line 73: You could optimise this a little.
```
    template <typename T>
    ETL_NODISCARD ETL_CONSTEXPR14 T advance(T value, T min_value, T max_value, typename traits<T>::difference_type n) ETL_NOEXCEPT
    {
      typedef typename traits<T>::unsigned_type unsigned_type;

      if (n == 0)
      {
        return value;
      }

      const unsigned_type current = static_cast<unsigned_type>(value);
      const unsigned_type step    = etl::absolute_unsigned(n);

      if ((n > 0) && (value < max_value))
      {
        const unsigned_type distance = static_cast<unsigned_type>(max_value) - current;

        value = (step >= distance) ? max_value : static_cast<T>(current + step);
      }
      else if ((n < 0) && (value > min_value))
      {
        const unsigned_type distance = current - static_cast<unsigned_type>(min_value);

        value = (step >= distance) ? min_value : static_cast<T>(current - step);
      }

      return value;
    }
  } 
```

My research so far indicates that your casting technique should work fine for all values.

include/etl/clamped_value.h
jwellbelove: My research so far indicates that your casting technique should work fine for all values.
I tried 
```
etl::clamped_value<int8_t, -128, 127> value(-50);
value += 10;
value -= 10;
```
And got correct results.

include/etl/clamped_value.h
drewr95: Yeah but I probably need to revisit this or specialize it for the floating point support that @rolandreichweinbmw requested

jwellbelove: Fair enough.
I may borrow some of this code to optimise the `advance` function for `etl::cyclic_value` when you've finished this PR.

drewr95: Okay sounds good. I plan on getting back into this, this week. Feel free to take over this implementation if you want it done sooner :)

include/etl/clamped_value.h
drewr95: @rolandreichweinbmw So i'm looking into supporting float types, but doing this with pre-C++20 support means that I can't have the template T types as floats.  Best I can think of is a `runtime_clamped_value<T>` that `clamped_value<>` can use internally.  

@jwellbelove 

I'd be curious in your thoughts or suggestions about that.

rolandreichweinbmw: As you already found, there is currently no float support at all in clamped_value before C++20 due to the non-template type parameter float only supported in C++20. For pragmatic reasons, maybe limit float support to C++20, and support the rest in C++14 as proposed?

drewr95: Yeah I'll do that

docs/utilities/clamped-value.md
```diff
@@ -0,0 +1,173 @@
...
+## Operators
+
+```cpp
+ETL_NODISCARD operator T() const noexcept;
+clamped_value& operator++() & noexcept;
+clamped_value operator++(int) noexcept;
+clamped_value& operator--() & noexcept;
+clamped_value operator--(int) noexcept;
+clamped_value& operator=(T value) &
+  ETL_NOEXCEPT_IF(ETL_NOT_USING_EXCEPTIONS);
```
jwellbelove: This gives the same result as just using `ETL_NOEXCEPT`.

If `ETL_NOT_USING_EXCEPTIONS == 1` then `ETL_NOEXCEPT_IF` is defined as 'nothing'.

jwellbelove: OK, it seems that I was wrong about that one.

drewr95: Maybe it should though?

jwellbelove: OK, I understand it now.

The change I tried broke the tests because `ETL_USING_EXCEPTIONS` == `1`, therefore  `ETL_NOEXCEPT` is defined as `noexcept` for >=C++11, which fails in the tests because `ETL_ASSERT` calls `throw`.

For the functions with `ETL_NOEXCEPT_IF(ETL_NOT_USING_EXCEPTIONS)`:-

- If exceptions are used then the `noexcept` should not be present.
- If exceptions are *not* used then the `noexcept` is redundant and should not be present. (compilers can produce warnings or errors in this situation)

Therefore the `ETL_NOEXCEPT_IF(ETL_NOT_USING_EXCEPTIONS)` should be removed completely.

jwellbelove: There is a third option that is not supported at the moment.

- If exceptions are used but `error_handler` does *not* throw exceptions, then `noexcept` *should* be present.

drewr95: I also believe there is a bit of a formatting issue with one of the advance functions last I looked

jwellbelove: I'll check and fix if necessary.
I shall be releasing `20.49.0` today.

jwellbelove: `clang-format` seems to want to force this line to have loads of spaces.

`static ETL_CONSTEXPR14 T                         advance(T value, T min_value, T max_value, typename traits<T>::difference_type n) ETL_NOEXCEPT`

drewr95: Regarding 20.49.0, did you already optimize `cyclic_value::advance` or is that coming at a later date?

PR #1570, opened 2026-08-25
Fix FreeRTOS.h/semphr.h include paths for ESP-IDF (fixes #1439)

### LINKED ISSUE #1439: mutex_freertos.h fails to compile on ESP-IDF / Arduino-ESP32 — <semphr.h> not found
See: https://github.com/ETLCPP/etl-arduino/issues/5

hiten011:
## Problem

`mutex_freertos.h` unconditionally includes the vanilla FreeRTOS header layout (`FreeRTOS.h` and `semphr.h` at the include root). On ESP-IDF and therefore Arduino-ESP32 too, since it's built on top of ESP-IDF these headers are nested one level deeper under `freertos/`. The flat path doesn't exist there, so any ESP32 target with `ETL_TARGET_OS_FREERTOS` defined fails to compile.

## Fix

Use `__has_include` to try the standard path first, fall back to the ESP-IDF nested path, and fail with a clear `#error` if neither is found

## Others

Also resolves a pre-existing include-style inconsistency in this file: the old code mixed "FreeRTOS.h" (quotes) with <semphr.h> (angle brackets). All includes now uniformly use angle brackets.

Confirmed via a full-repo search that `mutex_freertos.h` is the only file in `etl` that includes a real FreeRTOS header, self-contained fix, nothing else to update.

### DIFF
--- a/include/etl/mutex/mutex_freertos.h
+++ b/include/etl/mutex/mutex_freertos.h
@@ -31,8 +31,26 @@ SOFTWARE.
 
 #include "../platform.h"
 
-#include "FreeRTOS.h"
-#include <semphr.h>
+#if defined(__has_include) // Compiler supports __has_include
+
+  #if __has_include(<FreeRTOS.h>) && __has_include(<semphr.h>) // Standard FreeRTOS include layout
+    #include <FreeRTOS.h>
+    #include <semphr.h>
+
+  #elif __has_include(<freertos/FreeRTOS.h>) && __has_include(<freertos/semphr.h>) // ESP-IDF style include layout
+    #include <freertos/FreeRTOS.h>
+    #include <freertos/semphr.h>
+
+  #else // Neither include layout found
+    #error "FreeRTOS.h and semphr.h not found. Please include path to FreeRTOS in your project settings"
+  #endif
+
+#else
+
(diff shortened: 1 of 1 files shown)


include/etl/mutex/mutex_freertos.h
```diff
@@ -31,8 +31,26 @@ SOFTWARE.
...
 
-#include "FreeRTOS.h"
-#include <semphr.h>
+#if defined(__has_include)
+
+  #if __has_include(<FreeRTOS.h>) && __has_include(<semphr.h>)
+    #include <FreeRTOS.h>
+    #include <semphr.h>
+
+  #elif __has_include(<freertos/FreeRTOS.h>) && __has_include(<freertos/semphr.h>)
```
rolandreichweinbmw: Can you please document via comment lines for the respective `#if` and `#elif` which case they cover: Vanilla FreeRTOS, ESP-IDF, Arduino-ESP.

You already write it in the PR and commit message, but it's not obvious when seeing the code.

hiten011: sure

rolandreichweinbmw (approved):

jwellbelove (approved):

PR #1572, opened 2026-08-26
Fixed building and unit tests run under XCode 2027 Beta 6

serges147:
Under XCode 2027 Beta 6 (AppleClang 21.0.0.21000333) ETL build fails with:

- Compilation failures:
  - `etl::ideque::iterator::operator[]` index type should be `difference_type` (instead of `size_t`).
  - random-access `etl::poly_span::iterator` is missing `operator[]`.
- Unit tests failures of `test_multimap` suit:
  - fixed `equal_range`-related tests to properly verify result range (via new `Check_Equal` overload helper).
  - fixed `find`-related tests to not depend on particular order of `second` values (via new `Check_ContainsPair` helper). See note 1,2) at [cppreference](https://cppreference.com/cpp/container/multimap/find) about order.

Here is build log:
```
.../etl/test/../include/etl/deque.h:354:19: error: implicit conversion changes signedness: 'size_t' (aka 'unsigned long') to 'difference_type' (aka 'long') [-Werror,-Wsign-conversion]
  354 |         result += i;
      |                ~~ ^

/Applications/Xcode-beta.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/c++/v1/__algorithm/sift_down.h:48:23: error: type 'etl::private_poly_span::iterator<int>' does not provide a subscript operator
   48 |     __child += __comp(__first[__child], __first[__child + 1]);
      |                       ^~~~~~~ ~~~~~~~
...
/Users/sergei/Develop/git/serges147/etl/test/test_poly_span_dynamic_extent.cpp:412:12: note: in instantiation of function template specialization 'std::sort<etl::private_poly_span::iterator<int>>' requested here
  412 |       std::sort(s.begin(), s.end());
      |            ^

```
And output of unit tests (after fixing the build):
```
.../etl/test/test_multimap.cpp:1037:1: error: Failure in test_find: (i_compare->second == i_data->second)
.../etl/test/test_multimap.cpp:1044:1: error: Failure in test_find: (i_compare->second == i_data->second)
.../etl/test/test_multimap.cpp:1082:1: error: Failure in test_find_using_transparent_comparator: (i_compare->second == i_data->second)
.../etl/test/test_multimap.cpp:1089:1: error: Failure in test_find_using_transparent_comparator: (i_compare->second == i_data->second)
.../etl/test/test_multimap.cpp:1138:1: error: Failure in test_find_const: (i_compare->second == i_data->second)
.../etl/test/test_multimap.cpp:1145:1: error: Failure in test_find_const: (i_compare->second == i_data->second)
.../etl/test/test_multimap.cpp:1183:1: error: Failure in test_find_const_using_transparent_comparator: (i_compare->second == i_data->second)
.../etl/test/test_multimap.cpp:1190:1: error: Failure in test_find_const_using_transparent_comparator: (i_compare->second == i_data->second)
```


include/etl/deque.h
```diff
@@ -357,7 +357,7 @@ namespace etl
       }
 
       //***************************************************
-      const_reference operator[](size_t i) const
+      const_reference operator[](difference_type i) const
```
rolandreichweinbmw: Consider the above:

```
reference operator*() const
```

I.e. when the const overload is selected, a `reference` is returned. For consistency reasons, I would propose this here also, like this:

```
reference operator[](difference_type i) const
```

And at the same remove the `reference operator[](difference_type i)` since it's not necessary.

This is relevant when in code those return values are matched and can result in mismatches at call sites.

Same applies to poly_span.h

rolandreichweinbmw (commented):
The current PR still leaves the inconsistency of keeping `size_t` for parameters of `ideque::const_iterator::operator[]`. Can you adjust it also?

include/etl/deque.h
```diff
@@ -348,7 +348,7 @@ namespace etl
       }
 
       //***************************************************
-      reference operator[](size_t i)
+      reference operator[](difference_type i)
```
jwellbelove: In the STL, `std::deque` it uses an unsigned index
`reference operator[]( size_type pos );`

Allowing a possible negative index does not make any sense for containers.

rolandreichweinbmw: For containers, yes.

Note that this is the index for iterators where a negative value makes much more sense. (l.351 and l.610)

include/etl/deque.h
serges147: Done, please see last commit. Plz note that `poly_span` has only `iterator`; const_iterator` is missing - we should probably file (separate?) defect about it.

rolandreichweinbmw (approved):

include/etl/deque.h
jwellbelove: Ah, I see, I didn't notice it was the iterator's [] operator.

jwellbelove (approved):

PR #1578, opened 2026-08-28
Fix implicit conversion in binary rotations

rafajunio:
Explicitly cast rotated values to their input type before returning.
This prevents UBSan implicit conversion failures when etl::rotl and
etl::rotr operate on narrow unsigned types.

```bash
include/etl/binary.h:152:14: runtime error: implicit conversion from type 'int' of value 259 
(32-bit, signed) to type 'unsigned char' changed the value to 3 (8-bit, unsigned)
    #0 etl::rotate_left<unsigned char>(unsigned char, unsigned long)
       etl/include/etl/binary.h:152:14
    #1 etl::rotl<unsigned char>(unsigned char, int)
       etl/include/etl/bit.h:193:14
    #2 main rotate_narrowing_test.cpp:10:11
```
Test with:
```bash
clang++ -std=c++11 \
        -g -fno-omit-frame-pointer \
        -fsanitize=undefined,implicit-conversion \
        -fno-sanitize-recover=all \
        -Iinclude file/rotate_narrowing_test.cpp \
        -o test
```

### DIFF
--- a/include/etl/binary.h
+++ b/include/etl/binary.h
@@ -123,7 +123,7 @@ namespace etl
 
     ETL_CONSTANT size_t SHIFT = etl::integral_limits<typename etl::make_unsigned<T>::type>::bits - 1U;
 
-    return (value << 1U) | (value >> SHIFT);
+    return static_cast<T>((value << 1U) | (value >> SHIFT));
 #endif
   }
 
@@ -149,7 +149,7 @@ namespace etl
     }
     else
     {
-      return (value << distance) | (value >> SHIFT);
+      return static_cast<T>((value << distance) | (value >> SHIFT));
     }
 #endif
   }
@@ -168,7 +168,7 @@ namespace etl
 
     ETL_CONSTANT size_t SHIFT = etl::integral_limits<typename etl::make_unsigned<T>::type>::bits - 1U;
 
-    return (value >> 1U) | (value << SHIFT);
+    return static_cast<T>((value >> 1U) | (value << SHIFT));
 #endif
   }
 
@@ -194,7 +194,7 @@ namespace etl
     }
     else
     {
-      return (value >> distance) | (value << SHIFT);
+      return static_cast<T>((value >> distance) | (value << SHIFT));
     }
 #endif
   }


rafajunio:
I recommend that all test should run with sanitizer enabled. Something like:
```bash
$CXX -g3 --strip=never -Wall -fsanitize=address,undefined,implicit-conversion,local-bounds,memory ...
```

rolandreichweinbmw:
The previous review discussion refers to additional tests, but another force push is noted also, and in the current state no tests are added.

Can you please check if they can be added, to demonstrate the issue and the fix?

rafajunio:
> The previous review discussion refers to additional tests, but another force push is noted also, and in the current state no tests are added.
> 
> Can you please check if they can be added, to demonstrate the issue and the fix?

You guys already have the test, that is why I removed

- [test_rotl_8](https://github.com/etlcpp/etl/blob/master/test%2Ftest_bit.cpp#L229)
- [test_rotl_16](https://github.com/etlcpp/etl/blob/master/test%2Ftest_bit.cpp#L298)
- [test_rotr_8](https://github.com/etlcpp/etl/blob/master/test%2Ftest_bit.cpp#L403)
- [test_rotr_16](https://github.com/etlcpp/etl/blob/master/test%2Ftest_bit.cpp#L577)

My private project run with [-Weverything](https://clang.llvm.org/docs/UsersManual.html#diagnostics-enable-everything) that enable all diagnostics in `clang`, this problem can also be catch by the warning [-Wconversion](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Wconversion) in `gcc`

But my private project also run all the tests with sanitizers enabled, and then this rotation cause a undefined behavior. 

I did not run all `etl` tests with sanitizers, but those four in particular I isolate in a small binary to put the stack trace in the description.

I recomend run all etl tests with sanitizers and enable more warnings to catch those problems.

rolandreichweinbmw:
Thank you @rafajunio for the explanation.

Looking into the issue more closely, I found that implicit narrowing truncation is well defined in C++. Therefore, the change suggested by you does not fix an actual bug or undefined behavior.

We currently have `-fsanitize=address,undefined,bounds` in our sanitizer checks, as defined in `test/CMakeLists.txt`, not `-fsanitize=implicit-integer-truncation` as suggested by you in addition. One reason is that our current setup is focusing on undefined behavior and bugs. `implicit-integer-truncation` is intentionally separate in the compiler, since it is "only" detecting suspicious but well defined patterns.

Now, there is a trade-off regarding the change discussed here:
- Advantage: Explict definition of intent, no suspicious code anymore, and silences the UBSan finding (if activated eventually)
- Disadvantage: Makes the code more complex, and does not actually fix a bug

@jwellbelove , @drewr95, and colleagues, any opinion on that?

BTW: Currently, the sanitizers are not active at all in the ETL CI checks, and would need some changes in the code before actually being activated. I'm currently working on a branch that addresses this:

https://github.com/rolandreichweinbmw/etl/tree/fix-sanitizers

Will probably set up a PR for that soon.

rafajunio:
@rolandreichweinbmw, actually a more broader sanitizer like `implicit-conversion` also catch this. I change the commit message to adequate to this. 

Unfortunately I will discuss with my team to avoid use `ETL` outside of `etl::array` and `etl::span`. This is a small change for you, but for me really make difference in a critical software.

At the moment we have this current PR as a patch in our internal `ETL` build, but if is that hard to contribute I prefer to drop the usage of the library as much as we can.

This is not a new feature or something, is a small change that make no warn for you, but for me that use full misra, autosar, cert make a huge difference.

Either way thank you.

PR #1579, opened 2026-08-29
Fix sanitizer builds

rolandreichweinbmw:
Besides, remove unused ASAN_OPTIONS from github workflows.

Further, activate sanitizers for C++23 workflows, at the places where we also activated -O3 optimization. Not activating everywhere for CI efficiency reasons for now.

**Note that for delegate, an additional bound member needed to be added, to enable reading of the not null state in constexpr case.**

### DIFF
--- a/include/etl/private/delegate_cpp11.h
+++ b/include/etl/private/delegate_cpp11.h
@@ -492,7 +492,7 @@ namespace etl
     //*************************************************************************
     /// Assignment
     //*************************************************************************
-    delegate& operator=(const delegate& rhs) = default;
+    ETL_CONSTEXPR14 delegate& operator=(const delegate& rhs) = default;
 
     //*************************************************************************
     /// Create from Lambda or Functor.
--- a/.github/workflows/clang-c++11.yml
+++ b/.github/workflows/clang-c++11.yml
@@ -50,7 +50,6 @@ jobs:
       run: |
         export CC=clang
         export CXX=clang++
-        export ASAN_OPTIONS=alloc_dealloc_mismatch=0,detect_leaks=0
         cmake -D BUILD_TESTS=ON -DNO_STL=OFF -DETL_USE_TYPE_TRAITS_BUILTINS=OFF -DETL_USER_DEFINED_TYPE_TRAITS=OFF -DETL_FORCE_TEST_CPP03_IMPLEMENTATION=OFF -DETL_CXX_STANDARD=11 ./
         clang --version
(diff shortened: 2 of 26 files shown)


include/etl/private/delegate_cpp11.h
```diff
@@ -709,6 +720,9 @@ namespace etl
 
       ptr_type  ptr;
       stub_type stub;
+      /// Set when a stub is bound. Avoids inspecting the function pointer,
+      /// which UBSan instrumentation makes non-constexpr.
+      bool bound;
```
jwellbelove: I'm concerned about the increase in size of `etl::delegate`, just to get around a sanitizer issue.
Can we not (conditionally?) locally disable the check around the problem code?

jwellbelove: Is this a GCC only issue?

rolandreichweinbmw: Right. I'm also concerned. With the additional member, just offering one possibility to solve the current broken sanitizer checks.

It is about the sanitizer check which does not work currently with the constexpr function-pointer validity check as `(f != 0)`.

Looks like the additional member is the only design that is simultaneously constexpr-clean, UBSan-clean, and portable.

For instead trying to locally disabling the sanitizer check, I couldn't find a solution, e.g. `__attribute__((no_sanitize("null")))` or `__attribute__((no_sanitize("undefined")))` doesn't work with constant evaluation.

Maybe you have a better idea to somehow fix the current sanitizer breakage?

rolandreichweinbmw: > Is this a GCC only issue?

Yes, the issue is not present with clang.

jwellbelove: I get the `constexpr` issue with GCC even without the `-fsanitizer=null` option.

jwellbelove: I been experimenting for a while, with no success either.
I'm thinking that the GCC issue fix should be a compile time opt-in directive, so that users using GCC can choose to accept the overhead to get the `constexpr` functionality.

I also noticed that `function_ptr_stub` is the only stub not marked `ETL_CONSTEXPR14`.

rolandreichweinbmw:
Until we find a better solution, I'm **reverting** 2 changes from this PR:
* Activation of sanitizers in gcc C++23 workflow
* Adding the `bound` member to delegate

I think the rest of the PR still makes sense. Now, we still have the clang side of it fixed, and testing sanitizers n the clang C++23 workflow.

@jwellbelove , please have a look again.

PR #1580, opened 2026-09-01
Support variable number of arguments to inject at runtime

rolandreichweinbmw:
When you construct arguments programmatically, i.e. don't know the list of arguments already at compile time, you can't use the current interface.

For this, a separate constructor is added to `basic_format_args` to supply the internally used `basic_format_arg` as externally located data as `array_view`.

Providing a test to demonstrate the use also.


include/etl/format.h
```diff
@@ -855,6 +855,15 @@ namespace etl
     {
     }
 
+    // non-standard
+    // Constructs from externally owned storage, for cases where the number of
+    // arguments is only known at runtime and so no format_arg_store<Context, Args...>
+    // can be formed. The referenced storage must outlive this object.
+    explicit basic_format_args(etl::array_view<basic_format_arg<Context> > args_)
```
drewr95: Shouldn't array_view be swapped with a span<> or initializer_list<>?

Also there seems to be an unneeded space between template closing brackets >>

drewr95 (changes requested):

include/etl/format.h
rolandreichweinbmw: Agreed! Adjusted accordingly.

drewr95 (approved):

jwellbelove (approved):

PR #1582, opened 2026-09-01
Fix symbols to be exportable from C++20 named modules

### LINKED ISSUE #1543: ETL span is not exportable from a C++20 named module: dynamic_extent is TU-local
## Summary

`etl::span` cannot be exposed through a C++20 named module interface with a
recent GCC-based Arm GNU Toolchain.

GCC reports that `etl::span` exposes the TU-local entity
`etl::dynamic_extent`.

## Minimal reproducer

```cpp
module;

#include <cstdint>
#include <etl/span.h>

export module sample;

export using Bytes = etl::span<uint8_t>;
```

## Command

```sh
arm-none-eabi-g++ -std=c++23 -fmodules-ts -c sample.cppm
```

## Actual result

```text
In file included from sample.cppm:4:
.../etl/span.h:402:19: error: 'template<class T, unsigned int Extent>
template<class U, unsigned int Size> constexpr etl::span<T, Extent>::span(
const etl::span<U, Size>&, typename etl::enable_if<((Size == Extent) &&
(Size != etl::dynamic_extent)), void>::type*)' exposes TU-local entity
'etl::dynamic_extent'
```

The relevant ETL definition appears to be:

```cpp
namespace etl
{
  static ETL_CONSTANT size_t dynamic_extent =
    etl::integral_limits<size_t>::max;
}
```

Because `dynamic_extent` is namespace-scope `static`, it has internal linkage
and is TU-local. `etl::span` refers to it in public constructor declarations.
Exporting an alias of `etl::span` from a named module therefore exposes a
TU-local entity.

## Expected result

Either:

1. `etl::span` should be usable in a public C++20 named-module interface; or
2. ETL should document that ETL template types cannot currently be exported
   through named modules.

## Environment

- ETL version: 20.47.1
- Language mode: C++23
- Modules flag: `-fmodules-ts`
- Compiler: Arm GNU Toolchain 15.2.Rel1, GCC 15.2.1

```text
arm-none-eabi-g++ (Arm GNU Toolchain 15.2.Rel1 (Build arm-15.86)) 15.2.1 20251203
Copyright (C) 2025 Free Software Foundation, Inc.
```
- Host OS: Kubuntu 24.04
- Target: bare-metal `arm-none-eabi` / RP2040

The same pattern may also affect native GCC builds using the same modules
workflow.

## Possible direction

For C++17-and-newer module builds, an external-linkage definition such as:

```cpp
inline constexpr size_t dynamic_extent =
  etl::integral_limits<size_t>::max;
```

may avoid the TU-local exposure. I understand ETL supports older C++ standards,
so this may need to be conditional rather than replacing the existing
definition unconditionally.

This may not be the only instance that causes this error.  I would very much like to see full module compatibility for the ETL.

Is named-module compatibility supported or planned for ETL?

rolandreichweinbmw:
Fixes #1543

### DIFF
--- a/include/etl/basic_format_spec.h
+++ b/include/etl/basic_format_spec.h
@@ -135,65 +135,65 @@ namespace etl
   //***************************************************************************
   // Stream formatting manipulators.
   //***************************************************************************
-  static ETL_CONSTEXPR private_basic_format_spec::base_spec setbase(uint32_t base)
+  inline ETL_CONSTEXPR private_basic_format_spec::base_spec setbase(uint32_t base)
   {
     return private_basic_format_spec::base_spec(base);
   }
 
   //*********************************
-  static ETL_CONSTEXPR private_basic_format_spec::width_spec setw(uint32_t width)
+  inline ETL_CONSTEXPR private_basic_format_spec::width_spec setw(uint32_t width)
   {
     return private_basic_format_spec::width_spec(width);
   }
 
   //*********************************
   template <typename TChar>
-  static ETL_CONSTEXPR private_basic_format_spec::fill_spec<TChar> setfill(TChar fill)
+  ETL_CONSTEXPR private_basic_format_spec::fill_spec<TChar> setfill(TChar fill)
   {
     return private_basic_format_spec::fill_spec<TChar>(fill);
   }
 
   //*********************************
-  static ETL_CONSTEXPR private_basic_format_spec::precision_spec setprecision(uint32_t precision)
+  inline ETL_CONSTEXPR private_basic_format_spec::precision_spec setprecision(uint32_t precision)
   {
     return private_basic_format_spec::precision_spec(precision);
   }
 
   //*********************************
-  static ETL_CONSTANT private_basic_format_spec::base_spec bin(2U);
+  ETL_INLINE_VAR ETL_CONSTANT private_basic_format_spec::base_spec bin(2U);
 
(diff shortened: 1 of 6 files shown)


include/etl/basic_format_spec.h
```diff
@@ -135,65 +135,65 @@ namespace etl
...
-  static ETL_CONSTEXPR private_basic_format_spec::width_spec setw(uint32_t width)
+  inline ETL_CONSTEXPR private_basic_format_spec::width_spec setw(uint32_t width)
   {
     return private_basic_format_spec::width_spec(width);
   }
 
   //*********************************
   template <typename TChar>
-  static ETL_CONSTEXPR private_basic_format_spec::fill_spec<TChar> setfill(TChar fill)
+  ETL_CONSTEXPR private_basic_format_spec::fill_spec<TChar> setfill(TChar fill)
```
jwellbelove, line 151: Is this missing the `inline`?

rolandreichweinbmw: It is not missing since this line defines a function template for which `inline` is not necessary, in contrast to the non-template functions.

`inline` could be added and would be valid but it would not make a difference technically, and therefore it would be redundant.

jwellbelove: Yes, I missed that it was a template function.

jwellbelove (approved):

PR #1592, opened 2026-09-07
Add be_uint24_t etc. to unaligned_type.h

rolandreichweinbmw:
A typical use case is to endian convert 24, 40, 48 bits etc.

This change generalizes the pre-existing types in unaligned_type.h to those odd sizes.


rolandreichweinbmw:
You don't need to defer the upcoming ETL release for this PR.

include/etl/unaligned_type.h
```diff
@@ -741,164 +797,282 @@ namespace etl
...
+    };
+#endif
+
+    //*************************************************************************
+    /// Sign extends the low 'Size_ * 8' bits of an unsigned value into the
+    /// full width of the signed interface type.
+    /// Uses only well defined unsigned arithmetic, so it is usable in a
+    /// constexpr context.
+    //*************************************************************************
+    template <size_t Size_, typename TSigned, typename TUnsigned>
```
jwellbelove, line 926: Is `etl::sign_extend` not suitable for this?

rolandreichweinbmw: For the higher bit widths >32 the etl::sign_extend does not work well due to the limitation of 32 bit `signed value` which only gets padding bits to fulfill the bigger size, but in computations, data gets lost.

Also, the `constexpr` doesn't work there in this case.

What I propose is to maybe in a follow-up PR work on replacing `etl::sign_extend` with a few compromises. Due to slight semantic differences, we can't just take `private_unaligned_type::sign_extend` and lift it over to `etl::sign_extend`.

PR #1598, opened 2026-09-08
Help the compiler optimize unaligned_type.

denravonska:
This fixes a regression in 266b0f99 when the type was made constexpr. When switching to a loop to construct the value the compiler was unable to detect what was going on. We now unroll the loop at compile time into what's essentially a hand rolled shift and OR operation which the compiler detects and replaces with a byte swap. Tested on gcc (x86+arm) and clang (x86).

Introduce a static `value_from` which can be used to convert a value from an existing buffer. There were previously three options for this, all bad or awkward in different ways:

```
val = etl::be_uint32_t(address)
```
Copies to temporary storage.

```
val = *reinterpret_cast<const etl::be_uint32_t*>(address);
```
Fast but unreadable.

```
etl::be_uint32_t::copy_store_to_value(address, output);
```
Requires an output parameter.

The new addition is:
```
val = etl::be_uint32_t::value_from(address);
```

Additionally, update the documentation with value_from and the missing, existing constructors.

### DIFF
--- a/docs/types/unaligned-type.md
+++ b/docs/types/unaligned-type.md
@@ -167,6 +167,50 @@ Constructs with the supplied value.
 ---
 
 ```cpp
+unaligned_type(const void* address)
+```
+**Description**  
+Constructs from the `Size` bytes at the address.  
+From: `20.39.5`
+
+---
+
+```cpp
+unaligned_type(const void* address, size_t buffer_size)
+```
+**Description**  
+Constructs from the `Size` bytes at the address.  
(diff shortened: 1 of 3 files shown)


jwellbelove (approved):

rolandreichweinbmw:
Thank you for this PR! It's a real improvement!

rolandreichweinbmw (approved):

drewr95 (approved):

denravonska:
I just now realized that I should've done them in separate commits. Sorry, my bad :(

jwellbelove (approved):

PR #1599, opened 2026-09-09
Check const_map is valid

mike919192:
Following up #1483 and https://github.com/ETLCPP/etl/pull/1483#issuecomment-5518615957

Potential solution to #1341 

It was identified that the member function `is_valid` can be used to perform constexpr checking of const_map values.  `is_valid` checks the values are sorted and unique.  If the `is_valid` checks fails, then the call to the non constexpr function causes compile error.  

This check does not have any effect on runtime variables.  For that, a runtime assert could be added for the next step.

After implementing the check, I needed to rework some tests.  Some tests had values that were not sorted, and some have duplicate keys.  The values were corrected and the tests should still be performing the same checks.


include/etl/const_map.h
```diff
@@ -483,6 +483,11 @@ namespace etl
     size_type         max_elements;
   };
 
+  namespace etl_priv
```
jwellbelove: The convention used in the ETL is to create a `private_xxx` nested namespace in the `etl` namespace.

In this case it would be 
```
namespace etl
{
  namespace private_const_map
  {

  }
}
```

Issue #1601, opened 2026-09-11
etl::chrono spaceship operators require <compare> even when ETL_NO_STL is defined


HosseinSagha:
When building with ETL_NO_STL defined (freestanding/no-STL configuration) and C++20 or later, compilation fails with errors like:

`error: 'strong_ordering' is not a member of 'std'`

in etl/private/chrono/day.h, month.h, month_day.h, year.h, year_month.h, year_month_day.h, duration.h, and time_point.h.

Root cause: these headers gate their operator<=> overloads on #if ETL_USING_CPP20 alone. ETL_USING_CPP20 only reflects the language standard, not whether <compare>/std::strong_ordering is actually available. Elsewhere in the library (e.g. ETL_CONSTEXPR20_STL), ETL correctly gates C++20 STL-dependent features on ETL_USING_CPP20 && ETL_USING_STL. The chrono spaceship operators were missed.

Fix: change the guard in each affected file from:

`#if ETL_USING_CPP20`

to:

`#if ETL_USING_CPP20 && ETL_USING_STL`

This matches ETL's existing convention and allows the library to compile in freestanding/ETL_NO_STL environments (e.g. Zephyr RTOS builds) targeting C++20/23.

PR #1602, opened 2026-09-11
Exclude operator <=> from chrono when STL not available

HosseinSagha:
[#1601](https://github.com/ETLCPP/etl/issues/1601)

### DIFF
--- a/include/etl/private/chrono/day.h
+++ b/include/etl/private/chrono/day.h
@@ -211,7 +211,7 @@ namespace etl
     //***********************************************************************
     /// Spaceship operator
     //***********************************************************************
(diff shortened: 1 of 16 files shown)


rolandreichweinbmw (approved):

drewr95 (approved):

jwellbelove:
Could we get around this by defining our own `compare.h` with a compatible set of `std::strong_ordering` etc when `ETL_NO_STL` is defined?

jwellbelove:
Maybe a `ETL_NO_STL` solution should addressed in a separate PR.

jwellbelove (approved):

PR #1604, opened 2026-09-14
Implement spaceship operator <=> for No-STL case

rolandreichweinbmw:
This was disabled explicitly in recent commit
232c948674cff1b4b72c2d6b3acfa7342d096dd1 to prevent using the <compare> header from STL.

Now, re-enabling by implementing an ETL version of it in compare.h

This is a follow-up for https://github.com/ETLCPP/etl/pull/1602


drewr95 (approved):

include/etl/compare.h
```diff
@@ -42,6 +42,44 @@ SOFTWARE.
...
+#if ETL_USING_CPP20 && ETL_NOT_USING_STL
+  //***************************************************************************
+  /// Minimal ordering category for C++20 three-way comparisons when STL is
+  /// unavailable.
+  //***************************************************************************
+  class strong_ordering
+  {
+  public:
+
+    enum value_t
```
jwellbelove: I thought these were supposed (or by convention?) to have the type `strong_ordering`.

jwellbelove:
Should we include definitions for `weak_ordering` and `partial_ordering`?
Should there be tests and docs for this ordering?

include/etl/compare.h
rolandreichweinbmw: This actually _is_ the implementation of `strong_ordering`, isn't it?

rolandreichweinbmw:
> Should we include definitions for `weak_ordering` and `partial_ordering`? Should there be tests and docs for this ordering?

Agree. I updated accordingly.

jwellbelove (approved):

PR #1606, opened 2026-09-15
Simplify noexcept expressions using ETL_NOT_USING_EXCEPTIONS

mike919192:
Follow up to #1542 .  

It was determined that `ETL_NOT_USING_EXCEPTIONS` is always false in `ETL_NOEXCEPT_EXPR` and `ETL_NOEXCEPT_IF`, because when it is true, the macros are defined to empty string.

I did regex search for all `ETL_NOEXCEPT_EXPR` and `ETL_NOEXCEPT_IF` containing `ETL_NOT_USING_EXCEPTIONS`.  In the cases where the logic is or, `ETL_NOT_USING_EXCEPTIONS` is removed.  In the cases where the logic is and, the entire noexcept expression is removed.

### DIFF
--- a/docs/about.md
+++ b/docs/about.md
@@ -1,6 +1,6 @@
 ---
 title: "About"
-weight: 3003
+weight: 3004
 ---
 
 ## About me
--- a/docs/business-and-financial-case-for-commercial-sponsorship.md
+++ b/docs/business-and-financial-case-for-commercial-sponsorship.md
@@ -1,0 +1,59 @@
+---
+title: "Why CFOs should care about Open Source"
+description: "The business and financial case for commercially sponsoring the Embedded Template Library (ETL)."
+date: 2026-09-12
+weight: 3003
+---
+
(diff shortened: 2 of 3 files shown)


include/etl/vector.h
```diff
@@ -393,7 +393,7 @@ namespace etl
     typename etl::enable_if<!etl::is_integral<TIterator>::value
                               && etl::is_convertible<typename etl::iterator_traits<TIterator>::value_type, T>::value,
                             typename etl::enable_if<!etl::is_integral<TIterator>::value, void>::type>::type
-      assign(TIterator first, TIterator last) ETL_NOEXCEPT_IF((etl::is_nothrow_copy_constructible<T>::value && ETL_NOT_USING_EXCEPTIONS))
```
mike919192, line 396: I removed these statements for now because the && logic should mean that it never has any effect.  It could be that maybe the logic was faulty and there is an intended noexcept expression that should be fixed.

rolandreichweinbmw (approved):

drewr95 (approved):

jwellbelove (approved):

Issue #1585, opened 2026-09-02
const_map and default constructible value_type


HazardyKnusperkeks:
Before I open a PR I wanted to start a discussion.

Is there any particular reason for https://github.com/ETLCPP/etl/blob/0723fca18caa4097ff52ef7c045d009241e911a4/include/etl/const_map.h#L506-L507 I can't see it, because the members in the array are initialized inside the constructor and the actual size (in the case someone provides less elements than memory is allocated) is forwarded to `iconst_map` so even if there were unintialized memory, it is not accessed. Also I believe if there wasn't a default constructor, we will be notified by a compile error.

The reason I bring this up is, that we have types, that we want to use as keys, that should not be default constructible, but always have meaningful content. We disabled the assert locally and everything compiles fine (C++23).

jwellbelove:
There is a PR for const_map live at the moment. Maybe you could ask there.

https://github.com/ETLCPP/etl/pull/1483

rolandreichweinbmw:
There is a case where the `static_assert`s are useful: When you have less elements than Size, the `is_default_constructibe` ensures the ability to still compile and work.

To support the use case raised by @HazardyKnusperkeks , we could do:

```
static_assert(sizeof...(elements) == Size ||
              (etl::is_default_constructible<key_type>::value &&
               etl::is_default_constructible<mapped_type>::value),
              "key_type and mapped_type must be default constructible when fewer than Size elements are supplied");
```

At the same time, moving to the constructor where `elements` is known.

In contrast, the `static_assert`s in `const_map_ext` can simply be removed.

rolandreichweinbmw:
Regarding #1483:
* Same probably applies to `const_set`, `const_multimap` and `const_multiset` also
* The PR #1483 will overlap/conflict technically with this change here. But the concerns are separate and both should be implemented. So either way regarding the order of change, merge conflict needs to be resolved, but is expected to be simple.

Doing all in one PR would be reasonable also.

PR #1609, opened 2026-09-22
Do not enforce default constructability for const_(map|set) values

HazardyKnusperkeks:
There is an actual use case to use types, which are not default constructible. Change the assert in a way, that only iff not all elements in the internal array aren't initialized there is a default constructor.

Fixes #1585.


HazardyKnusperkeks:
I've done it just for `const_set` as this is the easiest to discuss if this is feasible. If accepted I will adapt the other `const_` classes too.

include/etl/const_set.h
```diff
@@ -420,6 +418,9 @@ namespace etl
     {
       static_assert((etl::are_all_same<value_type, etl::decay_t<TElements>...>::value), "All elements must be key_type");
       static_assert(sizeof...(elements) <= Size, "Number of elements exceeds capacity");
+
+      static_assert(sizeof...(elements) == Size || etl::is_default_constructible<key_type>::value,
+                    "key_type must be default constructible, if providing less than Size elements");
```
rolandreichweinbmw, line 423: I guess you would also remove the static_assert() for etl::is_default_constructible in l.458? Since the _ext variant is not about construction of the actual data at all?

HazardyKnusperkeks: Done.

